PR #8, opened 2019-10-17
Add Typescript typings

cowboyd:
## Motivation

Typescript provides a very nice level of auto-documenation and discoverability to its users, especially for a strange new api like the one provided by `effection`. By definining types for effection functions, this will let us bring it easily to the typescript community and allow it to clear the bar set by that community. We want to make sure that as the affection library evolves, its associated types also evolve.

## Approach

Rather than implement types at the `DefinitelyTyped` repository where they could conceivably get out of sync with this package, instead this adds an `index.d.ts` directly to this project that will ship with every NPM package.

To ensure that the typescript typings work, the typescript.test.js loads a set of files and then asserts that they either typecheck or that they dont.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,8 @@
 /dist
 /pkg
+/scratchpad
 /node_modules
 /benchmarks
 /README
+/**/*.d.ts
+/tests/typescript/*.ts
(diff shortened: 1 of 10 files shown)


frontsidejack:
This PR is available to use:
```bash
npm install effection@0.1.0-e08d61f
```
You can view the NPM package [here](https://www.npmjs.com/package/effection/v/0.1.0-e08d61f).
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a0b1c235740caf2d2c8efcbd2ca6d3eb3c7f4c94
</p>

PR #10, opened 2019-10-17
Upgrade `nyc` coverage tool

cowboyd:
## Motivation

We want to be above suspicion when it comes to security, and as such, there were a number of package with vulnerabilities that were being pulled in via the coverage tool. Technically, this shouldn't be a problem since coverage is only happening when running tests and not when effection is being used on the server, but we'll address it anyway.

## Approach
Upgrade `nyc` so that transitive dependencies are secure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "pack:publish": "pack build && cd pkg && npm publish $*"
   },
   "devDependencies": {
-    "nyc": "13.1.0",
     "@babel/core": "7.4.4",
     "@babel/preset-env": "7.4.4",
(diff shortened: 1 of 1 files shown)


frontsidejack:
This PR is available to use:
```bash
npm install effection@0.2.0-698df47
```
You can view the NPM package [here](https://www.npmjs.com/package/effection/v/0.2.0-698df47).
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9bb3d5512f714ae128b78c0c6c3df8fe1fe5fd74
</p>

PR #12, opened 2019-10-18
Make `fork()` a simple function

cowboyd:
Asynchronous operations are executed with the `fork()` method on `Execution` which is used to mark which exact process that you're forking off of. This is necessary because where you fork from determines who your parent will be and what happens when that parent is halted, or when you finish or error out who you need to report to.

The way we were accessing the current execution was via the `this` parameter inside of our operation functions. It was doen this way because it was easy. If you have a direct reference to the current execution, then it is very easy to fork off operations from it. Another rationale was that the `fork()` method _is_ effectful and so better to mark it as special by associating it with the `this` parameter. That way, it can be clearly seen as an imperative API.

It hasn't really panned out that way. In this day and age, it's is a very odd paradigm to be following as use of `this` is shunned. It also is a point of strangeness with the rest of the API since it is otherwise very functional in nature.

Since we can only have one execution running at a time (JS is single threaded after all), this change tracks a "current" execution which is used as the execution context any time that `fork` is called.

before:

```ts
export default function* main(env: Env): Sequence {
  this.fork(webAPI, [env.port]);
  this.fork(rtmAPI, [env.slackOAuthToken]);
}
```

after:

```ts
import { fork } from 'effection';

export default function* main(env: Env): Sequence {
  fork(webAPI, [env.port]);
  fork(rtmAPI, [env.slackOAuthToken]);
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,11 +152,11 @@ part of your main process. To do this, you would use the `fork` method
 on the execution:
 
 ``` javascript
-import { execute } from 'effection';
+import { execute, fork } from 'effection';
 
 execute(function*() {
-  this.fork(createFileServer);
-  this.fork(createHttpServer);
+  fork(createFileServer);
+  fork(createHttpServer);
 });
 ```
 
--- a/examples/async.js
+++ b/examples/async.js
@@ -1,14 +1,14 @@
 /* eslint no-console: 0 */
 /* eslint require-yield: 0 */
-import { execute, timeout } from '../src/index';
+import { execute, fork, timeout } from '../src/index';
 
 
 /**
  * Fires up some random servers
  */
 execute(interruptable(function*() {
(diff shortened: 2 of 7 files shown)


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `static-fork`.
```bash
$ npm install effection@static-fork
```
```bash
{
  "effection": "static-fork"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.2.0-f1b3e6d](https://www.npmjs.com/package/effection/v/0.2.0-f1b3e6d) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 49643114647bd28b8e9cb8eec930071e24489698
</p>

PR #13, opened 2019-10-18
Make generators (not just generator functions) a valid operation.

cowboyd:
Previously, the only way to define an operation with a co-routine was by passing a generator function to `yield`. That meant we had to do a lot of silly argument passing around the codebase just to support passing arguments to functions. Like:

```js

function* add(left, right) {
  return left + right
}

function* top() {
  let one = yield function*() { return 1; }
  let two = yield function*() { return 2; }
  return yield call(add, one, two);
}
```

But JavaScript supports passing arguments to functions just fine, right? Wouldn't it be better if we could just do it like this?

```js
function* add(left, right) {
  return left + right
}

function* top() {
  let one = yield function*() { return 1; }
  let two = yield function*() { return 2; }
  return yield add(one, two);
}
```

That's what this change does. By allowing you to invoke the generator function directly, we can do away with the `call` function altogether.

before:

```js
export function main*(env) {
  fork(webAPI, [env.port]);
  fork(rtmAPI, [env.slackOAuthToken]);
}
```

after:

```js
export function main*(env) {
  fork(webAPI(env.port));
  fork(rtmAPI(env.slackOAuthToken));
}
```

before:

```js
yield call(listen, client, self);
```

after:

```js
yield listen(client, self);
```


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `no-arg-operations`.
```bash
$ npm install effection@no-arg-operations
```
```bash
{
  "effection": "no-arg-operations"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.2.0-9fbcd6e](https://www.npmjs.com/package/effection/v/0.2.0-9fbcd6e) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 82671065eec23044b30af2e50cc000af0f024c9f
</p>

PR #19, opened 2019-10-24
Unroll continuations and state classes

cowboyd:
The implementation of execution as it stands is very "well factored" in the sense that it uses state classes to couple both state and behavior changes based on that state, as well as  monadic delimited continuations to track control flow. However, when using this in practice, it means that stack traces can become very deep very quickly and completely unhelpful as a result.

What's desired is a fundamental harmony with the underlying platform and that means when errors happen, which they always do, we want the stack traces to be as helpful as possible in diagnosing the problem.

This change integrates the individual state classes into the main Execution class (renamed Fork) and uses simple `if` statements to implement the behavior based on different states. This means that the individual methods `resume()`, `throw()`, and `halt()` are a bit more complex, but the advantage is that 100% of the logic is colocated in the method, and also that the stack traces they generate are much shallower.

Ironically, it still ended up being about 70 lines less code even when very descriptive exception messages were added.


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `unroll-continuations-and-state-classes`.
```bash
$ npm install effection@unroll-continuations-and-state-classes
```
```bash
{
  "effection": "unroll-continuations-and-state-classes"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.3.0-cb3050a](https://www.npmjs.com/package/effection/v/0.3.0-cb3050a) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against bcf3f49b0979d621fbc2c19fa960c32dcadd18c0
</p>

PR #21, opened 2019-10-30
🐛Don't mark a running fork as completed when a child is halted

cowboyd:
When you have the case of a parent `A` that is running, and also that parent has a forked child `B` that becomes halted. `A`, because it is still yielding should continue running. If on the other hand, `A` is waiting, then it should be completed because that was the last blocking child.

```
A
 - B
```

Howewer, we only checked that the parent did not have blocking children, and not that it was also just waiting (and not running).

This adds the check to the `join()` method to ensure that running parents continue running even when a child is halted.

### DIFF
--- a/src/fork.js
+++ b/src/fork.js
@@ -140,7 +140,7 @@ Thanks!
       if (child.sync) {
         this.throw(new Error(`Interupted: ${child.result}`));
       } else {
-        if (!this.hasBlockingChildren) {
+        if (this.isWaiting && !this.hasBlockingChildren) {
           this.finalize('completed');
         }
       }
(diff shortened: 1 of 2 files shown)


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `halt-forked-child-of-yielding-operation`.
```bash
$ npm install effection@halt-forked-child-of-yielding-operation
```
```bash
{
  "effection": "halt-forked-child-of-yielding-operation"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.3.1-3c0e577](https://www.npmjs.com/package/effection/v/0.3.1-3c0e577) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d4f5825a02475e3e1d273c6050a08e87ebf62d2d
</p>

PR #22, opened 2019-10-30
Always cleanup children

cowboyd:
When debugging an effection application, we noticed that in the case where you had halted children, they were not being removed from parents that were still running. That meant that you have a fork tree looking like:

```
Fork {
   state: waiting
   operation: main
   children:
     Fork {
       state: running
       operation:
       children:
     }
     Fork {
       state: running
       operation:
       children:
           Fork {
             state: halted
             operation: setupConnection
             children:
           }
     }
}
```

Which means that if the the parent is long running, then you can accrue lots of halted children which is a memory leak.

This makes sure that _every_ time that a child joins its parent, it is removed from the list of children. This is safe because join _only_ happens on finalization of a child.

### DIFF
--- a/src/fork.js
+++ b/src/fork.js
@@ -140,7 +140,7 @@ Thanks!
       if (child.sync) {
         this.throw(new Error(`Interupted: ${child.result}`));
       } else {
-        if (!this.hasBlockingChildren) {
+        if (this.isWaiting && !this.hasBlockingChildren) {
           this.finalize('completed');
         }
       }
--- a/src/fork.js
+++ b/src/fork.js
@@ -125,25 +125,31 @@ https://github.com/thefrontside/effection.js/issues/new
 Thanks!
 `);
     }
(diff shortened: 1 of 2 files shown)


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `always-cleanup-children`.
```bash
$ npm install effection@always-cleanup-children
```
```bash
{
  "effection": "always-cleanup-children"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.3.1-2f60e15](https://www.npmjs.com/package/effection/v/0.3.1-2f60e15) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2a0924c96decf6257b8a07a6ac701a6f3002a9ff
</p>

PR #51, opened 2019-11-26
Make then/catch/finally callable from typescript

jnicklas:
While it is currently possible to use executions with aysnc/await and `Promise.all`, actually `then`/`catch`/`finally` is not allowed from TypeScript. This just adds these methods to the public interface.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,11 +5,13 @@ declare module "effection" {
 
   export interface Sequence extends Generator<Operation, any, any> {}
 
-  export interface Execution<T = any> {
+  export interface Execution<T = any> extends PromiseLike<any> {
     id: number;
     resume(result: T): void;
     throw(error: Error): void;
     halt(reason?: any): void;
+    catch<R>(fn: (Error) => R): Promise<R>;
+    finally(fn: () => void): Promise<any>;
   }
 
   export function fork<T>(operation: Operation): Execution<T>;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -5,11 +5,13 @@ declare module "effection" {
 
   export interface Sequence extends Generator<Operation, any, any> {}
 
-  export interface Execution<T = any> {
+  export interface Execution<T = any> extends PromiseLike<any> {
```
jnicklas: Eventually, it would be nice to be able to specify the type of this promise, since the argument to fork is *usually* a generator, and we know the return value of a generator, we should actually be able to thread this value through, I think?

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `thenable`.
```bash
$ npm install effection@thenable
```
```bash
{
  "effection": "thenable"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-8e2f32f](https://www.npmjs.com/package/effection/v/0.4.0-8e2f32f) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 4d91606b178d498ed4ed72d708357f9381ca6879
</p>

jnicklas:
Merging this, since it should be pretty uncontroversial, imo, and it'll make it easier to proceed with the send/receive stuff.

PR #52, opened 2019-12-03
♻Always resume fork evaluation on next tick

cowboyd:
Currently, forks execute synchronously until their first yield point. Which means that if we have:

```js
fork(function* outer() {
  fork(function* inner() {
    throw new Error('boom!');
  });
})
```

It will attempt to throw an error into the outer generator function while that generator function is still running. This causes the original exception to be obscured.

For a full write up see here https://github.com/thefrontside/effection.js/issues/26

This change implements the first option in that ticket by making sure that calling `Fork#resume()` does not continue the computation at that point, but instead merely _schedules_ it to be continued. This was a pretty easy change to make. In fact, the biggest impact seems to be that forks do exist for some time in the `unstarted` state and so it's important that `unstarted` be considered a blocking state since otherwise, a parent with no yield points would just consider itself done if it had no blocking children.

Pros
====

* Very simple resolution to the issue, and simplicity has an elegance all its own.

Cons
====

* _Loses the stack trace when there is an exception_: In the event that you _do_ have an error the stack trace will be very short, basically just a direct line from `setTimeout`:
  ```js
    at child.resume();
    at applicationErrorLocation();
  ````
* _Testing is more complex:_ I'm not sure if this is indicative of a deeper issue, but it's definitely more fiddly and that makes me a tad nervous
* Loses the simplicity of a single, continuous computation: There is a  certain elegance to the idea the entire tree is once continuos  computation only broken appart by yield points. By introducing  another form of asynchrony (forks happening in next tick). There are   now two timing mechanisms that you have to hold in your head so that   you don't get caught out by subtle bugs. For example, take the code here:

  ```js
  function* networkStuff(host, port) {
    let socket = createConnection(port, host);

    let monitor = fork(function* errors() {
      let [error]: [Error] = yield on(socket, 'error');
      throw error;
    })

    yield on(socket, 'ready');

    socket.write(data);
  }
  ```
  There is a potential to "miss" an error event here because with this  change, the `error` event listener will not be installed until at   least the next tick of the run loop _after_ the socket connection has   been attempted. If an error is emitted between those two points, then  it will be dropped on the floor. This seems especially likely with something like EADDRINUSE.

It's possible to avoid this by creating the socket and then connecting to it in two separate steps:

```js
 function* networkStuff(host, port) {
    let socket = new Socket(); 

    let monitor = fork(function* errors() {
      let [error]: [Error] = yield on(socket, 'error');
      throw error;
    })

    //must happen after monitor install
    socket.connect(host, port);

    yield on(socket, 'ready');

    socket.write(data);
  }
```

But it does feel weird that if you were to move the socket connection up above the monitor creation, then you'd be back in the same boat.

### DIFF
--- a/tests/async.test.js
+++ b/tests/async.test.js
@@ -130,7 +130,7 @@ describe('Async executon', () => {
   describe('with a mixture of synchronous and asynchronous executions', () => {
     let execution, one, two, sync, boom;
     beforeEach(async () => {
-      await new Promise((resolve, reject) => {
+      await new Promise((resolve) => {
         boom = new Error('boom!');
         execution = fork(function*() {
           fork(function*() { yield cxt => one = cxt; });
@@ -139,7 +139,7 @@ describe('Async executon', () => {
             yield cxt => (sync = cxt) && resolve();
           };
         });
-      })
+      });
       expect(one).toBeDefined();
       expect(two).toBeDefined();
       expect(sync).toBeDefined();
--- a/tests/async.test.js
+++ b/tests/async.test.js
@@ -10,7 +10,7 @@ describe('Async executon', () => {
   describe('with asynchronously executing children', () => {
     let execution, one, two, three;
 
-    beforeEach(() => {
+    beforeEach((done) => {
       execution = fork(function() {
         fork(function*() {
           yield cxt => one = cxt;
@@ -21,7 +21,10 @@ describe('Async executon', () => {
         });
 
         fork(function*() {
-          yield cxt => three = cxt;
+          yield cxt => {
+            three = cxt;
+            done();
+          };
         });
       });
     });
@@ -126,15 +129,17 @@ describe('Async executon', () => {
 
   describe('with a mixture of synchronous and asynchronous executions', () => {
     let execution, one, two, sync, boom;
-    beforeEach(() => {
-      boom = new Error('boom!');
-      execution = fork(function*() {
-        fork(function*() { yield cxt => one = cxt; });
-        fork(function*() { yield cxt => two = cxt; });
-        yield function*() {
-          yield cxt => sync = cxt;
-        };
-      });
+    beforeEach(async () => {
+      await new Promise((resolve, reject) => {
+        boom = new Error('boom!');
+        execution = fork(function*() {
+          fork(function*() { yield cxt => one = cxt; });
+          fork(function*() { yield cxt => two = cxt; });
+          yield function*() {
+            yield cxt => (sync = cxt) && resolve();
+          };
+        });
+      })
       expect(one).toBeDefined();
       expect(two).toBeDefined();
       expect(sync).toBeDefined();
@@ -252,9 +257,12 @@ describe('Async executon', () => {
 
   describe('A parent that block, but also has an async child', () => {
     let parent, child;
-    beforeEach(() => {
+    beforeEach((done) => {
       parent = fork(function*() {
-        fork(function*() { yield cxt => child = cxt; });
+        fork(function*() {
+          done();
+          yield cxt => child = cxt;
+        });
         yield x => x;
       });
     });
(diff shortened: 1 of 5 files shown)


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `schedule-fork-in-next-tick`.
```bash
$ npm install effection@schedule-fork-in-next-tick
```
```bash
{
  "effection": "schedule-fork-in-next-tick"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-8761532](https://www.npmjs.com/package/effection/v/0.4.0-8761532) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 65f226e74e73f47e2ab0bb5806890ded66cffbf8
</p>

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `schedule-fork-in-next-tick`.
```bash
$ npm install effection@schedule-fork-in-next-tick
```
```bash
{
  "effection": "schedule-fork-in-next-tick"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-c28c1a6](https://www.npmjs.com/package/effection/v/0.4.0-c28c1a6) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e53ad7c5eaf82981730e510defd9e7f590aa4ffb
</p>

jnicklas:
@cowboyd in my implementation of this in BigTest, I worked around this issue by filtering events into messages, or throwing directly. For the error event this looks like this:

https://github.com/bigtestjs/server/blob/0874779c728af0b9687de1e36dec8262e63b8f9e/src/effection/events.ts#L31-L40

The crucial component here is that this monitor is installed synchronously and also *removed* synchronously, which is the reason for the addition of `atExit`.

jnicklas:
When I played around with this I realized that there is another subtle issue that needs to be handled. When we use `setTimeout` to schedule the start of the Fork, it might actually already be finalized when we hit `resume`, which will cause a nasty error. The reason is that since the `resume` is asynchronous, we may have already finalized the fork between the set timeout and the resume. This is actually pretty realistic if there is an error in the parent fork which causes the parent fork to crash and therefore terminate its children. So something like this:

``` js
fork(function*() {
  fork(function*() { yield });
  throw new Error('boom');
});
```

My solution to this was to store the timeout on the fork, and call `clearTimeout` in finalize. I added a `start` method to fork to make this easier, this way it can be called both from `Fork#fork` and from the `fork` free function when creating a root fork.

tests/async.test.js
```diff
@@ -21,7 +21,10 @@ describe('Async executon', () => {
         });
 
         fork(function*() {
-          yield cxt => three = cxt;
+          yield cxt => {
+            three = cxt;
+            done();
```
jnicklas: Can we really be sure that *all* of the forks have completed here?

cowboyd:
Another solution that is also very conceptually simple, but would not involve any asynchrony, but that would involve more of an API change, would be to make `fork` an operation. In other words, calling `fork` would not actually do anything except construct an operation that spawns the fork.

```js
fork(operation);
```

now becomes

```js
yield fork(operation);
```

We would need a `join` operation that would wait on a fork, so what is now:

```js
yield fork(operation);
```

would become

```js
yield join(fork(operation));

// or using the soon-to-come pipe operator :)

yield operation |> fork |> join
```

While this would require more implementation and a larger API surface area (I think we might need a `begin` method` to enter the root fork.) It's still conceptually simpler since you don't have the weirdness of a `fork` function having a side-effect, which has tripped us up in the past. It's also not then possible to have rogue functions adding forked children without prior knowledge.

Any time you want to fork, you'll have to yield control temporarily.

Just a thought.

PR #53, opened 2019-12-04
Update action source from master to v1.1

minkimcello:
We're going to be introducing some breaking changes and so we created a "save point" by tagging and releasing the current state of our `thefrontside/actions` as `v1.1`. There might be other bugs but going from `master` to `v1.1` in this repository's workflows should not be any different as `master` and `v1.1` are identical at the moment.

Updated `release` and `synchronize-npm-tags` to `v1.1`. I left the preview workflow action as `v1` because the reason why `v1` was even released was for that workflow to begin with.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Tag and Publish
-      uses: thefrontside/actions/synchronize-with-npm@master
+      uses: thefrontside/actions/synchronize-with-npm@v1.1
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `mk_action-version`.
```bash
$ npm install effection@mk_action-version
```
```bash
{
  "effection": "mk_action-version"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-7f4287a](https://www.npmjs.com/package/effection/v/0.4.0-7f4287a) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 405ccb9fd3a892d7cdae9ebf8eca04f411719f13
</p>

Issue #33, opened 2019-11-13
Forking a controller


jnicklas:
If I have a function which returns a controller, I currently have to write this:

``` js
fork(function*() { yield returnsController() });
```

It would be nice if there were some syntax sugar which would allow me to write:

``` js
fork(returnsController());
```

With the exact same behaviour as above.

Yay or nay?

cowboyd:
Yes, very much so. It's becoming clearer that there is still something more primitive to uncover. I think that the controller might be it, and that what we really want to do is yield / fork to a controller, and in both cases, the process is resolving the thing into a controller.

In other words, when we see: `yield X`, we need to have a function `f(X) -> Controller` so that we're yielding to a controller. Same with `fork(X)` we can use `f` to convert to a controller and then proceed.

cowboyd:
Closed by #57 

`fork` is now an operation that takes any operation:

* `fork(new Promise())`
* `fork()`
* `fork(fork())`

are all valid forks.

PR #57, opened 2019-12-15
Convert `fork` into an operation

### LINKED ISSUE #26: Forks that throw immediately try to abort their parent while the parent is still running
Take the following case:

```js
import { fork } from 'effection';

fork(function* outer() {
  fork(function* inner() {
    throw new Error('boom!');
  });
});
```

This aborts with an error as expected, but the error is not super useful, since it does not actually contain the original exception, but rather a complaint about incorrect `Generator` states:

```
TypeError: Generator is already running
    at outer.throw (<anonymous>)
    at thunk.iterator (fork.js:70:30)
    at Fork.fn [as enter] (fork.js:181:14)
    at Fork.enter (fork.js:158:23)
    at Fork.thunk (fork.js:70:12)
    at Fork.join (fork.js:152:7)
    at Fork.join [as finalize] (fork.js:196:19)
    at Fork.finalize [as thunk] (fork.js:172:12)
    at Fork.thunk [as resume] (fork.js:91:12)
    at Fork.resume (fork.js:89:12)
```

This is because when an exception occurs in a `Fork`, we catch it and then progressively call `throw` on the parent `Fork`s all the way up to the root fork (provided that nobody catches it). There problem occurs here where the `outer` generator is actually part of the parent stack of the `inner`. 

This is an invalid operation on a generator that is currently running, which `outer` still is. In other words, it has not yielded yet.

While this does not change the logic of a program in the sense that it will still fail at the right point, it nevertheless _hides the root cause_ of the exception by replacing the actual error with the `TypeError: Generator is already running`

It's not obvious what the solution is, but some of the ones discussed so far are:

1. make all forks happen in the next tick of the run-loop. That way, the parent is guaranteed to continue to a yieldpoint before the exception is thrown and so `parent.throw()` will be a valid operation
2. detect when the parent is currently on the call stack, and re-throw the error to allow it to percolate to the parent after child finalization.
3. collect forked children on a queue which is then resumed in-order when the parent reaches the next yieldpoint.

We need to research these possible approaches, as well as identify others before coming to a conclusion.

cowboyd:
Motivation
----------

Originally, this change came out of problems with sequencing a fork. Specifically, if there was an error inside of fork, and so it tried to throw() inside of its parent, then the error was hidden by the fact that [its parent generator was already running]. This was symptomatic of a deeper problem of the generator syntax being too tightly coupled to the underlying concept of the execution of a sequence of steps, and also of the concept of synchronous and asynchronous execution being too coupled.

As a purely practical consideration, we wondered "what if the parent generator was yielded at the time we tried to create the fork?" That way, the fork would be either be at its first yield point, or have
failed when we tried to resume the parent, but this turned out to be a profoundly positive re-orientation of the effection architecture to separate out the concepts of context and executions from the generator synax entirely which enables all sorts of nice things.

Fundamentally, this change makes effection a function-oriented effects library rather than a generator-oriented effects library. That is, you don't _need_ generators to declare a tree of processes, it's just nicer to do it that way. For example:

```js
spawn(function*() {
  yield fork(function*() {
    return 5;
  });
});
```

can now be re-written using the function-equivalent:

```js
spawn(({ spawn }) => {
  spawn(fork(({ resume }) => resume(5)));
})
```

The nice effect that this has is that the generator syntax is implemented as just a plugin to the basic runtime, which means that in addition to solving the problem at hand, it also [allows you to fork a control function directly][2], something that had always felt like it should work, but annoyingly didn't.

This feels like the right direction to take it, because it is the solution to several seemingly unrelated problems, and it even feels as though it could provide a pathway to an [asynchronous halt operation][3] by making not just `fork` but also `halt` an operation.

Finally, because structural concurrency is implemented ultimately as operations that do nothing more than manipulate the execution context, other structured effects (state, messaging, etc...) can be implemented in user space. Using this library, I was able to implement messaging operations based on the [send-and-receive][4] branch completely in user-space, and in which both `send()` and `receive()` were operations.

You can now seemlessly move between function syntax and generator syntax since they are all just operations that run inside the bounds of an execution context.

For example, here is a "hybrid" version of the previous example using a mixture of function syntax and generator syntax.

```js
spawn(({ spawn }) => {
  spawn(fork(function*() {
    return 5;
  });
})
```

From the perspective of the JavaScript runtime, execution is still completely synchronous internally, and the code is much, much simpler.

Details
-------

What was formerly known as a `Fork` and was coupled to a generator, has now become two objects, an `ExecutionContext` and a `ControlFuction`. Where the `ExecutionContext` tracks the state of the frame of execution and the control function calls `resume`, `fail`, or `halt` at the appropriate time. Every operation is eventually resolved to a control function.

The key new primitive that is introduced is the `spawn` function, which is passed into the control function with each execution context:

```js
spawn(({ spawn }) => {
  call(function*() {
    yield timeout(100);
    return 5;
  });
});
```

This provisions a new execution context as a child of the current context, looks up the operation passsed in order to find its control function, and then enters that child context. Essentially, this is an implementation of a call stack, except instead of it being a stack, it's a tree.

Most effection code however, will never need to know about `call`, since most operations do not actually create any children. But for those that do such as the generator operation and the `fork` operation, it allows them to be treated uniformly.

By default, child contexts created by `spawn` are detached. This means that the only constraints is that their parents will not be considered finished until they themselves are finished, error handling, halt handling, and resume handling is up to the operation that calls spawn.

```js
let child = spawn(operation);
child.ensure(() => {
  if (child.isErrored) {
    console.error(child.result);
  }
});
```
As mentioned previously, the default behavior is to cause the parent context to also fail.

It's worth noting that with this power, it is possible to implement completely in user-space try/catch/finally as operations, for example:

```js
function*() {
  yield tcf({
    try: () => do something,
    catch: (e) => console.log('error = ', e),
    finally: () => teardown()
  })
}
```

Open Questions
______________

- [ ] I'm not sure if the interface for the controls is good. Right
  now, the full signature of a control function is `({ resume, fail,
  ensure, call, context })`, but part of be feels like it should be
  more like a promise: `({ resolve, reject, ensure, call, context
  })`. Although maybe even using pseudo reserved words would be better
  to re-enforce the concepts like `({ $continue, $throw, $call,
  $finally })`.
- [x] compact context on `fork`? Currently, when you fork, the child operation is appended directly to the parent:
```
  - top
     - child (forked)
```
~~However, with this change, the fork _will_ have a place in the call tree:~~
```
- top
  - fork
    - child
```
Is this what we want? We can get rid of it using the quasi-monadic trick that if a control function returns another control function, then the second control function is run inside the enclosing context. so that it looks the way it did before.

> Update, by spawning in the parent, fork nicely avoids this suboptimal configuration

[1]: https://github.com/thefrontside/effection.js/issues/26
[2]: https://github.com/thefrontside/effection.js/issues/33
[3]: https://github.com/thefrontside/effection.js/issues/35
[4]: https://github.com/thefrontside/effection.js/pull/49

resolves #26, #33 
closes #52


src/context.js
```diff
@@ -0,0 +1,191 @@
...
+      child.halt(result);
+    }
+    this.finalizePromise();
+  }
+
+  createController(operation) {
+    let controller = ControlFunction.for(operation);
+    if (!controller) {
+      throw new Error(`cannot find controller for ${operation}`);
+    }
```
cowboyd: This is how we can make operations and effects super-general. Instead of just giving up if we don't find the controller for a given operation, we can search _upwards_ on the stack of execution contexts, checking to see at each point if it can handle the operation.

cowboyd:
For some reason, this line is type checking when it shouldn't https://github.com/thefrontside/effection.js/blob/8f73e348e47e6d1b798a5d66751e431d8d180d26/types/sequence.test.ts#L20-L21 and I'm not certain why.

README.md
```diff
@@ -64,20 +64,20 @@ siblings are immediately halted.
...
 
-The process primitive is the `Execution`. To create (and start) an
+The process primitive is the `Execution`. To create (and enter) an
 `Execution`, use the `fork` function and pass it a generator. This
 simplest example waits for 1 second, then prints out "hello world" to
 the console.
 
 ``` javascript
-import { fork, timeout } from 'effection';
+import { enter, timeout } from 'effection';
```
jnicklas: How about `run` instead of `enter`? Sounds a bit better to me.

src/context.js
```diff
@@ -0,0 +1,191 @@
+import { ControlFunction, HaltError } from './control';
+
+export class ExecutionContext {
```
jnicklas: As mentioned in #61, I think `Task` might be a nice name for this?

src/context.js
```diff
@@ -0,0 +1,191 @@
...
+      }
+    });
+    child.enter(operation);
+    return child;
+  }
+
+  ensure(hook) {
+    let { exitHooks } = this;
+    exitHooks.add(hook);
+    return () => exitHooks.delete(hook);
```
jnicklas: If we ever re-assign `this.exitHooks`, then holding onto this reference could actually be a bit problematic, I think? Maybe it's better to just use `this.exitHooks` directly?

taras:
pretty epic

examples/countdown.js
```diff
@@ -1,6 +1,7 @@
 /* eslint no-console: 0 */
-import { fork, timeout } from '../src/index';
+import { enter, timeout } from '../src/index';
 
+import { interruptable } from './inerruptable';
```
jnicklas: 🐞 "inerruptable"

package.json
```diff
@@ -6,7 +6,7 @@
   "author": "Charles Lowell <cowboyd@frontside.io>",
   "license": "MIT",
   "private": false,
-  "types": "types/index.d.ts",
+  "types": "index.d.ts",
```
jnicklas: Isn't this change in master already? Not sure why we are getting a diff here.

src/context.js
```diff
@@ -0,0 +1,191 @@
...
+      }
+    } else {
+      throw new Error(`
+Tried to call Fork#resume() on a Fork with state '${this.state}' This
+should never happen and so is almost assuredly a bug in effection. All of
+its users would be in your eternal debt were you to please take the time to
+report this issue here:
+https://github.com/thefrontside/effection.js/issues/new
+
+Thanks!`);
```
jnicklas: Is it just me or is this not actually true anymore? `resume` is part of the public API now, so someone creating a controller could very well call `resume` twice, which would still be wrong, but not a bug in effection?

src/context.js
```diff
@@ -0,0 +1,191 @@
...
+    }
+    if (this.options.resume) {
+      this.options.resume(value);
+    }
```
jnicklas: IMO this should be moved into the `if` statement above, I know we can't get here if we throw, but this still makes the flow harder to follow.

src/context.js
```diff
@@ -0,0 +1,191 @@
...
+
+  get root() {
+    if (!this.parent) {
+      return this;
+    } else {
+      return this.parent.root;
+    }
+  }
+
+  call(operation, options = {}) {
```
jnicklas: I am somewhat confused by this method name, I think it doesn't describe the operation it performs very well. Maybe something like `spawn` would be better?

src/context.js
jnicklas: Since this is similar to the Promise API, it might be worth mimicking how that API works, where I think resolving after the first time is simply a no-op.

jnicklas:
First of all, I'm super, super happy with this. This change makes Effection better in every way. One thing I am wondering if we can push even further towards similarity with the Promise API. I think it's becoming clearer and clearer that an ExecutionContext is really just a fancy promise with some additional stuff (ability to spawn children, halt, etc). Maybe it should also *look* even more like a promise. So if we have a promise like this:

``` javascript
let promise = new Promise((resolve) => {
  setTimeout(resolve, 2000);
});
```

Maybe we should be able to create an execution context in a similar way:

``` javascript
let context = new ExecutionContext(({ resolve }) => {
  setTimeout(resolve, 2000);
});
```

This isn't so different from what we have now, we're basically just combining the constructor and `enter`. The main difference would be in the returned object. The point of the way the Promise API is designed is to not leak any mutable state outside the constructor. Once constructed, and when passed to some other collaborator, a Promise is completely immutable. This is a very nice property, which I think we would do well to steal.

I am somewhat torn on whether using an object as the argument to this callback function is better or worse.

```
new ExecutionContext((resolve) => {}) // this?
new ExecutionContext(({ resolve }) => {}) // or this?
```

On one hand it makes the API much more convenient, especially since we'll have a bunch of functions we need to send in, but it does mean allocating an extra object for this. I think it's probably worth it, since we'll have at least four (I think?) methods which we want to send in, and four positional arguments is pretty awkward.

jnicklas:
This is also why I am partial towards `Task` as a name. This looks very nice and friendly to me, and somewhat a natural progression from promises:

``` javascript
let context = new Task(({ resolve }) => {
  setTimeout(resolve, 2000);
});
```

jnicklas:
So just after leaving this comment I realized that it is pretty much nonsense, since we can't expose the constructor as a public API anyway. I'm leaving it here anyway, since I think there is definite value in seeing the similarity between Promise and ExecutionContext, and I think pushing further in this direction can only be a good thing.

README.md
cowboyd: Yeah, `run` feels good, I had also considered `main()` since it's basically a single entry point into the top of your execution tree. The only reason I'd want to stick with `enter` or `main` is that really, this function is ideally meant to be called only once, where as `run` feels like maybe it's something you should do multiple times. Thoughts?

cowboyd:
> So just after leaving this comment I realized that it is pretty much nonsense, since we can't expose the constructor as a public API anyway. I'm leaving it here anyway, since I think there is definite value in seeing the similarity between Promise and ExecutionContext, and I think pushing further in this direction can only be a good thing.

So would you vote in favor of changing the flow control functions to align with the Promise API?

So:

```ts
  interface Controls<T> {
    resume(result: T): void;
    fail(error: Error): void;
    ensure(callback: (context: Context<T>) => void): void;
    call<R>(operation: Operation<R>): Context<R>
    context: Context<T>;
  }
```
becomes
```ts
  interface Controls<T> {
    resolve(result: T): void;
    reject(error: Error): void;
    finally(callback: (context: Context<T>) => void): void;
    call<R>(operation: Operation<R>): Context<R>
    context: Context<T>;
  }
```
> It looks like you cannot have a variable named `finally`, so `({ finally }) => xyz` would be an invalid control function.

This is definitely a step in that direction and would re-use the knowledge that people already have.

```js
let good = enter(({ resolve }) => resolve('hello world'));
good.result //=> 'hello world'
good.isCompleted //=> true

let bad = enter(({ reject }) => reject(new Error('not good')));
bad.result //=> Error('not good')
bad.isErrored //=> true
```

Also, perhaps we could align our states with promises as well:

```
isCompleted -> isResolved
isErrored -> isRejected
isSettled = isCompleted || isErrored && !isWaiting
```

Thoughts?

package.json
cowboyd: I tried to rebase and include the fix and I think I mucked it up 😕

src/context.js
cowboyd: Nice catch, not true at all.

src/context.js
cowboyd: The analogy that I was holding in my mind was normal function evaluation, except instead of a call stack, we have a call tree.

When you invoke a function inside most languages, the rough algorithm is:

```
1. allocate a stack frame
   1.a set arguments on the new frame
   1.b set reference to parent frame for looking up lexical variables
2. transfer control to new stack frame
3. function body runs
4. transfer control back to parent frame with return value from child frame
```

I picked `call` because that's very similar to the semantics of what we're doing here except instead of a call stack, we have a call tree.

allocate a child, passing in a reference to the parent, enter it, and then handle the result.

README.md
cowboyd: Another idea that just occurred to me is that we could have a top-level `call` method, that way, you would just enter a tree with a top-level `call` on an operation.

taras: `main` sounds more clear if the intention is to create a single entry point. I agree that `run` and `call` both feel like something that happens a lot.

cowboyd:
> For some reason, this line is type checking when it shouldn't
> 
> https://github.com/thefrontside/effection.js/blob/8f73e348e47e6d1b798a5d66751e431d8d180d26/types/sequence.test.ts#L20-L21
> 
> and I'm not certain why.

@taras @jnicklas any insight here would be very much appreciated since I'm completely stuck :)

You can see the error locally when running `$yarn test:types`. That line _should_ be a type error, but for some reason it isn't.

taras:
> You can see the error locally when running `$yarn test:types`. That line _should_ be a type error, but for some reason it isn't.

@cowboyd is it only not throwing an error in CI?

src/context.js
cowboyd: Should it be a warning?

cowboyd:
> @cowboyd is it only not throwing an error in CI?

No, it throws locally, but I can't figure out why 😖

taras:
> > @cowboyd is it only not throwing an error in CI?
> 
> No, it throws locally, but I can't figure out why 😖

It's supposed to throw wherever it runs. Is that not happening?

cowboyd:
> > > @cowboyd is it only not throwing an error in CI?
> > 
> > 
> > No, it throws locally, but I can't figure out why 😖
> 
> It's supposed to throw wherever it runs. Is that not happening?

No, it's not throwing with is weird. The TypeScript type checker is saying that the code is a-ok, but it should be a type error because `5` is not a valid operation. Something is amiss.

README.md
jnicklas: `main` would feel more natural if we'd only ever call this once, like a `main` function in a C-program. But this is not necessarily the case.

taras: Agree, `main` works if it's expected to be unique in the process.

src/context.js
jnicklas: With promises, this is a no-op without warning, so I think that behaviour is acceptable.

src/context.js
jnicklas: I can sort of see that, but when reading and trying to understand how this fits together, the name really threw me off. I didn't understand what this was for at all, and only by reading the source did I understand that this spawns a new child, so at least the name did not fit with my mental model of what `call` would usually refer to.

And regarding the algorithm you sketched out, the last step "transfer control back to parent frame with return value from child frame" doesn't really happen when we call `call`, we get back what is essentially a Future/Promise of a value, not the value itself. Awaiting it is a separate step. And crucially the operation is executed concurrently with other operations.

I think that `spawn` is a pretty common name for this, but `fork` might also be an acceptable alternative, though it'd be nice to not overload that term.

README.md
cowboyd: I think that this is fundamentally the same operation as what you will do to create a sub child, so `spawn`, `call` or what have you. Let's resolve that discussion, and then that will inform what to do here.

src/context.js
cowboyd: If I understand what you're saying, `call` has a very strong association with _synchronous_ function invocation. In fact, `call` is already a method on `Function` and the semantics of that are well known. In effection however, we could be synchronous or we could be asynchronous, and there's really no way to tell which it's going to be. On the other hand (at least for me) `spawn` and `fork` have very strong associations with _asynchronous_ execution, but again, we could be synchronous, or we might just be asynchronous. There's no way to tell.

Really, what we have is something akin to scheme's `call/cc` (call with current continuation) but that would be a silly name to use. Ideally, we can find a term that's neutral with respect to timing, but does capture the idea of doing some abstract operation within a bounded context. A litmus test is that it should read  "X this operation using that context" e.g.

* `call` this operation using that context
* `spawn` this operation using that context
* etc....

`invoke`, `start`, `begin`, `execute`, `perform`, `run` are some alternatives.

I kinda like `perform` because it is used in ember-concurrency. It also has an association with other similar efforts at structured effects:

* https://github.com/macabeus/js-proposal-algebraic-effects
* https://esdiscuss.org/topic/one-shot-delimited-continuations-with-effect-handlers

fwiw, I got the call nomenclature from https://github.com/phenax/algebraic-effects#state-effect-counter-example

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `fork-join-primitive-operations`.
```bash
$ npm install effection@fork-join-primitive-operations
```
```bash
{
  "effection": "fork-join-primitive-operations"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-891b693](https://www.npmjs.com/package/effection/v/0.4.0-891b693) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d7d7ac5f835ee6c1945a6c4e37cc96e7c0601da5
</p>

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `fork-join-primitive-operations`.
```bash
$ npm install effection@fork-join-primitive-operations
```
```bash
{
  "effection": "fork-join-primitive-operations"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-7ebadb1](https://www.npmjs.com/package/effection/v/0.4.0-7ebadb1) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 11be07970dc420063ae39d785aa7e4b8a4ea0ee3
</p>

README.md
cowboyd: renamed to `spawn` for now.

src/context.js
cowboyd: Let's hold off for now and see where it goes.

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `fork-join-primitive-operations`.
```bash
$ npm install effection@fork-join-primitive-operations
```
```bash
{
  "effection": "fork-join-primitive-operations"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-680f556](https://www.npmjs.com/package/effection/v/0.4.0-680f556) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 25203c668c0f707eba24445582b157b23d509e8e
</p>

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `fork-join-primitive-operations`.
```bash
$ npm install effection@fork-join-primitive-operations
```
```bash
{
  "effection": "fork-join-primitive-operations"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-c5a73d3](https://www.npmjs.com/package/effection/v/0.4.0-c5a73d3) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d866303dbacc7a699a2f11e8469790d054f945b7
</p>

PR #63, opened 2020-01-24
An operation based send/receive

cowboyd:
Taken from the implementation in https://github.com/thefrontside/effection.js/pull/49


index.d.ts
```diff
@@ -32,4 +32,23 @@ declare module "effection" {
   export function join(context: Context): Operation;
 
   export function timeout(durationMillis: number): Operation;
+
+
+  // send/receive operations, will probably be moved to library at some point.
```
cowboyd: I put these at the bottom of the type declarations since I think it's likely that we'll extract this into an Addon package like `@effection/mailbox` or `@effection/events`

index.d.ts
```diff
@@ -32,4 +32,23 @@ declare module "effection" {
...
+  type PredicateFn = (value: any) => boolean;
+  type AnyTypeName = "undefined" | "object" | "boolean" | "number" | "bigint" | "string" | "symbol" | "function" | "array"
+  type Primitive = string | boolean | number | bigint | symbol | null | undefined;
+  interface PatternMatchObject { [key: string]: Pattern }
+  export type Pattern = Primitive | PatternMatchObject | Pattern[] | PredicateFn
+
+  export function receive(): Operation;
+  export function receive(context: Context): Operation;
+  export function receive(pattern: Pattern): Operation;
+  export function receive(pattern: Pattern, context: Context): Operation;
```
cowboyd: I switched the order of the arguments for the cases where you pass both a pattern and a context. My thinking was that if we support something like a curried interface, it's more likely curry in the pattern that can be received in many context, rather than curry in a single context that can receive multiple patters.

PR #73, opened 2020-02-06
Strictly define order of operations when halting a tree

cowboyd:
Motivation
----------

When converting fork to an operation, we changed the way that continuation and failure are reported up the tree to the parent of the operation. Instead of calling the `join()` method on the parent, we used the general "exit hook" mechanism to control the flow of the program. When a child was created, the parent added shutdown hooks to it that would handle the result of the child finishing and use that to pass control back to the parent.

The problem with this is that if one of your shutdown hooks has the responsibility of running other shutdown hooks, which could cause shutdowns, that run other shutdown hooks, you can zig and zag across different contexts and experience wildly different orders of shutdown and context settlement depending on how a context resolves (success, failure, cancellation, etc..)

In fact, we were observing this behavior, when shutdown hooks were running in different orders, we'd see vastly different results.

What we'd like to see however is a strict reverse-depth-order teardown so that whenever a portion of the tree is de-commissioned, nodes that can have dependencies on prior nodes are de-commissioned first.

For example, in the following tree:

(A,
  (B)
  (C))

As children, both `B`, and `C` can depend on A, and even though `B` and `C` are siblings, since `C` was created later in the computation, it is conceivable that it depends on `B`. The reverse is not (easily) true. Therefore, the order of destruction we want is `C` -> `B` -> `A`.

Approach
--------

We address this problem by splitting shutdown callbacks into two distinct categories. The first is the original mechanism, added by `ensure()`, and which now has _zero_ effect on the control flow of the application. The exit hooks added in this way are just a flat list of functions to run a single time when the context is finalized. Each hook in the list only operates within the single context, and is mainly just meant to be added from user-land to control runtime resources.

The second kind of hook is used exclusively for handing off control of the computation from the finalizing context. In other words, "the computation represented by this context is now complete, what happens next?" While exit hooks can be many, there can only be a single continuation hook, and if we think about it, this make sense. Like a normal function, it can do a great many things, call out to a number of other functions, but it can only return to a single place.

In summary, by placing these two callbacks into different worlds, we can guarantee that shutdown hooks are run in a very precise order, and that they cannot (without dedicated nefarious hackery) interfere with the continuation of the computation.

Implementation
--------------

This uses the simplest thing that could possibly work: leave the current teardown mechanism as-is, and introduce a single callback to a new context. This callback is aptly name "continue", and is invoked as the very last thing a context _ever does_.

Now, finalize can only every be invoked a single time, and has a very
strict structure.

1. record results
2. halt all children in reverse order
3. run all exit hooks in reverse order
4. finalize the promise.
4. continue.

New Guarantees:

- spawn: all spawned child contexts are destroyed before their parent context is finalized. There is no continuation by default, so it just goes away without having any effect on the parent. However, if the parent is halted at any point during its run, it will be halted immediately.
- monitor: In addition to spawn guarantees, any failed monitor raises failure in its parent
- fork: In addition to monitor guarantees. parent may not continue until completed (though it may spawn more operations).
- join: In addition to fork guarantees, parent may not run any operations until joined child is completed.

For further context:

https://github.com/thefrontside/effection.js/issues/70

Open Questions
--------------

- There is probably a way to model the internal continuations semi-monadically so that we don't have to store state like required   children and `waiting ` and what not on the context, and so that in   `monitor` and `fork` and other control functions, we don't have to   fiddle with that state, but instead allow the control function   itself to somehow pass control to the "next" thing, which naturally   unfolds the computation. If it sounds hand-wavy, it is, but until   that elegant solution is found, we can start with a hackier one.
- We will need some way to "re-direct" the continuation of a context   in the case of joining a fork. In other words, if we join a fork,   something else needs to happen when it continues, not what happens when it was originally conceived.

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -11,27 +11,19 @@ export class ExecutionContext {
 
   get isBlocking() { return this.isRunning || this.isWaiting || this.isUnstarted; }
 
-  get hasBlockingChildren() {
-    for (let child of this.children) {
-      if (child.isBlocking && child.isRequired) {
-        return true;
-      }
-    }
-    return false;
-  }
-
-  constructor(parent = undefined, isRequired = true) {
+  constructor(parent = undefined, continuation = x => x) {
     this.id = this.constructor.ids++;
     this.parent = parent;
     this.children = new Set();
+    this.requiredChildren = new Set();
     this.exitHooks = new Set();
     this.state = 'unstarted';
-    this.isRequired = isRequired;
     this.mailbox = { messages: new Set(), receivers: new Set() };
     this.resume = this.resume.bind(this);
     this.fail = this.fail.bind(this);
     this.ensure = this.ensure.bind(this);
     this.spawn = this.spawn.bind(this);
+    this.continue = continuation.bind(null, this);
   }
 
   get promise() {
@@ -73,15 +65,9 @@ export class ExecutionContext {
     }
   }
 
-  spawn(operation, isRequired) {
-    let child = new ExecutionContext(this, isRequired);
+  spawn(operation, continuation = x => x) {
+    let child = new ExecutionContext(this, continuation);
     this.children.add(child);
-    child.ensure(() => {
-      this.children.delete(child);
-      if (this.isWaiting && !this.hasBlockingChildren) {
-        this.finalize('completed');
-      }
-    });
     child.enter(operation);
     return child;
   }
@@ -125,10 +111,10 @@ Thanks!`);
   }
 
   resume(value) {
-    if (!this.isRunning) { return; }
-
-    this.result = value;
-    if (this.hasBlockingChildren) {
+    if (this.isRunning) {
+      this.result = value;
+    }
+    if (this.requiredChildren.size > 0) {
       this.state = 'waiting';
     } else {
       this.finalize('completed', value);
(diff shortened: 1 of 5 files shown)


frontsidejack:
A preview package of this pull request has been released to NPM with the tag `strict-teardown-order`.
```bash
$ npm install effection@strict-teardown-order
```
```bash
{
  "effection": "strict-teardown-order"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-1b42438](https://www.npmjs.com/package/effection/v/0.4.0-1b42438) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against af7dd8f6a7b0dd563d3c4e7ed4a105c34b903a49
</p>

frontsidejack:
A preview package of this pull request has been released to NPM with the tag `strict-teardown-order`.
```bash
$ npm install effection@strict-teardown-order
```
```bash
{
  "effection": "strict-teardown-order"
}
```
Once the branch associated with this tag is deleted (usually once the PR is merged or closed), it will no longer be available. However, it currently references [effection@0.4.0-2f073e9](https://www.npmjs.com/package/effection/v/0.4.0-2f073e9) which will be available to install forever.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e3129b7207ca2117df843755721661dd8b4541d6
</p>

jnicklas:
🎉🎉🎉 Let's get this merged and upstreamed ASAP

PR #84, opened 2020-03-10
Setup monorepo and move effection packages from bigtest

minkimcello:
## Motivation
We're converting `thefrontside/effection.js` into a monorepo in order to migrate some (or all) of the [`bigtest/effection`](https://github.com/thefrontside/bigtest/tree/master/packages/effection/src) packages over here.

## Approach
- Moved current effection package into the `package/` directory.
- Configured `package.json` to use `yarn workspaces`.
  - Tested by `yarn install`ing and `yarn test`ing from root directory
- Updated `release` and `preview` workflows to the latest versions to be compatible with monorepos
  - ⚠️ One small caveat (mentioned below) regarding the publish script for the current effection package.
  - ⭐️ Added test workflow too.

## Todo
- [x] @cowboyd Which packages from bigtest/packages/effection do we need to move over?
  - ✅ No need to do now.
- [x] The original effection package ran `npm run pack:publish` script but we may not want to run that same script for all the other incoming packages of this new monorepo... or set up all future packages to use the same `pack` and `publish` process. 🤷‍♂. 
  - ✅ Will move to `microbundle` after this PR.
- [x] Should we remove `.circleci/` and configure testing from github actions instead?
  - ✅ Also later.


jnicklas:
We'll probably want to split up what is in the `bigtest/effection` package into some smaller packages. For example the `Mailbox` should probably be in a `@effection/mailbox` package, but IMO there is no need to make these changes at this point in time, we can always add these extractions later.

cowboyd:
Also from the get-go, I'd like to split up the tests for each package to run in their own job. This has been a pain-point in the bigtest repo where it would be much nicer if each package ran its tests in parallel.

Let's also keep everything in CircleCI for now. We can migrate to GH actions with a follow-on.

As for the whole `pack:publish` script, this is because we use Pika to build effection. I actually think we want to move away from pika in favor of `microbundle` precisely so that the build isn't so weird. Let's do that after this.

cowboyd:
Also, I think I'd like to take the `.js` off the name of the repo, but that can also come later.

minkimcello:
@cowboyd 

> Also from the get-go, I'd like to split up the tests for each package to run in their own job. This has been a pain-point in the bigtest repo where it would be much nicer if each package ran its tests in parallel.

We'll also do this in a separate PR after this one? To run the tests in parallel I think we'll just need to create separate jobs in a test workflow for each package because I believe the jobs run synchronously.

We could tell the job to `cd` into each package and run the tests or create an action that detects if there has been any changes from its commit to the previous to decide if it it should run tests or not to prevent taking up unnecessary bandwidth.

.circleci/config.yml
```diff
@@ -46,8 +46,8 @@ workflows:
...
-  test-types:
-    jobs:
-      - run:
-          name: "test types"
-          script: test:types
+  # test-types:
+  #   jobs:
+  #     - run:
+  #         name: "test types"
+  #         script: test:types
```
cowboyd: We definitely don't want to disable the test:types Why was it failing?

cowboyd:
>We could tell the job to cd into each package and run the tests or create an action that detects if there has been any changes from its commit to the previous to decide if it it should run tests or not to prevent taking up unnecessary bandwidth.

I think it's simpler to just run all the tests in parallel. This should be both simple and fast.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0-monorepo",
   "description": "Effortlessly composable structured concurrency primitive for JavaScript",
   "repository": "git@github.com:thefrontside/effection.js.git",
-  "author": "Taras Mankovski <taras@frontside.io>",
+  "author": "Charles Lowell <cowboyd@frontside.io>",
```
cowboyd: Let's change this to `"Frontside Engineering <engineering@frontside.io>"`

cowboyd:
> > We could tell the job to cd into each package and run the tests or create an action that detects if there has been any changes from its commit to the previous to decide if it it should run tests or not to prevent taking up unnecessary bandwidth.
> 
> I think it's simpler to just run all the tests in parallel. This should be both simple and fast.

@minkimcello Ah, I misunderstood.... I think I see what you're saying.... for example the `test:types` job would:

```
cd packages/effection && yarn test:types
```

In that case, yes, that is what we should do

.circleci/config.yml
minkimcello: I'm getting this error when I run `yarn test:types`:

```
yarn run v1.21.1
$ yarn workspaces run test:types

> effection
$ dtslint types --localTs node_modules/typescript/lib --expectOnly

Error: Errors in typescript@local for external dependencies:
../index.d.ts(3,15): error TS2300: Duplicate identifier 'Operation'.
../index.d.ts(5,8): error TS2300: Duplicate identifier 'OperationFn'.
../index.d.ts(7,8): error TS2300: Duplicate identifier 'Controller'.

    at /Users/minkimcello/Projects/thefrontside/effection.js/node_modules/dtslint/bin/index.js:193:19
    at Generator.next (<anonymous>)
    at fulfilled (/Users/minkimcello/Projects/thefrontside/effection.js/node_modules/dtslint/bin/index.js:6:58)
```

minkimcello:
> > > We could tell the job to cd into each package and run the tests or create an action that detects if there has been any changes from its commit to the previous to decide if it it should run tests or not to prevent taking up unnecessary bandwidth.
> > 
> > 
> > I think it's simpler to just run all the tests in parallel. This should be both simple and fast.
> 
> @minkimcello Ah, I misunderstood.... I think I see what you're saying.... for example the `test:types` job would:
> 
> ```
> cd packages/effection && yarn test:types
> ```
> 
> In that case, yes, that is what we should do

I think we're talking about two different things now?

So for `yarn test:types`, we could either run `yarn workspaces run test:types`  or `cd packages/effection && yarn test:types` depending on if the other packages will have the same script. (See reply above in the changes-requested for why it's failing).

For running tests in parallel, how were you picturing we do that?

scenario 1:
```
on: push
jobs:
  tests:
    steps:
    - run: yarn workspaces run test
      // wouldn't this run asynchronously and what we're trying to upgrade from?
    - run: yarn test:package-1 // cd packages/package-1 && yarn test
    - run: yarn test:package-2 // cd packages/package-2 && yarn test
      // this is equivalent to yarn workspaces run test
```
scenario 2:
```
on: push
jobs:
  package-1-test:
    steps:
    - run: cd packages/package-1 && yarn test
  package-2-test:
    steps:
    - run: cd packages/package-2 && yarn test
  package-3-test:
    steps:
    - run: cd packages/package-3 && yarn test
  // this SHOULD run synchronously
```
what I was saying could be a possibility in the future once we do the javascript-refactor:
```
on: push
jobs:
  package-1-test:
    steps:
    - uses: thefrontside/actions/run-tests@v1
      with: 
        package: packages/package-1
  package-2-test:
    steps:
    - uses: thefrontside/actions/run-tests@v1
      with: 
        package: packages/package-2
  package-3-test:
    steps:
    - uses: thefrontside/actions/run-tests@v1
      with: 
        package: packages/package-3
// `run-tests` action would check commit with previous commit to see if the directory of `package` argument has any changes to determine whether we should run the test for that package.
```

cowboyd:
We want scenario 2, and while scenario 3 sounds nice, we probably just want to run all the tests in parallel for every build for now because some packages depend on others, and so we need to know that they'll pass too if their underlying dependencies change.

So the job (which can be renamed to test-effection-types would `cd packages/effection` and run `yarn test:types`

minkimcello:
@cowboyd Could you help me find out why the test is failing? https://github.com/thefrontside/effection.js/pull/84/checks?check_run_id=498405243

And also since we have the test workflow now, is it okay to remove CircleCI?

cowboyd:
> @cowboyd Could you help me find out why the test is failing? https://github.com/thefrontside/effection.js/pull/84/checks?check_run_id=498405243
> 
> And also since we have the test workflow now, is it okay to remove CircleCI?

Let's try and limit the scope of the change and run all the tests on circle, not GitHub Actions for now since we know it was working there.

minkimcello:
@cowboyd We're getting the same error in circleci: https://circleci.com/gh/thefrontside/effection.js/357?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

.circleci/config.yml
taras: @minkimcello you should never disable tests unless you got everyone's buy-in to do it.

minkimcello: @taras oh okay

minkimcello: @cowboyd Should I have not resolved this myself?

jbolda:
It may be easier to manage the parallel runs with a matrix.
```
on: push
jobs:
  test:
    strategy:
      fail-fast: false
      matrix:
        package: ['package-1', 'package-2', 'package-3']
    steps:
      - run: yarn workspace packages/${{ matrix.package}} test
```

Issue #86, opened 2020-03-12
Returning with an associated supended context makes it outlive its scope


jnicklas:
Sorry for the utterly confusing title, it's the best I could come up with.

This isn't really a bug in Effection, but rather a consequence of some of the design patterns we've started to adopt. Since I ran into this a few times and it is potentially very confusing, I thought it best to actually document this somewhere so we can try to find a solution to it.

The problem occurrs if we're using the RAII-like pattern of creating some value object which has an implicitly created context associated for it, or an ensure block. For example, we might create an http server (value) with an associated ensure block which closes the server (context), like this:

``` typescript
function *createServer() {
  let server = new Server();
  yield suspend(ensure(() => server.close());
  return server;
}
```

In  some other context we could then use the server like this:

``` typescript
function *main() {
  let server = createServer();
  yield fork(function*() {
    server.doSomething();
  });
}
```

The structured concurrency guarantees ensure that the server won't be closed before any fork in the `main` function has completed. So we have effectively bound the server to the scope if was created in, and ensured it will be de-allocated at the end of this scope.

Now let's imagine we want to create a special kind of server, and we want to re-use the `createServer` primitive:

``` typescript
function *createSpecialServer() {
  let server = yield createServer();
  doSomethingSpecial(server);
  return server;
}
```

Now we have a problem! The server returned from `createServer` is bound to the scope it was created in, which is the `createSpecialServer` scope, and the structured concurrency guarantees ensure that it won't close until that scope is complete.

BUT at the end of the scope we are RETURNING the server, and effectively passing it UP to the parent. But we just noted that the server will close at the end of the scope!

``` typescript
function *createSpecialServer() {
  let server = yield createServer();
  doSomethingSpecial(server);
  return server; // close will run here!!!
}
```

So what we return from `createSpecialServer` is a closed server!

Okay, so we realize that we need to hoist the server up one scope, maybe we can do this:

``` typescript
function *createSpecialServer() {
  let server = yield suspend(createServer());
  doSomethingSpecial(server);
  return server;
}
```

Unfortunately this doesn't work, since `suspend` will return a `Context` and not the server, so if we use `suspend` we can't actually get the value!

So fundamentally we have a problem in that this pattern simply does not compose, and when it doesn't compose it fails in a surprising, and (as should be evident by this lengthy explanation) pretty difficult to understand way.

As I said this isn't a bug in Effection per se, but it is something we need to think about as we're trying to find recommended practices in how to use Effection.

[...]

PR #87, opened 2020-03-27
Allows contexts to escape to parent via return

jnicklas:
When a context returns another context, that context is kept alive. Meaning that unlike other children, it won't be halted when the context exits, it is then linked to the parent. This provides a desired escape hatch to allowing contexts escape the scope in which they were created.

Initially we used the `suspend` pattern for this purpose, but as documented in #86, this pattern does not compose well, and its behaviour is confusing at times.

In the future, we might extend this functionality with what we have dubbed resources, where a resource basically bundles some JavaScript object, and an associated context. This PR does not implement such a feature yet.

This PR also makes a few other changes which were necessary or convenient for implementing this feature:

- execution contexts now track whether they are required. This simplifies things, since we want to preserve whether they are required when they are returned

- execution contexts are not initialized with a parent, but rather they can be linked to another context through `link` and `unlink`. This allows for more flexibility

- linked children always propagate errors to their parents. Instead of this being part of the functionality of spawn and fork. This does require some trickery to make generator work, but I think it is worth it.

- `monitor` has been renamed to `spawn` (an alias is retained for now), because it does not add any functionality on top of `context.spawn`.

- It is currently no longer supported to use a `fork` or `spawn` as the top level operation. Because it hurts my brain and I can't figure out what sensible behaviour actually should be.

### DIFF
--- a/packages/effection/src/context.js
+++ b/packages/effection/src/context.js
@@ -11,18 +11,16 @@ export class ExecutionContext {
 
   get isBlocking() { return this.isRunning || this.isWaiting || this.isUnstarted; }
 
-  constructor(parent = undefined, continuation = x => x) {
+  constructor(isRequired = false) {
     this.id = this.constructor.ids++;
-    this.parent = parent;
+    this.isRequired = isRequired;
     this.children = new Set();
-    this.requiredChildren = new Set();
     this.exitHooks = new Set();
     this.state = 'unstarted';
     this.resume = this.resume.bind(this);
     this.fail = this.fail.bind(this);
-    this.ensure = this.ensure.bind(this);
     this.spawn = this.spawn.bind(this);
-    this.continue = continuation.bind(null, this);
+    this.ensure = this.ensure.bind(this);
   }
 
   get promise() {
@@ -64,9 +62,9 @@ export class ExecutionContext {
     }
   }
 
-  spawn(operation, continuation = x => x) {
-    let child = new ExecutionContext(this, continuation);
-    this.children.add(child);
+  spawn(operation, required = false) {
+    let child = new ExecutionContext(required);
+    this.link(child);
     child.enter(operation);
     return child;
   }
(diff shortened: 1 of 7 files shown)


packages/effection/src/context.js
```diff
@@ -113,7 +112,7 @@ Thanks!`);
     if (this.isRunning) {
       this.result = value;
     }
-    if (this.requiredChildren.size > 0) {
+    if (Array.from(this.children).some((c) => (!this.parent || c !== value) && c.isRequired)) {
```
jnicklas: This conditional is quite complicated. Not sure how to reformulate this in a better way though.

packages/effection/src/context.js
```diff
@@ -148,14 +149,43 @@ Original error:`);
       }
     }
 
+    if(this.result instanceof ExecutionContext) {
```
jnicklas: Is `instanceof` the right choice here?

packages/effection/src/control.js
```diff
@@ -98,9 +104,6 @@ export const GeneratorControl = generator => ControlFunction.of(self => {
             }
           }
         });
-        if (child.isBlocking) {
-          self.context.requiredChildren.add(child);
-        }
```
jnicklas: I don't really understand this conditional or what it is supposed to do.

packages/effection/tests/order-of-operations.test.js
```diff
@@ -24,17 +24,15 @@ describe('Order of Shutdown Operations', () => {
   beforeEach(() => {
     order = [];
 
-    let self = ({ resume, context: { parent }}) => resume(parent);
```
jnicklas: The reason I had to rewrite this is because returning the `parent` here actually transfers ownership of it now 😱

packages/effection/tests/execution.test.js
```diff
@@ -76,40 +76,4 @@ describe('execution', () => {
...
-      it('fails the process', () => {
-        expect(exec.isErrored).toBe(true);
-      });
-      it('has the error as the result', () => {
-        expect(exec.result).toBeDefined();
-        expect(exec.result.message).toEqual('boom!');
-      });
-    });
-  });
 });
```
jnicklas: For some reason figuring out how to make this work really hurt my brain. But then I realized that maybe this just shouldn't be allowed as a pattern? Ideally it would just work, but it turned out to be surprisingly tricky.

jnicklas:
Closing this in favour of #89, since I did a ton of additional work which also fixed a lot of bugs in this PR.

PR #89, opened 2020-03-30
Implement support for resources 

jnicklas:
A resource ties an execution context to an object. This allows an object to be passed around between contexts while retaining an associated context. This context is "kept alive" as long as the object's scope is alive. If the object is returned out of the current scope, the context is hoisted to the parent scope along with the object. This allows an escape hatch from contexts being local to where they were created and makes it possible to implement the RAII pattern.

### DIFF
--- a/packages/effection/src/context.js
+++ b/packages/effection/src/context.js
@@ -13,9 +13,10 @@ export class ExecutionContext {
 
   get isBlocking() { return this.isRunning || this.isWaiting || this.isUnstarted; }
 
-  constructor(isRequired = false) {
+  constructor({ isRequired = false, allowContextReturn = true } = {}) {
     this.id = ids++;
     this.isRequired = isRequired;
+    this.allowContextReturn = allowContextReturn;
     this.children = new Set();
     this.exitHooks = new Set();
     this.state = 'unstarted';
@@ -66,14 +67,14 @@ export class ExecutionContext {
   }
 
   spawn(operation) {
-    let child = new ExecutionContext(false);
+    let child = new ExecutionContext({ isRequired: false });
     this.link(child);
     child.enter(operation);
     return child;
   }
 
   fork(operation) {
-    let child = new ExecutionContext(true);
+    let child = new ExecutionContext({ isRequired: true });
     this.link(child);
     child.enter(operation);
     return child;
@@ -124,7 +125,7 @@ Thanks!`);
           this.link(contextOf(value));
         }
       }
-      if(Array.from(this.children).every((c) => (this.parent && c === value) || !c.isRequired)) {
+      if(Array.from(this.children).every((c) => (this.allowContextReturn && c === value) || !c.isRequired)) {
         this.finalize('completed', value);
       } else {
         this.state = 'waiting';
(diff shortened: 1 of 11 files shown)


jnicklas:
Just a note that this is definitely a breaking change, did find a few places in bigtest where this was breaking things in subtle ways. Mostly where we were returning a context for no real reason, which is now doing things when previously it was just useless.

jnicklas:
@cowboyd marking this as ready for review :)

packages/effection/src/context.js
```diff
@@ -110,52 +118,97 @@ Thanks!`);
...
+      this.fail(child.result);
+    } else if (this.isWaiting && Array.from(this.children).every((c) => !c.isRequired)) {
+      this.finalize('completed');
+    }
+  }
+
+  link(child) {
+    if(this.id === child.id) {
+      throw new Error('cannot link context to itself');
+    }
```
cowboyd, line 192: Is it worth adding in a check here to make sure that the new parent is higher in the tree than the old parent? It feels like it is not ok, to transfer ownership "down-tree", although we may want a way to do that explicitly in the future.

packages/effection/src/index.js
```diff
@@ -1,10 +1,13 @@
 export { timeout } from './timeout';
-export { fork, join, monitor } from './control';
+export { fork, join, spawn, spawn as monitor } from './control';
```
cowboyd, line 2: We should mark monitor as deprecated, maybe emit a console warning once.

packages/effection/src/context.js
jnicklas: That was sort of what I was thinking, I think it makes sense to expose `link` as an operation, to essentially "move" a child somewhere, though there might be some weird implications to that which I haven't thought through, so I didn't add it in this PR. But we could definitively add such a check for now, until we're sure that it is actually safe to do this.

jnicklas:
>     1. when should the `create` be considered done?

A resource starts off in the context where it was created, and it effectively functions like a child of this context, so when the context is finalized the resource is halted.

The one exception to this is when the child resource is returned out of the context. Any child resource or child context which is returned from a context is not halted on finalization.*

Additionally, a resource can exit early, that is if the resource's context becomes completed. This is a trivial example:

``` javascript
yield resource({}, function*() { return });
```

The associated context here does nothing, so there is nothing keeping it alive. Compare to this:

``` javascript
yield resource({}, function*() { yield });
```

I've actually thought about whether a resource which finishes early should be considered an error, as this is probably never what you want, and it's actually already bitten me once or twice, but I felt that that's maybe a tad bit too opinionated 🤷‍♂️
 
>     2. is binding the resource to the current context necessary when the only purpose is to return it to the parent context?

Unfortunately not, since everything in Effection is synchronous, whatever happens up until the first yield point will run synchronously on creation of the resource. An error which occurs here will need to be handled by ... someone. So by creating the resource in the current context, and then moving it upwards, we ensure that there is someone there who can handle this error.

\* (except, and here is where things get a bit weird: if the context that returns the resource is itself a resource context or it is the root context, some other weird rules apply, which feel like they make sense, but actually kinda don't 😂. But even though the implementation differs slightly here, functionally it feels like it behaves about the same)

jnicklas:
One thing that I  just realized is that we should consider a guard against returning a context or resource which is NOT a child of the current context, while this probably works just fine, it feels like it violates the "spirit" of this API, and shouldn't be allowed.

PR #95, opened 2020-04-16
Add `Deffered` an alternative to explicit control functions

cowboyd:
Motivation
-----------

One of the ongoing discussions we've been having is that we want to move away from control functions and really focus on having generators and promises be the base primitives that people work with in effection.

Approach
----------

As a baby step towards that goal, this introduces a `Deferred` function that returns an object containing a `resolve` / `reject` / `promise` calling `resolve()` will resolve `promise`, and calling `reject`, will reject `promise`.

This provides a simple control pattern very similar to the exsting control functions, but based on standard and familiar apis: promise and generator.

> Ok, generators may not feel familiar, but they are at least "familiar-ish" in the sense that they're just JS code.

Thus, we can re-write any control function to be `Deferred` + `try/finally` with a yield to the futures promise.

As an example, `timeout` is written in this way.

### DIFF
--- a/packages/effection/src/deferred.js
+++ b/packages/effection/src/deferred.js
@@ -1,0 +1,23 @@
+/**
+ * Create an object that holds a promise along with the control
+ * functions needed to control it:
+ *
+ *   let { resolve, reject, promise } = Deferred();
+ *
+ *   promise.then(num => console.log('your number is ', 10);
+ */
+export function Deferred() {
+  let resolve;
+  let reject;
+
+  let promise = new Promise((_resolve, _reject) => {
+    resolve = _resolve;
+    reject = _reject;
+  }) ;
+
+  return {
+    resolve,
+    reject,
+    promise
+  };
+}
--- a/packages/effection/src/index.js
+++ b/packages/effection/src/index.js
@@ -1,3 +1,4 @@
+export { Deferred } from './deferred';
 export { timeout } from './timeout';
 export { fork, join, spawn, spawn as monitor } from './control';
 
--- a/packages/effection/src/timeout.js
+++ b/packages/effection/src/timeout.js
@@ -1,3 +1,4 @@
+import { Deferred } from './deferred';
 /**
 * Create an execution controller that resumes after the specified
 * duration. E.g.
@@ -7,9 +8,12 @@
 *     console.log(`Hello ${target}!`);
 *   }
 */
-export function timeout(duration) {
-  return ({ resume, ensure }) => {
-    let timeoutId = setTimeout(() => resume(), duration);
-    ensure(() => clearTimeout(timeoutId));
-  };
+export function* timeout(duration) {
+  let { resolve, promise } = Deferred();
+  let timeoutId = setTimeout(resolve, duration);
+  try {
+    yield promise;
+  } finally {
+    clearTimeout(timeoutId);
+  }
 }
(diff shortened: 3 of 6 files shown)


jnicklas:
In terms of terminology, this kind of constructor is usually called a "deferred" object in JS, it even used to be part of the promise API

https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm/Deferred
https://api.jquery.com/category/deferred-object/

So that might be a better name than future, considering there is so much precedent for it.

That being said, while the idea of using promises as the sole leaf type and abandoning controller functions entirely is very appealing, I also realized that there's a fundamental problem with it, in that currently we rely very heavily on the fact that operations *can* be synchronous. We use this extensively for example to set up event listeners. Suppose for example that we have two mailboxes, both of which subscribe to some event source:

``` javascript
let mailbox1 = yield Mailbox.subscribe(source1, "some-event");
let mailbox2 = yield Mailbox.subscribe(source2, "other-event");
```

If yield points are always async (as they would be if we used promises only), then there is no way of setting up both of these listeners without introducing a race condition. Here "other-event" could theoretically occur in between the first listener being set up and the second listener.

It's kind of a challenge of effection that yield points are *sometimes* async, but not always. It honestly bugs me a bit, but I don't see a good way of resolving this either.

cowboyd:
These are good points. I'd forgotten about good old `Deferred` although while there is precedent, there is also precedent for associating it with "being deprecated" or "obsolete". Not sure which way to go there 

> If yield points are always async (as they would be if we used promises only), then there is no way of setting up both of these listeners without introducing a race condition. Here "other-event" could theoretically occur in between the first listener being set up and the second listener.
>
> It's kind of a challenge of effection that yield points are sometimes async, but not always. It honestly bugs me a bit, but I don't see a good way of resolving this either.

🤔 🤔 🤔 

Yeah, it's a tough one and I see the conundrum. This really exposes a fundamental weakness in the effection story though, right? I mean, without any prior knowledge about the underlying implementation, shouldn't the "reasonable developer"  naturally assume that all yield points were async? There's certainly nothing marking which ones are async and which ones are sync, and so how could you possibly reason about race conditions without something explicit (unless your name was Jonas or Charles)? 

I feel like no matter what, we need some way to explicitly synchronize operations so that anybody can be sure of avoiding race conditions. Because sure we can do it today, but only really by a happy accident of implementation, and certainly not in any way that is easy to communicate to potential future effection users.... Here's a non-exhaustive list off the top of my head:

1. some sort of transaction api that you could thread through operations to have their pieces that need to be synchronized happening in a single tick of the runloop:

```js
let transaction = yield createTransaction([
  Mailbox.subscribe('event-one'),
  Mailbox.subscribe('event-two')
]);
let [one, two] = yield transaction;
```

It looks a bit awkward, but on the other hand, it clearly signals that race conditions are possible and forces you to acknowledge them.

2. Add a different type of operations that are explicitly sync , e.g. `type Operation = Promise | Generator | Atomic`. `Atomic` ops would get passed something similar to today's  control function, but we would verify that `resume` or `fail` was called on the _same_ tick of the run loop. It would immediately throw an error if they were not called synchronously. So you could not, for example implement `timeout` with an `Atomic`, but you could implement `fork`

3. Assume that if you need to do something synchronously, you should do it in the body of your generator in between yield points. Those are natural atomic ops after all. In other worlds, Is this the platform telling us that subscribe should not be an operation, but that the api should be something different instead?

jnicklas:
Closing this, due to lack of activity. Effection 2 uses a Deferred implementation internally, so we could ship it as part of the package if we want to, but I'm not 100% convinced of the necessity.

PR #99, opened 2020-04-23
Add event package for working with events

jnicklas:
Working with event emitters has been a really common need when working with Effection. Over time we evolved some helpers for making this more efficient.

We have also extensively used mailboxes for working with events. However, with the advent of resources, it is now pretty easy to implement and efficient and easy to use event iterator. I've called this operation `on`. This is a more light-weight solution than mailboxes. There is some overlap in functionality, but I think this is acceptable. Basically we could replace a lot of the mailboxes in bigtest with this, and it would make it clearer where we actually need mailboxes and where we just use them to subscribe to events.


jnicklas:
There is an example of the new `on` operation in the [README](https://github.com/thefrontside/effection.js/blob/a2ebf310414e034c76a017d455852ab9b5def0e6/packages/events/README.md).

jnicklas:
It seems that NodeCodeFormatter stripped a ton of dependencies from `yarn.lock` which seems... not good?

package.json
```diff
@@ -8,7 +8,9 @@
   "private": true,
   "workspaces": {
     "packages": [
-      "packages/*"
+      "packages/effection/pkg",
```
jnicklas, line 11: Had to do this to be able to use the main effection package from `events`, not sure if this is correct or if there is a better solution?

cowboyd: Yup, until we use yarn2 which _should_ resolve internal packages in dependency order, this is the only way I know. /cc @taras

packages/events/README.md
```diff
@@ -0,0 +1,59 @@
...
+Node.js style EventEmitters (`on`/`off`) and with browser style EventTarget
+(`addEventListeners`/`removeEventListener`).
+
+## Usage
+
+The `once` operation can be used to wait for an event to occur exactly once.
+
+``` typescript
+import { main } from '@effection/node';
+import { once } from '@effection/node';
```
cowboyd, line 13:
```suggestion
import { once } from '@effection/events';
```

packages/events/README.md
```diff
@@ -0,0 +1,59 @@
...
+  yield once(window, "DOMContentLoaded");
+  console.log('Hello World!');
+});
+```
+
+The `on` operation returns a subscription which can be used to loop over events:
+
+``` typescript
+import { main } from '@effection/node';
+import { on } from '@effection/node';
```
cowboyd, line 25:
```suggestion
import { on } from '@effection/events';
```

packages/events/README.md
```diff
@@ -0,0 +1,59 @@
...
+It is a common pattern, primarily in Node to have an event called `error` which
+has an error as its first argument. `@effection/events` includes a helper to
+make handling this situation convenient:
+
+``` typescript
+import { main } from '@effection/node';
+import { once, throwOnErrorEvent } from '@effection/node';
+
+main(function* sayHello() {
+  let someSocket();
```
cowboyd, line 46:
```suggestion
  let someSocket = createSocket();
```

packages/events/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+  yield throwOnErrorEvent(someSocket);
+
+  yield once(someSocket, "open");
+
+  let messages = yield once(someSocket, "message");
+
+  while(true) {
+    let message = yield someSocket.next();
```
cowboyd, line 55:
```suggestion
    let message = yield messages.next();
```

packages/events/src/on.ts
```diff
@@ -0,0 +1,35 @@
...
+      removeListener(this.source, this.eventName, listener);
+    }
+
+  }
+}
+
+
+export function* on<T extends Array<unknown> = unknown[]>(source: EventSource, name: string): Operation<Subscription<T>> {
+  let subscription = new Subscription<T>(source, name);
+  return yield resource(subscription, subscription.subscribe());
```
cowboyd, line 34: very nice 😍

cowboyd:
> It seems that NodeCodeFormatter stripped a ton of dependencies from `yarn.lock` which seems... not good?

It says very clearly in our terms of service that bots may not drink on the job. We may have to fire NodeCodeFormatter.

PR #100, opened 2020-04-23
Replace pika with microbundle

jnicklas:
Pika doesn't play well with the monorepo architecture that we have. Because it has its own `package.json` file in `pkg`, the main effection package is essentially unusable from other packages in the repo. With microbundle we don't have this problem.


cowboyd:
I ran this against the bigtest test suite which was all green

PR #101, opened 2020-04-24
Add helpers for working with child processes

jnicklas:
This adds helpers for working with child processes which automatically clean up after themselves.


cowboyd:
Is there a worry that these will inconveniently clash with our existing effection primitives `spawn` and `fork` ?

cowboyd:
We'd like to be able to import the modules and namespaces in a similar way to the way node does it:

instead of 

```ts
import { fork } from '@effection/node'
```

which conflicts with `fork` in `effection`

we'd like to be able to import `fork` from child process in "namespaced" way. Something like:

```ts
import * as ChildProcess from '@effection/node/child_process'

export function * echo() {
  yield ChildProcess.fork("echo", ["Hello World"]);
}
```

And also be able to access the named export directly:

```ts
import { fork } from '@effection/node/child_process';
```

or, import the ChildProcess namespace from `@effection/node`:

```ts
import { ChildProcess } from `@effection/node`
```

PR #110, opened 2020-04-27
Export function of child_process

minkimcello:
## Motivation
[This](https://github.com/thefrontside/effection.js/pull/101#issuecomment-620168884) comment.

## Approach
I think I may be underestimating/misunderstanding the issue but can't we just do: 
```ts
// > packages/node/src/index.ts

export * as ChildProcess from './child_process';
``` 

This seems to allow us to:
```ts
import { ChildProcess } from `@effection/node`;

export function * echo() {
  yield ChildProcess.fork("echo", ["Hello World"]);
}
``` 

Not knowing how `effection` works I'm not sure how I can test to make sure it doesn't conflict with `@effection`'s `fork` but hovering over the imported functions seems to display the intended result:

Screen Shot 2020-04-27 at 3 37 43 PM

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -1,3 +1,2 @@
 export { main } from './main';
-export { spawn, fork } from './child_process';
 export * as ChildProcess from './child_process';
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -1,2 +1,3 @@
 export { main } from './main';
 export { spawn, fork } from './child_process';
+export * as ChildProcess from './child_process';


cowboyd:
What about the other forms?

```ts
import { spawn } from `@effection/node/child_process`
```

```ts
import * as ChildProcess from `@effection/node/child_process`
```

minkimcello:
@cowboyd I believe both of those forms work if you specify the `dist` directory in that path.

Screen Shot 2020-04-27 at 6 51 07 PM

```ts
// > node/src/index.ts
export { spawn, fork } from './child_process';
```
:point_up: Was this conflicting with `fork` from `@effection`?
Because when it's exported and I import it as:
```ts
import { fork, spawn } from '@effection/node';
```
It looks to be importing the correct function.

cowboyd:
> I believe both of those forms work if you specify the dist directory in that path.

Is there a way to do it without having to put `dist/` in the middle?

minkimcello:
@cowboyd Went down a bit of a rabbit hole and found these issues: [#10866](https://github.com/microsoft/TypeScript/issues/10866), [#26722](https://github.com/microsoft/TypeScript/issues/26722), and [#33079](https://github.com/microsoft/TypeScript/issues/33079). I think these people are mentioning what it is that we want to do with `@effection/node/child_process`.

Could we use something like [`module-alias`](https://www.npmjs.com/package/module-alias)?

cowboyd:
Based on the last ticket, It seems like for node > 13 we can use package exports coupled with tsconfig paths. But maybe we can add that later?

We got burned with module-alias before so not keen to go down that road,

cowboyd:
@minkimcello Thanks for going down these rabbit holes, otherwise known as "research." This is exactly the information that we need.

PR #119, opened 2020-05-14
Fixed sourcemaps for events and node packages

taras:
## Motivation

We saw strange errors while installing Effection from BigTest. One of the problems was that sourcemaps were not working. This was happening because sourcemaps referenced ts file which were not being included in the package.

I fixed a bunch of small problems that I came across.

## Approach

1. Bumped microbundle to 0.12.0
2. Changed `typings` entry point to `types` in @effection/events and @effection/node packages
3. Switched to using @frontside/tsconfig

### DIFF
--- a/packages/events/src/on.ts
+++ b/packages/events/src/on.ts
@@ -2,6 +2,10 @@ import { resource, Operation } from 'effection';
 import { EventSource, addListener, removeListener } from './event-source';
 import { once } from './once';
 
+function isNotEmpty<T>(array: T[]): array is { shift(): T; } & Array<T> {
+  return array.length > 0;
+}
+
 export class Subscription<T extends Array<unknown>> {
   private events: T[] = [];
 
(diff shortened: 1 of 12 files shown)


packages/events/src/on.ts
```diff
@@ -2,14 +2,18 @@ import { resource, Operation } from 'effection';
 import { EventSource, addListener, removeListener } from './event-source';
 import { once } from './once';
 
+function isNotEmpty<T>(array: T[]): array is { shift(): T; } & Array<T> {
```
cowboyd, line 5: I wonder if you could you type this as `array is [T,...T[]]`? as perhaps more directly typing it as "an array of at least one element" rather than an array whose shift method defiintely returns an item.

Also, can we move this helper function to the bottom of the file?

cowboyd: I went ahead and updated it to use vanilla destructuring instead of `shift()` which seems to require no annotations at all.

PR #120, opened 2020-05-16
Use tsdx to build without transpiling generators

taras:
## Motivation

We made a decision to ship generators in our distribution bundles because IE11 compatibility is not important to us. It's surprisingly difficult to get this to work. We tried using microbundle but that turned out to be even more complicated because their [modern and cjs formats have mutually conflicting configuration](https://github.com/developit/microbundle/issues/618).

## Approach

I tried `tsdx` but it also transpiles generators by default. There is an [RFC to make tsdx](https://github.com/jaredpalmer/tsdx/issues/634) more customizable but in the mean time I patched `tsdx` to exclude `@babel/plugin-transform-regenerator` from Babel plugins using [patch-package](https://www.npmjs.com/package/patch-package).

1. Added patch-package as a dependency in root and run `patch-package` on `postinstall`
2. Added @babel/preset-modules to effection dependencies because for some reason it started complaining that it was missing (it might be the result of an upgrade that was caused by tsdx installation)
3. Updated events to prepack with tsdx
4. Added module entry to events because it's also used in the web
5. Updated node to prepack with tsdx
6. tsdx uses rollup under the hood which requires that modules use ES6 syntax instead of commonjs. So I added module: "esnext" to events and node tsconfig.

## Learning

patch-package is pretty cool. It has a nice workflow. When you need to change a dependency, modify it's code in node_modules and run `npx patch-package <package_name>`. It'll compare a fresh copy of the package that you're modifying to your changed copy and created a diff. It'll automatically apply this diff when you run `yarn`.


patches/tsdx+0.13.2.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/node_modules/tsdx/dist/babelPluginTsdx.js
++++ b/node_modules/tsdx/dist/babelPluginTsdx.js
+@@ -73,10 +73,6 @@ exports.babelPluginTsdx = rollup_plugin_babel_1.default.custom(() => ({
+             { name: '@babel/plugin-proposal-optional-chaining' },
+             // Adds syntax support for default value using ?? operator
+             { name: '@babel/plugin-proposal-nullish-coalescing-operator' },
+-            {
+-                name: '@babel/plugin-transform-regenerator',
+-                async: false,
+-            },
```
cowboyd, line 12: suprisingly clean!

taras: @cowboyd I suspect it's because jest is included with tsdx. It has the same types as mocha and it's probably loaded alphabetically which causes mocha types to complain.

packages/events/tsconfig.dist.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "./tsconfig",
+  "compilerOptions": {
+    "module": "esnext"
```
cowboyd, line 4: What's the default in `@frontside/tsconfig` or does it say?

taras: It defaults to module: "commonjs"

cowboyd: It's a side discussion, but should it?

taras: not sure, probably not.

PR #121, opened 2020-05-18
Add channels, a new core primitive

jnicklas:
This is an evolution of and replacement for `Mailbox` as used extensively in the [bigtest](https://github.com/thefrontside/bigtest/) package. There are a number of key differences which reflect some observations and lessons learned from `Mailbox`:

1. Every subscriber has its own queue of messages, this means we can subscribe from multiple places without accidentally "stealing" messages.
1. Instead of `receive` taking some argument which defines the parameters of the next message, we can use `filter` to change the parameters of the entire subscription. This has the advantage of not accumulating messages we are not interested in and leaving them in the queue.
1. Map and filter are not operations, their effect is only realized when subscribing. With mailbox we found that using `map` was awkward because we would have to perform a double operation, which required an awkward double `yield`.
1. Mailbox was written before resources, so it was not designed around their strengths.
1. This was written with Typescript in mind from the beginning, so there is a much greater emphasis on type safety.


jnicklas:
Why the name change? I dunno, it feels logical to me, and makes it easier to use both alongside eachother while we transition. 🤷‍♂️

taras:
@jnicklas I fixed your build and added tsdx

PR #126, opened 2020-05-25
Setup changesets for CHANGELOG management

taras:
## Motivation

When there is a problem, last thing you want to be doing is trying to decipher Git history to figure out what change might be causing the problem. In these situation a good CHANGELOG is the real MVP, but keeping a good changelog is difficult. It requires discipline and attention to detail, both of these qualities are in short supply. 

This is even more difficult in packages that have dependencies on other packages in the mono repo. Dependencies in those packages need to be updated to reference new version. That's a lot of things to keep track of.

## Approach

Use [changesets](https://github.com/atlassian/changesets) to automate generation of changelogs. Here is the TL;DR for changesets

Frontside Diagrams 003

1. You don't have to touch versions or reference versions
2. You run `yarn changeset` to create changesets
3. Each changeset describes which packages need to be bumped and by what level: patch, minor, major
4. Changesets are accumulated in .changesets directory
5. Once a PR with changesets is merged into master, a new Release PR will be created with CHANGELOG files updated and versions bumped on all packages effected by the update
6. Once the Release PR is merged, transparent publishing will take care of publishing the package

### Changeset bot

Changeset bot will tell you how each package will be bumped in the release PR.

image

### peerDependencies

> if you list a package as a peerDependency of another package, this causes the package with the peerDependency to be released as a major change. This is because peerDependency changes will not be caught by package installation.

[Read more](https://github.com/atlassian/changesets/blob/master/docs/decisions.md#the-versioning-of-peer-dependencies)

## TODO

- [x] We need to run `yarn changeset version`on release publish. How do we do this?
- [x] Do we need to use pre release to publish previews?
- [x] How do we create PRs with change versions that we need to commit back to the code?


jbolda:
RE: your TODOS. I still see value in publishing the previews. To accomplish the other two, add the following to the end of your `release.yml`, and it will open a PR when there are changes. The PR will contain the versions bumped and changelogs updated. It is effectively running `changeset version` for you. When you merge in their PR, your `thefrontside/actions/synchronize-with-npm` will still take care of the publishing.

```yml
      - name: Create Release Pull Request
        uses: changesets/action@master
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

minkimcello:
@jbolda @taras Some questions:

1. I'm assuming user needs to run `yarn changeset` manually for every pull request? Can that be automated? And if user forgets to run that command, does changeset remind you to do so? Or is that something we should incorporate to the preview workflow?
2. Do we want the release workflow to run for both user-pr merge and changeset-pr merge? Or should we find a way to prevent it from running for user merges?

jbolda:
> 1. I'm assuming user needs to run `yarn changeset` manually for every pull request? Can that be automated? And if user forgets to run that command, does changeset remind you to do so? Or is that something we should incorporate to the preview workflow?

Yes, the user should, but the bot also has a link to make it easy for a maintainer to add one through the Github UI. The bot will note of there are no changesets as well.

> 2. Do we want the release workflow to run for both user-pr merge and changeset-pr merge? Or should we find a way to prevent it from running for user merges?

Referring to the changesets release workflow here? I think it is most appropriate that it runs on every push to master. It will either open a new PR (if one isn't open) or continue to update an existing open one. That would allow a maintainer to release at their discretion of which it could include any number of changes from either source.

taras:
> Can that be automated? 

I can't be automated because it requires user input. The developer has to specify the level of change: patch, minor or major

> Do we want the release workflow to run for both user-pr merge and changeset-pr merge?

I think it's fine to run both. There is one caveat. For preview prs, we might want to allow changesets to do versioning because it automatically updates all of the dependencies. So, we might actually want to run `yarn changeset version` and not run our own version logic.

minkimcello:
@taras

> So, we might actually want to run yarn changeset version and not run our own version logic.

Is it possible to configure it so that it uses version-SHA? Or else how would we publish the preview packages?

jbolda:
Oh, are you using an "alpha" version number to publish the branch to?

taras:
image

@minkimcello I wonder if we can set all versions to that tag

cowboyd:
😍 😍 😍 This all looks amazing! Only question(s):

1. Must you run `yarn changeset` on every PR, or only on release?
2.  [smart-flies-rush.md](https://github.com/thefrontside/effection/blob/a7d8077a27b4e02c51f8d7c163607174c5083afd/.changeset/smart-flies-rush.md) ???

minkimcello:
@jbolda We use the commit sha and tag it at the end of semver to give it a unique version number so that it doesn't conflict when we publish to NPM.

@taras problem is I don't think we can publish by tag. once the version numbers overlap, it'll return an NPM error.

taras:
> Must you run yarn changeset on every PR, or only on release?

Don't have to run `yarn changeset` at all if you don't want to. You just need to create a file in .changeset directory. You can do that by clicking [Click here if you're a maintainer who wants to add another changeset to this PR](https://github.com/thefrontside/effection/new/tm/setup-changesets?filename=.changeset/famous-ants-repeat.md&value=---%0A%0A---%0A%0ASetup%20changesets%20for%20CHANGELOG%20management%0A) on Changesetbot's comment. All it does it create a file for you that describes how to bump packages. 

The idea is that you would create a changeset when you create a PR that has something releaseable. You want to do it while the changes are fresh on your mind. 

> smart-flies-rush.md ???

This the kind of file it creates. Once this is merged, a PR is Release PR is created (or reused if one is already created). When the Release PR is merged, those changeset files are removed.

taras:
> @taras problem is I don't think we can publish by tag. once the version numbers overlap, it'll return an NPM error.

How are you making these tags?

minkimcello:
> @minkimcello I wonder if we can set all versions to that tag

@taras I was referring to this. We can't publish with `v0.0.0-branch_name` because the second publish will overlap with the first.

taras:
@minkimcello I think we can just do `v0.0.0-${SHA}` if `yarn changeset version` will allow for that

cowboyd:
@taras @minkimcello 
> I think we can just do v0.0.0-${SHA} if yarn changeset version will allow for that

One thing that I've noticed with our current snapshot publishing is that it's not compatible with semver. In other words, if the current version of my software depends on effection `^0.6.2`, then I really want my snapshot `0.6.2-${SHA}` to be compatible with the `^0.6.2` range. If we are going to be messing with the snapshot publishing, now would be a great time to revisit how we construct the version and make sure that they are compatible with the same set of ranges as the base from which they were derived.

jbolda:
Looking through the action it looks like you are bumping the package version and applying the git sha, then tossing that effort away. Is that true? It seems like that would still work ignoring the point Charles is making.

minkimcello:
@jbolda Yeah, we're only using the git sha version to publish the preview package and the action tosses out the modified `package.json`. It's unable push to master anyway.

@cowboyd Isn't that a side effect that we want? If `0.6.2-sha` is published and recognized as semver, wouldn't everybody else using `^0.6.2` potentially receive breaking changes?

PR #128, opened 2020-05-28
RFC: Effection Subscriptions

cowboyd:
[RFC: Effection Subscriptions](https://github.com/thefrontside/effection/blob/iteration-proposal/rfc/001-iteration.md)


cowboyd:
I'm going to shut this down since it was used for exploratory work to launch the subscription API, but now the API is divergent in congress with usage, and evolving independently.

PR #129, opened 2020-05-29
Subscriptions Level 1

cowboyd:
This provides the `Subscription` interface and the `createSubscription` operation as detailed in the subscription RFC #128

### DIFF
--- a/packages/subscription/src/semaphore.ts
+++ b/packages/subscription/src/semaphore.ts
@@ -1,0 +1,12 @@
+export class Semaphore<T> {
+  waiters: Array<(value: T) => void> = [];
+
+  signal = (value: T) => {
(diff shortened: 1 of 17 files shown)


packages/subscription/src/semaphore.ts
```diff
@@ -0,0 +1,12 @@
...
+
+  signal = (value: T) => {
+    let next = this.waiters.pop();
+    if (next) {
+      next(value);
+    }
+  }
+
+  wait = () => new Promise<T>(resolve => this.waiters.push(resolve));
+}
```
jnicklas, line 12: This is so cool, I've though about whether a semaphore/condvar is a thing that could be a low level primitive for us, and this is such a neat and simple implementation. Love it!

PR #133, opened 2020-06-04
Implement subscriptions level 2

cowboyd:
According to the [RFC][2], the second level after attaining the core subscription operation `createSubscription` is to have a way to

1. Make any object subscribeable
2. Given any subscription, have a way to transform that subscription by mapping or filtering.

Approach
--------

Define an "interop" symbol to be exported that should return a
subscribe operation:

```ts
import { SymbolSubscribable, Subscribable } from '@effection/subscription';
import { Message } from 'socket';

class WebSocket implements Subscribeable<Message, void> {
  constructor(public socket: Socket) {}

  [SymbolSubscribeable]() {
    return on(this.socket, 'message');
  }
}
```

Now you can iterate over the array using the builtin `forEach`
function:

```ts
let ws = new WebSocket(socket);
yield forEach(ws, function*(message) {
  console.log('received message');
});
```

A operation that produces a subscription, or a subscribeable can be transformed with higher-order functions using the `Subscribable.from` function.

```ts
yield Subscribable.from(ws)
  .map(message => JSON.parse(message))
  .filter(message => message.type == 'command')
  .forEach(function*(command) {
    yield handleCommand(command);
  })
```

Todos & Open Questions
----------------------

- There's a bit of an asymmetry in that the "return value" of a subscription is distinct from the members of the individual items of the subscription. Operations like map, and filter _do not_ operate on the return value, only the individual items.
- we may want to consider exporting the transform functions directly instead of requiring an initial call to `Subscribeable.from()`. E.g.

```js
import { map } from '@effection/subscription';

export function * handleSocket(websocket: WebSocket) {
  yield map(websocket, JSON.parse).forEach(handleMessage);
}
```

vs

```js
import { Subscribeable } from '@effection/subscription';

export function * handleSocket(websocket: WebSocket) {
  yield Subscribeable.from(websocket).map(JSON.parse).forEach(handleMessage);
}
```

Learning
--------

Apparently, you can use [`Symbol.for()`][1] to lookup/create a symbol in a runtime-wide fashion. We use that instead of the normal `Symbol()` constructor for the exported interop delegate. This way, no matter which "version" of the `@effection/subscription` module you are getting the  `SymbolSubscribeable` from, it will return the same Symbol reference.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for
[2]: https://github.com/thefrontside/effection/pull/128

### DIFF
--- a/.changeset/cyan-pants-film.md
+++ b/.changeset/cyan-pants-film.md
@@ -1,0 +1,5 @@
+---
+"@effection/subscription": minor
+---
+
+add higher order functions like `map`, `filter`, and `forEach` for working with subscriptions
--- a/packages/subscription/README.md
+++ b/packages/subscription/README.md
@@ -87,3 +87,12 @@ that match `predicate`.
 ``` javascript
 Subscribable.from(websocket).filter(message => message.type === 'command');
 ```
+
+### Subscribable#first()
+
+An operation that produces the first item in a subscription or
+undefined if the subscription has no items.
+
+``` javascript
+let message = yield Subscribable.from(websocket).first();
+```
--- a/packages/subscription/README.md
+++ b/packages/subscription/README.md
@@ -55,3 +55,35 @@ One of the greatest advantages of using `createSubscription` is that
 the `Subscription` produced is an effection resource, and so will
 automatically be shut down when no longer needed. That way, there is
 no need to call the `unsubscribe()` method ever.
+
+### SymbolSubscribeable
+
+In order to facilitate interoperation of subscription producers and
+consumers, any object can implement the `[SymbolSubscribeable]()`
+method in order to be turned into a subscription. This follows the
+pattern of `Symbol.iterator`, and `Symbol.observable`. Any object that
+implements this method can be consumed as a subscription.
+
+### Subscribeable.from(source)
+
+In order to lift functions into the context of a subscription, you can
+use `Subscribeable.from` which will return an instance of
+`Subscribeable` that allows you to transform a subscription produced
+
+### Subscribeable#map(fn)
+
+Returns a new subscribeable whose items are transformed by `fn`. For
+example:
+
+``` javascript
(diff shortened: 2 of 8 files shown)


cowboyd:
> One minor nitpick is that I think the spelling should be Subscribable, not Subscribeable?

I think you're right! 😬

For the return value of `first()`. Do you think it should be `Operation<T | undefined>` or should we go full `Operation<Maybe<T>>`?

jnicklas:
`Operation<T | undefined>` is fine and fits well with the rest of the ecosystem.

jnicklas:
One downside of `first` is that going from `T | undefined` to `T` is a bit annoying, and in many cases we will regard the `undefined` case as exceptional. If we did have a `Maybe` type, we could have a method on it to expect a value is present and raise otherwise, but I'm not a huge fan of using monadic types, since it doesn't integrate too well with the rest of the JS ecosystem. How do you feel about having something like `expect()` as a method, which is basically just `yield first() || throw new Error("shouldn't be empty")`?

cowboyd:
I'm a-ok with having `first()` and `expect()`. 

However, if our expecation is that 99% of the time we're going to want to use `expect()` instead of `first()`, then maybe we sould have `first()` throw an exception on undefined, and have a `maybeFirst()` that produces `T | undefined`. Thoughts?

There's also the annoying case of when `T` _is_ `undefined` which is technically possible, but stupid. After all, there are easier ways to get an undefined value than pulling it through a subscription.

jnicklas:
I think it's a bit of an established convention that `first` returns `undefined` if the collection is empty (e.g. underscore, Ruby's #first, Rust's #nth, etc) so I think it'd probably be a bit surprising to have that behaviour on `first`. Even if `expect` (or whatever we call it) is more common, it still feels better to stick to conventions.

Let's keep it as is for now, and we can add something else on top of it.

PR #137, opened 2020-06-08
feat: better child_process defaults for windows

jbolda:
This sets better defaults for `spawn` on Windows, and allows them to be overwritten.

### DIFF
--- a/.changeset/clever-badgers-begin.md
+++ b/.changeset/clever-badgers-begin.md
@@ -2,4 +2,4 @@
 "@effection/node": major
 ---
 
-Let spawn shell be set, but default to process.env.shell which is a good default for Windows along with windowsHide.
+Remove shell default and only force detached as that is a requirement.
--- a/.changeset/clever-badgers-begin.md
+++ b/.changeset/clever-badgers-begin.md
@@ -1,5 +1,5 @@
 ---
-"@effection/node": patch
+"@effection/node": major
 ---
 
 Let spawn shell be set, but default to process.env.shell which is a good default for Windows along with windowsHide.
--- a/.changeset/clever-badgers-begin.md
+++ b/.changeset/clever-badgers-begin.md
@@ -1,0 +1,5 @@
+---
+"@effection/node": patch
+---
+
+Let spawn shell be set, but default to process.env.shell which is a good default for Windows along with windowsHide.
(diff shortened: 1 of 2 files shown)


packages/node/src/child_process.ts
```diff
@@ -34,9 +34,14 @@ function *supervise(child: ChildProcess, command: string, args: readonly string[
 }
 
 export function *spawn(command: string, args?: ReadonlyArray<string>, options?: SpawnOptions): Operation {
+  // Use the process.env.shell if it isn't set. This will likely be most useful for Windows
+  // as there are more terminals and the defaults are not amazing.
+  // windowsHide also hides the terminals from showing when a child process is init. Setting the boolean
+  // to true makes this operate in line with other OS, and how one might expect.
   let child = childProcess.spawn(command, args || [], Object.assign({}, options, {
-    shell: true,
+    shell: !options && !options.shell ? process.env.shell : options.shell,
```
cowboyd, line 42: @jnicklas what was the original reason for setting the `shell: true` property?

@jbolda since we're using a modern TypeScript, you should be able to say `options?.shell || process.env.shell`

jnicklas: I think it's a bit of a mistake actually? We do want `shell: true` in bigtest, where this was extracted from, but we probably shouldn't always set it here.

cowboyd:
@jbolda Any chance you could change this PR to _remove_ the `shell` and `windowsHide` properties, so that the only thing we do is run the child in its own process group?

jbolda:
Removing the shell would technically be a breaking change I think. Are we ok with that? Should I switch the bump to a minor instead?

.changeset/clever-badgers-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effection/node": major
+---
+
+Let spawn shell be set, but default to process.env.shell which is a good default for Windows along with windowsHide.
```
cowboyd, line 5: @jbolda While major is the right change, I think the changelog message is no longer needed.

jbolda: 🤦‍♂️

cowboyd:
Going to hold off til Monday to merge... thanks @jbolda!

.changeset/clever-badgers-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effection/node": major
```
cowboyd, line 2:
```suggestion
"@effection/node": minor
```
Even though this is a breaking change, I think we want to stay pre 1.0 for now since we may want to re-visit the whole `ChildProcess` api or bring it into its own package

PR #138, opened 2020-06-08
Migrate fetch() into @effection from bigtest/webdriver

minkimcello:
## Motivation

We need fetch in Release Manager app. fetch API was implemented in [bigtest/webdriver](https://github.com/thefrontside/bigtest/commit/2ad53034fddc515fefd34ddaf17720c456605655#diff-7a867561ff00c4276c08b5f7c85b28fa). We wanted to pull it out and make it available in effection/fetch.

## Approach

Created a package called `@effection/fetch` which uses `cross-fetch` and wrote a test for it. 

## TODOs
- [x] Generate preview package to consume in bigtest to see if it still works.
  - Tests passing with fetch consumed from this pull request's preview can be seen [here](https://github.com/thefrontside/bigtest/pull/317)
- [x] Consolidated .gitignore into just one on the root of repo.
- [x] Swap out node-fetch for cross-fetch
  - `node-fetch` takes in [arguments for options](https://github.com/node-fetch/node-fetch#api) but `cross-fetch` does [not](https://github.com/lquixada/cross-fetch#usage). Any hints on how I can refactor this part?
- [x] Write tests?
  - ~~Being did.~~ Did.


packages/fetch/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "repository": "https://github.com/thefrontside/effection.git",
+  "author": "Frontside Engineering <engineering@frontside.io>",
+  "license": "MIT",
+  "files": [
+    "dist/*",
+    "README.md"
+  ],
+  "scripts": {
+    "lint": "eslint '{src,test}/**/*.ts'",
+    "prepack": "tsc --outdir dist --declaration --sourcemap",
```
taras, line 16: Let's use tsdx as we do in other packages like effection/event & effection/node

packages/fetch/src/native-fetch.ts
```diff
@@ -0,0 +1,6 @@
+export { RequestInfo, RequestInit, Response } from 'node-fetch';
+export { AbortController } from 'abort-controller';
+
+import fetch from 'node-fetch';
```
taras, line 4: Is this going to be replaced with cross-fetch?

packages/fetch/package.json
minkimcello: Done

minkimcello:
@taras No .gitignore?

PR #139, opened 2020-06-09
Add `match` method to subscribable 

jnicklas:
This was originally proposed for the channel API in #121.

I believe that this is the most common way of using `filter` and it is much easier to read and write. Also given `DeepPartial<T>` is is actually typesafe in TypeScript as well.

### DIFF
--- a/.changeset/subscribable-match.md
+++ b/.changeset/subscribable-match.md
@@ -1,0 +1,5 @@
+---
+"@effection/subscription": minor
+---
+
+Add the `match` method to `Subscribable` to filter a subscription by pattern
--- a/packages/subscription/README.md
+++ b/packages/subscription/README.md
@@ -88,6 +88,16 @@ that match `predicate`.
 Subscribable.from(websocket).filter(message => message.type === 'command');
 ```
 
+### Subscribable#match(reference)
+
+Return a new `Subscribable` that only produces items from its source that match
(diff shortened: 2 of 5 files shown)


jnicklas:
I don't really understand how this changeset stuff works, do I need to add anything?

cowboyd:
> I don't really understand how this changeset stuff works, do I need to add anything?

It's a really manual process for now (the app we're working on will automate this) but the idea is to capture the impact of a change on a future release at the moment of the change rather than the moment of the release.

In this case, you would add a `subscribable-match.md` to the `.changesets` directory containing the impact of the change... something like:

```md
---
"@effection/subscription": minor
---

Add the `match` method to `Subscribable` to filter a subscription by pattern
```

Now, we've recorded as part of this PR that this contains a `minor` change to the `@effection/subscription` package. Changesets can have more than one package effected, each with its own semver impact.

Then, when we go to release, instead of doing all the version bumps manually, we just run the `changset version` and it will bump all versions according to the impact of the unreleased changes and generate the changleog for you.

cowboyd:
@jnicklas I went ahead and added [the changeset](https://github.com/thefrontside/effection/pull/139/commits/3336949dc2fa981438ecd1a49437e25bc9295185)

And you can see that the bot updated the changeset comment to include the impact of this change: https://github.com/thefrontside/effection/pull/139#issuecomment-641386694

PR #144, opened 2020-06-10
Add basic channel implementation

jnicklas:
A channel is an unbounded multi producer, multi consumer queue for low-throughput synchronization, where ease of use is more important than performance.

Supersedes #121


jnicklas:
I did a port of #121 to the new subscription mechanism, which makes channels be an almost trivial thing. I do think the better typing alone makes it worth to have it over just a plain `EventEmitter` though. Also, you don't have to extract the first value from the array in the subscription which is nice. 

One thing I did realize here though, is that making sending synchronous is problematic in terms of backpressure and buffer size, since the channel basically has to be unbounded. This is the case with the current `EventEmitter` based implementation, since the [queue in the subscription](https://github.com/thefrontside/effection/blob/a3d3a9c6af29ffe3fc7d7240cf979806cfc6926a/packages/subscription/src/subscription.ts#L12) is unbounded as well. I've added a caveat to the README which explains this. I still think having synchronous sends is good for most of the use-cases we're currently thinking of, but for high-performance stuff, we'll definitely want a bounded version of channel which can handle backpressure properly and where sends are async.

Another option of course is to have sends fail completely if the channel is full and bounded, rather than blocking. Honestly I don't know enough about message queueing to make a judgement on the tradeoffs here.

/cc @cowboyd

jnicklas:
Also worth noting that this implementation allows both multiple concurrent subscribers working of the same subscription (by sharing the subscription between consumers), and multiple concurrent subscribers each with its own subscription (or a mix thereof). This is somewhat similar to consumer groups in Apache Kafka. I believe(?) that this is at least very difficult to implement properly while also handling backpressure. An implementation with backpressure would probably only allow a single subscription, sort of like the current implementation of `Mailbox` in bigtest.

jnicklas:
I think a duplex channel is probably something different entirely, with an entirely different API. It will definitely need separate generic types for sendable and receivable messages. I imagine the API would be something like:

``` typescript
let [tx, rx] = DuplexChannel.pair<Ttx, Trx>()
```

Where DuplexChannel would look something like:

``` typescript
class DuplexChannel<TReceive, TSend> implements Subscribable<TReceive, void> {
  protected pair?: DuplexChannel<TSend, TReceive>;

  static pair<Ttx, Trx>() {
    let tx = new DuplexChannel<Ttx, Trx>();
    let rx = new DuplexChannel<Trx, Ttx>();
    tx.pair = rx;
    rx.pair = tx;
    return [tx, rx];
  }

  send(message: TSend) {
    this.pair.emit(message);
  }
  
  protected emit(message: TReceive) {
    // ...
  }
}
```

I think there is probably very little overlap between a duplex channel and a regular channel, and I don't think it makes sense to try and construct a duplex channel from a regular channel.

jnicklas:
This is of course only relevant to paired duplex channels, but non-paired duplex channels are basically just a `Subscribable`+`send`. We could provide an interface for this if we want to, but I don't think there is much point in even adding an implementation for it.

jnicklas:
One thing I started thinking about is if it should be possible to `close` a channel, which would cause all subscriptions to terminate. I do think this makes sense, but then we also probably want to have some way of allocating a channel as a resource so that it closes at the end of the scope. Would it make sense to provide both APIs, where you can choose to either manually close the channel, or close it automatically?

cowboyd:
> One thing I started thinking about is if it should be possible to `close` a channel, which would cause all subscriptions to terminate. I do think this makes sense, but then we also probably want to have some way of allocating a channel as a resource so that it closes at the end of the scope. Would it make sense to provide both APIs, where you can choose to either manually close the channel, or close it automatically?

What if we always made channel a resource that closed automatically, but also allowed explicit `close()` where `close` is a no-op after the first call. That's consistent with how we use "channel-like" things like sockets work today, and then you wouldn't have to worry about different API's for user-managed vs auto-managed shutdown.

packages/channel/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
cowboyd, line 1: We should extends `@frontside/tsconfig`

packages/channel/src/channel.ts
```diff
@@ -0,0 +1,34 @@
...
+    return createSubscription(function*(publish) {
+      yield spawn(forEach(on(bus, 'message'), function*([message]: [T]) {
+        publish(message)
+      }));
+      yield once(bus, 'close');
+    });
+  }
+
+  close() {
+    this.bus.emit('close');
```
cowboyd, line 32: @jnicklas Something about this feels right in that closing a channel, completes all subscriptions. So we don't have to change anything about the `close()` method to make it a no-op, it just means that on subsequent calls, nobody is listening 😎 

What do you think about having `createChannel` be an operation that produces a channel resource. It's hard to imagine a case where you wouldn't want your channel to be a resource that shut down when it passes out of scope. It certainly would not be good to have subscriptions waiting for channels, so ensuring `close()` is always called seems like the right thing whether you're doing it manually or in a resource teardown.

cowboyd:
@jnicklas two questions:

1. what do you think about having `close` be able to take an argument, that allows for transmission of information about how the subscription ended. For example, in websockets, the close event could then be passed  to the subscription so that subscribers can tell if the shutdown was clean or not.
2. Also, in line with the gist I showed you last week, what about sidestepping `EventEmitter` completely by [using a simple Set of callbacks][1]? That way there's a lot less indirection, and we avoid an extra [450] lines of JavaScript as a dependency.

[1]: https://gist.github.com/cowboyd/3e2c499a8e43079a2ae383e04e1b4188
[2]: https://github.com/Gozala/events/blob/master/events.js

packages/channel/src/channel.ts
jnicklas: I'm a bit torn on this. I think it'll be rather common to have channels which never close (e.g. Atom in bigtest) and in those cases, having to allocate a channel via a resource is a bit inconvenient. Having *both* as an API might be a bit confusing.

jnicklas:
> what do you think about having close be able to take an argument, that allows for transmission of information about how the subscription ended. For example, in websockets, the close event could then be passed to the subscription so that subscribers can tell if the shutdown was clean or not.

Sounds like a good idea.

> Also, in line with the gist I showed you last week, what about sidestepping EventEmitter completely by using a simple Set of callbacks? That way there's a lot less indirection, and we avoid an extra [450] lines of JavaScript as a dependency.

I do think that there are probably some subtle performance and stability issues in re-implementing an event emitter, that we shouldn't dismiss too easily. I do think we will want to implement this ourselves eventually, but it'd be nice to avoid any potentially annoying issues for the time being. Since the event emitter is private we can do this easily in the future without breaking compatibility in any way.

packages/channel/src/channel.ts
cowboyd: Given that we now have the ability to `close()` with an argument which is a win, I think that answers the question of always creating a channel as a resource, since an automatically closed channel can not know ahead of time what the argument to `close` should be.

In the event that you want to autoclose, it's up to the programmer about what to send to the `close()` method and they can manually do that when they create the channel as a resource..

cowboyd:
> One thing I did realize here though, is that making sending synchronous is problematic in terms of backpressure and buffer size, since the channel basically has to be unbounded. This is the case with the current `EventEmitter` based implementation, since the [queue in the subscription](https://github.com/thefrontside/effection/blob/a3d3a9c6af29ffe3fc7d7240cf979806cfc6926a/packages/subscription/src/subscription.ts#L12) is unbounded as well. I've added a caveat to the README which explains this. I still think having synchronous sends is good for most of the use-cases we're currently thinking of, but for high-performance stuff, we'll definitely want a bounded version of channel which can handle backpressure properly and where sends are async.
> 
> Another option of course is to have sends fail completely if the channel is full and bounded, rather than blocking. Honestly I don't know enough about message queueing to make a judgement on the tradeoffs here.

@jnicklas Yeah, after playing around with a couple different APIs I think it is pretty important to maintain a synchronous `send()` as the baselines since I think this will end up becoming the primary way to "push" values into effection from the outside where there is no way to evaluate operations. Perhaps we can split `Channel` in the future into a `Port` interface for pushing raw values, and then enhancing `Channel` to do more fancy stuff.

jnicklas:
@cowboyd I think all questions on this are resolved? Should we merge this?

PR #149, opened 2020-06-13
Only require the event listener methods we use

cowboyd:
Motivation
-----------
In [BigTest][1], we're settling on using the `ws` websocket library across the board, but sockets from that library don't implement the full `EventTarget` api, only `{add,remove}EventListener`. This means that we can't use those sockets easily because they cannot be passed to the underlying subscription mechanism which is `on()` from '@effection/events'.

Approach
----------
This relaxes the requirement of what can constitute an EventSource to just things that have the `addEventListener` and `removeEventListener` methods, since those are the only ones we use.

[1]: https://github.com/thefrontside/bigtest

### DIFF
--- a/.changeset/narrow-event-target-interface.md
+++ b/.changeset/narrow-event-target-interface.md
@@ -1,0 +1,7 @@
+---
+"@effection/events": patch
+---
+
+Only require EventTarget-like objects for the `on()` method to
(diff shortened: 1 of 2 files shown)


jnicklas:
Great idea!

PR #155, opened 2020-06-22
Chainable methods on subscriptions

### LINKED ISSUE #154: Subscribables vs subscriptions
When we first added subscriptions, we were anticipating on mostly working with `Subscribable` types, and not so much with `Subscription` types, so all of the convenience methods like `map` and `filter` work on subscribables. After working with subscriptions a bit, I think this was a mistake. The natural thing to work off of is a subscription. Sometimes there isn't even a subscribable.

I also realized that it is possible to implement `filter` and `map` without having to make the actual functions operations. For example, filter could be implemented like this:

``` typescript
function filter<T, TReturn>(subscription: Subscription<T, TReturn>, predicate: (value: T) => boolean): Subscription<T, TReturn> {
  return {
    *next() {
      while(true) {
        let result = yield subscription.next();
        if(result.done) {
          return result;
        } else if(predicate(result.value)) {
          return result;
        }
      }
    }
  }
}
```

This take a subscription and return a new subscription. We could even have all of the free functions return a chainable interface, so we could have something like this:

``` typescript
let subscription = yield subscribe(someSubscribable);
yield subscription.map(...).filter(...).forEach(...);
```

I think this is reasonably ergonomic.

Thoughts? cc @cowboyd

jnicklas:
This adds a free function called `subscribe` which returns a subscription from a subscribable. It also adds chainable methods via `ChainableSubscription` which makes it possible to call `map`, `filter` and `match` on subscriptions.

Finally it deprecates `Subscribable.from` and chaining off of a subscribable.

Closes #154

### DIFF
--- a/.changeset/subscription-chain.md
+++ b/.changeset/subscription-chain.md
@@ -1,0 +1,4 @@
+---
+"@effection/subscription": minor
+---
+Add a free `subscribe` function, allow chaining of `map`, `filter` etc on subscriptions and deprecate chaining on subscribables.
--- a/packages/subscription/src/chainable-subscription.ts
+++ b/packages/subscription/src/chainable-subscription.ts
@@ -1,0 +1,76 @@
+import { Operation } from 'effection';
+import { Subscription } from './subscription';
+import { DeepPartial, matcher } from './match';
+
+export class ChainableSubscription<T,TReturn> implements Subscription<T,TReturn> {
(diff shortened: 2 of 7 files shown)


jnicklas:
That makes sense to me. I pushed a commit to undo the deprecation. Let's keep both for a while and see how it pans out.

PR #156, opened 2020-06-22
convert fetch response into resource

cowboyd:
We were running into an issue where if the bytes of the body were not fully downloaded by the time headers had been received, the request was being aborted, which was causing the `response.text()` promise to become rejected.

This converts the `Response` object to a resource so that only when the response is no longer used will the request be aborted.

There were a bit of gymnastics that we had to adopt in order to make this work since it is possible that fetch could be halted _before_ the underlying response was returned from native fetch. In that event, we still need to abort the request.


cowboyd:
@jnicklas As an aside, I've been thinking about how to make this code feel nicer rather than what we ended up doing. I think that an intuitive solution would be to treat the `AbortController` itself as the primary resource, and then the `Response` object is just a "container" resource for the abort controller, that way when the container passes out of scope, so does the controller:

If we had that, then the `fetch` operation could be made to look coherent again with something along the lines of a `retain()` operation :

```ts
import { resource, retain } from 'effection';

export function* fetch(info: RequestInfo, init: RequestInit = {}): Operation<Response> {
  let controller = yield resource(new AbortController(), function*() {
    try { 
      yield; 
    } finally { 
      controller.abort(); 
    }
  });
  init.signal = controller.signal;

  let response: Response = yield nativeFetch(info);
  return yield retain(response, controller);
}
```

PR #159, opened 2020-06-23
Add changeset action to automate release pull requests

minkimcello:
## Motivation
Whenever we want to publish a release we need to manually checkout, run `yarn changeset version`, git add and push, and then create the pull request on Github. Not only do we need to run the command and create the pull request manually, we also need to update the release branch as more pull requests get merged to the main branch. Using the changeset action will mitigate all those inconveniences.

## Approach
- Added changeset action to release workflow. When a pull request gets merged it will see if changeset files exist on the repository and if it does it'll create a pull request (or push to a pre-existing release pull request).
  - The release action will run regardless of whether or not changesets files exist.
- Added conditional to the publish-preview action so that the action only runs if the head branch of the pull request is _not_ `changeset/release-master`.
Screen Shot 2020-06-23 at 2 11 13 PM

## TODOs
- [ ] Should we create the same conditional on the test workflow so that we don't use resources on testing for the release pull request?

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -3,7 +3,7 @@ name: Publish Preview
 on:
   pull_request:
     branches:
-      - master
+      - 'master'
 
 jobs:
   publish-previews:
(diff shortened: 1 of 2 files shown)


cowboyd:
@minkimcello Does this actually work? I thought we'd discovered that it didn't.

minkimcello:
@cowboyd Which part of it wasn't working? Because I tested the functionality earlier on one of my private repos and it was working fine. All we need it to do is to generate the pull request.

PR #168, opened 2020-06-26
Ensure subscription resources are retained

### LINKED ISSUE #167: Wrapping with `ChainableResource` strips resources
There is a subtle bug in how #155 is implemented.

It is quite common to return a resource as a subscription, like this:

``` typescript
class Something {
  *[SymbolSubscribable]() {
    return yield resource(...);
  }
}
```

The problem comes in the way that the `subscribe` function is implemented in `@effection/subscription`. Somewhat simplified it boils down to:

``` typescript
function* subscribe(source) {
  let inner = yield source[SymbolSubscribable]();
  return new ChainableSubscription(inner);
}
```

The problem here is that the resource is bound to the context inside the subscribe function, and it is *not* returned from this context, meaning it will expire when subscribe exits. This is of course wrong.

jnicklas:
Closes #167

### DIFF
--- a/packages/subscription/src/chainable-subscription.ts
+++ b/packages/subscription/src/chainable-subscription.ts
@@ -1,9 +1,21 @@
-import { Operation } from 'effection';
+import { Operation, resource } from 'effection';
 import { Subscription } from './subscription';
 import { DeepPartial, matcher } from './match';
+import { SubscriptionSource, subscribe } from './subscribable';
+
+const DUMMY = { next() { throw new Error('dummy') } };
 
 export class ChainableSubscription<T,TReturn> implements Subscription<T,TReturn> {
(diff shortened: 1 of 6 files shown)


packages/subscription/src/chainable-subscription.ts
```diff
@@ -1,10 +1,22 @@
-import { Operation } from 'effection';
+import { Operation, resource } from 'effection';
 import { Subscription } from './subscription';
 import { DeepPartial, matcher } from './match';
+import { SubscriptionSource, subscribe } from './subscribable';
+
+const DUMMY = { next() { throw new Error('dummy') } };
```
jnicklas, line 6: This is a bit of a hack, but it was the only way I could come up with which doesn't also weaken the types of `ChainableSubscription` by making the inner subscription optional.

PR #172, opened 2020-07-06
Bump frontside/eslint-config

minkimcello:
## Approach
- Upgraded `@frontside/eslint-config` from `v1.0.0` to `v1.1.1` which comes prepacked with:
  - `@typescript-eslint/eslint-plugin`
  - `@typescript-eslint/parser`
  - `eslint-plugin-prefer-let`
- Added `eslint` and ~`typescript`~ to the root of the monorepo.

## TODOs
- [x] Once [this](https://github.com/thefrontside/javascript/pull/6) is merged and published I can update `eslint-config` to the latest version.


cowboyd:
Why is typescript a duplicate dependency? Does `tsdx` depend on it explicitly?

minkimcello:
@cowboyd ~~I'm not sure `tsdx` depends on it~~ It doesn't look like `tsdx` requires `typescript` to be installed. I'm guessing we just never bothered to remove it after we started using `tsdx`, but `eslint-config` comes with `typescript` whether we need it or not so I was able to remove it from the individual packages. It still builds fine using `tsdx`.

cowboyd:
Even though our `eslint-config` depends on typescript, and so some version of the typescript compiler will be in the project, it's not a good practice to rely on implicit dependencies when your need is explicit. The heuristic should be, "if I removed `@frontside/eslint-config` from my dependencies, can I guarantee my project continue to work otherwise?" In this case, our code is in typescript, and so the answer is no. 

Our code is in typescript, and so we therefor have a primary dependency on packages whose contract is to provide the typescript compiler.

What could work, if we want to optimize our dependencies is to create a `@frontside/typescript` package which provides `typescript` _and_ `@frontside/eslint-config` _and_ `@frontside/tsconfig` as part of its publicly stated contract.

minkimcello:
@cowboyd Oh I see. That makes sense. Should I have named `frontside/eslint-config` to `frontside/eslint` if we're providing `eslint` within `eslint-config`? Or maybe I should not have added `eslint` and `typescript` as dependencies to `eslint-config`?

package.json
```diff
@@ -30,7 +30,8 @@
   "devDependencies": {
     "@frontside/eslint-config": "mk_remove-deps",
     "eslint": "5.16.0",
-    "patch-package": "6.2.2"
+    "patch-package": "6.2.2",
+    "typescript": "^3.9.5"
```
cowboyd: I don't think we need typescript in the root

cowboyd (changes requested):
Each package should have the things it needs to build in its dependencies, not inherit them from the ambient environment.

cowboyd:
We shouldn't have `eslint` at the root of the repo either.

minkimcello:
So generally do we keep devDependencies in its own package and only bring dependencies to the root if all packages utilize it?

Should I add `eslint` to every package? It was only in `packages/effection`.

cowboyd:
> So generally do we keep devDependencies in its own package and only bring dependencies to the root if all packages utilize it?

We should never bring dependencies to the root.

As for `eslint`, yes, let's add it to just `effection`

minkimcello:
> We should never bring dependencies to the root.

What about dev dependencies? We have devDeps at the root of both effection and bigtest monorepos. And also I put `@frontside/eslint-config` at the root for both. Should that also be in each package instead?

> As for `eslint`, yes, let's add it to just `effection`

Why is it okay for `eslint` to only be in `packages/effection` when every package runs eslint script?

cowboyd:
> Why is it okay for eslint to only be in packages/effection when every package runs eslint script?

Oh, I thought you said that only `effection` used `eslint`. Every package that uses `eslint` should have it as a dependency.

cowboyd (approved):

PR #174, opened 2020-07-08
Main exit code

jnicklas:
Makes it possible to set a custom exit code and silence error printing on some exceptions. This will allow us to fail cleanly with the exit code we want and without printing a bunch of debug information we don't want.


cowboyd (commented):
In my experience, whenever an error percolates up to the very top, it means that it was completely unexpected and more often than not indicates an error with the error handling itself. 

In those instances, you always want to log it in full.  On the other hand, when you want to unwind the stack at any given point in a silent way for "expected" errors you use a custom error class, which is exactly what the `SuiteError` does https://github.com/thefrontside/bigtest/pull/394/commits/1578af410e2debfce6aa57976bab79ca21027174#diff-d20d9ac60915a9bedbe0c12f2697a728

It feels like this functionality ought to be more easily at hand for anybody wanting to create a custom exit. If we brought the functionality of `SuiteError` over to `@effection/node` we could have a `MainError` class that let anybody define custom exits and still have the benefit of type-checking (especially if we evolve the interface).

We might even want to do things like pass stream handling for `stderr`

```ts
if (!options.valid) {
  throw new MainError({exitCode: 23, print: function* (stderr, stdout) => {
    yield stderr.promises.write(options.helpMessage);
  })
}
```

Then, I think we wouldn't even need the `effectionSilent` flag, just inside the main block:

```ts
catch (e) {
  if (e instanceof MainError) {
   yield e.print(process.stderr, process.stdout);
   process.exit(e.exitCode);
  } else {
    console.error(e);
    process.exit(1);
  }
```

jnicklas:
@cowboyd I implemented this the way you suggested, but chose to go with checking the error name instead of an `instanceof` to prevent possible packaging issues.

cowboyd:
> @cowboyd I implemented this the way you suggested, but chose to go with checking the error name instead of an `instanceof` to prevent possible packaging issues.

Ah yes, that’s actually more the JavaScript Way :tm:

packages/node/src/main.ts
```diff
@@ -11,8 +30,15 @@ export function main<T>(operation: Operation<T>): Context<T> {
         process.on('SIGUSR1', debug);
         return yield operation;
       } catch(e) {
-        console.error(e);
-        process.exit(-1);
+        if(e.name === 'EffectionMainError') {
+          if(!e.silent) {
```
cowboyd: I don't know if this is your experience, but mine is that most of the time that I go to the effort of throwing a custom error designed to kill the process, I'm going to make the output custom as well, so what do you think about making `silent` true by default?

In other words, usually what you want is to either a) write a message to stderr (no stack trace), or b) log it fully as an uncaught error. (b) seems to correspond to the case where where you don't have a `MainError`.

Having a boolean option to either do nothing or fully log the error using the default `console.error`, doesn't seem flexible enough.

In other words, let's say we wanted to do this:

```
$ bigtest server --launch waterfox
❗ can't find a driver for 'waterfox'. Did you mean 'firefox'?
```

What if we  _always_ log the `message` of a `MainError` if it's defined (we can make it undefined by default), and then do our `console.error` if there is a `logBacktrace` option.

then again, whether to log the backtrace doesn't feel like something that should be controlled by the site where the exception is thrown, but rather from where it's caught via a `LOG_LEVEL` or something.

jnicklas: This is a great suggestion :) I don't really think it's that useful to be able to print the stacktrace at all though. I did leave an option for it in there, just in case.

cowboyd (approved):
It’s complicated, but should allow us good control  over expected  error exits ::)

PR #178, opened 2020-07-10
Chaining subscriptions via `subscribe(...)`

jnicklas:
The distinction between ChainableSubscription and Subscribable.from is pretty odd and hard to understand. What if we could have our cake and eat it too? This allows us to do:

``` typescript
yield subscribe(people).map((s) => s.name).expect();
```

AND

``` typescript
let subscription = yield subscribe(people).map((s) => s.name)
yield subscription.expect();
```

AND

``` typescript
let subscription = yield subscribe(people);
yield subscription.map((s) => s.name).expect();
```

Why do we want all three of them? Because their semantics are slightly different when it comes to consumption of the subscription, and because we don't always have an actual subscribable object.

This is all also very easy to understand, because even though there is a lot of trickery involved in making this actually work, it *looks* very simple.

This also deprecates `Subscribable.from` and `forEach` since they are no longer needed.

### DIFF
--- a/packages/subscription/src/chainable-subscribable.ts
+++ b/packages/subscription/src/chainable-subscribable.ts
@@ -1,0 +1,62 @@
+import { Operation } from 'effection';
+import { ChainableSubscription } from './chainable-subscription';
+import { DeepPartial } from './match';
+import { Subscribable } from './subscribable';
+import { SymbolSubscribable } from './symbol-subscribable';
+import { Subscription } from './subscription';
+
+export interface ChainableSubscribableMethods<T,TReturn> extends Subscribable<T,TReturn> {
+  filter(predicate: (value: T) => boolean): ChainableSubscribable<T, TReturn>;
+  match(reference: DeepPartial<T>): ChainableSubscribable<T,TReturn>;
+  map<R>(mapper: (value: T) => R): ChainableSubscribable<R, TReturn>;
+  first(): Operation<T | undefined>;
+  expect(): Operation<T>;
+  forEach(visit: (value: T) => Operation<void>): Operation<TReturn>;
+  next(): Operation<IteratorResult<T, TReturn>>;
+}
+
+export type ChainableSubscribable<T, TReturn> = Operation<ChainableSubscription<T, TReturn>> & ChainableSubscribableMethods<T, TReturn>;
+
+export function makeChainable<T, TReturn>(
+  operation: Operation<ChainableSubscription<T, TReturn>>
+): ChainableSubscribable<T, TReturn> {
+  return Object.assign(operation, {
+    *[SymbolSubscribable](): Operation<Subscription<T,TReturn>> {
+      return yield operation;
+    },
+
+    filter(predicate: (value: T) => boolean): ChainableSubscribable<T, TReturn> {
+      return makeChainable(function*() {
+        return yield ChainableSubscription.wrap<T, TReturn>(operation, (inner) => inner.filter(predicate));
+      });
+    },
+
+    match(reference: DeepPartial<T>): ChainableSubscribable<T,TReturn> {
+      return makeChainable(function*() {
+        return yield ChainableSubscription.wrap<T, TReturn>(operation, (inner) => inner.match(reference));
+      });
+    },
+
+    map<R>(mapper: (value: T) => R): ChainableSubscribable<R, TReturn> {
+      return makeChainable(function*() {
+        return yield ChainableSubscription.wrap<T, TReturn, R>(operation, (inner) => inner.map(mapper));
+      });
+    },
+
+    *first(): Operation<T | undefined> {
+      return yield (yield operation).first();
+    },
+
+    *expect(): Operation<T> {
+      return yield (yield operation).expect();
+    },
+
+    *forEach(visit: (value: T) => Operation<void>): Operation<TReturn> {
+      return yield (yield operation).forEach(visit);
+    },
+
+    *next(): Operation<IteratorResult<T, TReturn>> {
+      return yield (yield operation).next();
+    }
+  });
+};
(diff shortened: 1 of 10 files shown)


jnicklas:
could-should

packages/subscription/src/chainable-subscription.ts
```diff
@@ -12,7 +12,20 @@ export class ChainableSubscription<T,TReturn> implements Subscription<T,TReturn>
...
+    fn: (inner: ChainableSubscription<T, TReturn>) => ChainableSubscription<R, TReturn>,
+  ): Operation<ChainableSubscription<R, TReturn>> {
+    let chain = new ChainableSubscription<R, TReturn>(DUMMY);
+
+    return yield resource(chain, function*() {
+      let subscription = yield source;
+      chain.subscription = fn(subscription);
       yield;
     });
   }
```
jnicklas, line 31: The purpose of this is to retain any resources created in the wrapped operation.

packages/subscription/src/subscribe.ts
```diff
@@ -0,0 +1,8 @@
+import { Operation } from 'effection';
+import { ChainableSubscription } from './chainable-subscription';
+import { SubscriptionSource } from './subscription-source';
+import { makeChainable, ChainableSubscribable } from './chainable-subscribable';
+
+export function subscribe<T, TReturn>(source: SubscriptionSource<T,TReturn>): Operation<ChainableSubscription<T, TReturn>> & ChainableSubscribable<T,TReturn> {
+  return makeChainable(function*() { return yield ChainableSubscription.of(source) });
```
cowboyd, line 7: Is there any reason not to just avoid the intermediate generate operation?  to just make the ChainableSubscription
```suggestion
  return makeChainable(() => ChainableSubscription.of(source));
```

packages/subscription/src/chainable-subscribable.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export type ChainableSubscribable<T, TReturn> = Operation<ChainableSubscription<T, TReturn>> & ChainableSubscribableMethods<T, TReturn>;
+
+export function makeChainable<T, TReturn>(
+  operation: Operation<ChainableSubscription<T, TReturn>>
+): ChainableSubscribable<T, TReturn> {
+  return Object.assign(operation, {
+    *[SymbolSubscribable](): Operation<Subscription<T,TReturn>> {
+      return yield operation;
+    },
```
cowboyd, line 26:
```suggestion
    [SymbolSubscribable]: () => operation,
```

packages/subscription/src/chainable-subscribable.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    map<R>(mapper: (value: T) => R): ChainableSubscribable<R, TReturn> {
+      return makeChainable(function*() {
+        return yield ChainableSubscription.wrap<T, TReturn, R>(operation, (inner) => inner.map(mapper));
+      });
+    },
+
+    *first(): Operation<T | undefined> {
+      return yield (yield operation).first();
```
cowboyd, line 47: Ladies and gentlement, the first double yield 🙌

cowboyd (commented):
YES YES YES YES

This is definitely it. This is the API we want. Way to thread the needle @jnicklas! 

Really, the only feedback I have is the potential elimination of extra intermediate operations. E.g. using

```ts
return makeChainable(() => ChainableSubscription.wrap<T, TReturn>(operation, (inner) => inner.filter(predicate)));
```

instead of 

```ts
return makeChainable(function*() {
          return yield ChainableSubscription.wrap<T, TReturn>(operation, (inner) => inner.filter(predicate));
        });
```

since our new fluent api will let us create chains of operations really quickly, we might as well eliminate the number of intermediate ops that exist only to pass through a value.

packages/subscription/src/chainable-subscribable.ts
jnicklas: Hopefully this will allow us to avoid the double yield in other places 😈

jnicklas:
Frustratingly, removing the wrappers around operations does *not* work. Tests fail completely. I'm trying to track down why, but it's frustratingly hard to understand. I *think* this has something to do with when the operation is turned into a generator, since in one case we're passing a `Generator` and in the other a `GeneratorFunction`, but it's really hard to track down :(

cowboyd:
I that case, let's keep it as-is. I think when we refactor effection internals we can revisit this since there will be less ambiguity about what an operation is. https://gist.github.com/cowboyd/10d326c5f9ff2b612cf61858c6c3970e

<blockquote><div> Gist</div><div><strong><a href="https://gist.github.com/cowboyd/10d326c5f9ff2b612cf61858c6c3970e">Proposed Effection operations that are non-ambiguous</a></strong></div><div>Proposed Effection operations that are non-ambiguous - operation.ts</div></blockquote>

cowboyd (approved):

jnicklas:
Okay, so it really bothered me that I couldn't figure out why this was failing without the wrappers, so I dug a bit deeper and I found the issue.

Somewhere we are doing this:

``` typescript
export function makeChainable<T, TReturn>(
  operation: Operation<ChainableSubscription<T, TReturn>>
): ChainableSubscribable<T, TReturn> {
  return Object.assign(operation, {
    // ... lots of other stuff here
    *next(): Operation<IteratorResult<T, TReturn>> {
      return yield (yield operation).next();
    }
  });
};
```

And then in `subscribe`:

``` typescript
export function subscribe<T, TReturn>(source: SubscriptionSource<T,TReturn>): ChainableSubscribable<T,TReturn> {
  return makeChainable(ChainableSubscription.of(source));
}
```

Now `ChainableSubscription.of(source)` returns a `Generator` and a generator is advanced by calling `next` on it. And our `makeChainable` function happily overrides this method to do something else entirely, obviously breaking our generator.

picard-facepalm-face-palm

While fixing this certainly makes it behave *better*, and I think we should remove the `next` method, simply because it is confusing, unfortunately, it still doesn't work quite, due to [this test](https://github.com/thefrontside/effection/blob/d7c7d4c26afdaee3f4e19fe9bae118fcf4c8d477/packages/subscription/test/subscription.test.ts#L96-L108) failing.

jnicklas:
Let's ship it then!

PR #184, opened 2020-08-25
Make `createSubscription` chainable

jnicklas:
Subscriptions created via `createSubscription` are chainable on both sides of the yield statement. This enables patterns like:

``` typescript
yield on(window, "message").filter((m) => ...).forEach(...)
```

There is a subtle downside and backwards incompatibility with this, which is that due to `Operation` being invariant over its type parameter. Since we return a `Operation<ChainableSubscription<T>>` but we sometimes expect a `Operation<Subscription<T>>` and while...

```
ChainableSubscription <: Subscription
```

...really *should* imply that...

```
Operation<ChainableSubscription> <: Operation<Subscription>
```

...this is currently not (always) the case, so some subtle type errors might occur.

### DIFF
--- a/.changeset/create-subscription-chainable.md
+++ b/.changeset/create-subscription-chainable.md
@@ -1,0 +1,7 @@
+---
+"@effection/subscription": minor
+"@effection/events": patch
+"@effection/channel": patch
+---
+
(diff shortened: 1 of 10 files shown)


cowboyd (approved):
Love it. Subscriptions are starting to get really pleasant to use.

Issue #181, opened 2020-08-06
Make tests run on windows


cowboyd:
For good cross platform support, we should make sure our tests run on windows

jbolda:
Is the expectation to run it in CircleCI?

cowboyd:
We run our tests on github actions, so expectation would be for them to run there. I believe @taras made a little headway getting the windows VM setup ther.

taras:
Not much headway, I thought GitHub Supports Windows server. Can we use that? Otherwise we can use CircleCI

jbolda:
Aye, it does @taras. I am way more familiar with Github Actions so I can open a PR to switch it over.

cowboyd:
closed by #202

PR #185, opened 2020-08-28
chore: test in github actions

jbolda:
This adds a test script on PR and push (which is triggered when it is merged into a branch). It will run on ubuntu, mac and windows. There is also a step to cache the node_modules, but this will generally not trigger in PRs (as each PR has it's own cache) unless there a multiple commits. It also sets a timeout to catch situations where the build hangs (like what is currently happening on windows).

This is the first step to address #181. We can enable windows once they pass in CI to keep the builds green.


cowboyd:
@jbolda Nice! Do we need to do anything else in order to shutdown the CircleCI build? Also, can we name the workflow/job something other than `test/test` maybe `Test Workflow/Run Tests`? I know it's only slightly more descriptive, and we only have 3 checks now, but when the checks start to get into the 20s, it's better to get even that little extra distinction.

jbolda:
It may make sense to revoke their repo access, but that depends how you have things set up. Maybe @taras knows?

I am certainly open to renaming. Perhaps we could set the name as `Package on OS` similar to what we have in https://github.com/thefrontside/bigtest/pull/486.

taras:
> Do we need to do anything else in order to shutdown the CircleCI build?

I think it's just stopping CircleCI from following our repos.

jbolda:
@cowboyd Name has been updated. I currently have this running all the packages in one run. Should we have it split up like in BigTest? Split up will eat up the number of available runners much quicker, but it does make it a little more transparent.

cowboyd:
I think our test run is still fast enough that we can do it in a single run.

cowboyd (approved):

PR #186, opened 2020-08-28
feat(@effection/node): switch to cross-spawn, implement windows support

### LINKED ISSUE #182: Problems spawning processes in Windows
When you spawn a child process in windows,

1. it does not die when you call process.kill
2. causes a separate window to open up every time

Potentially related https://stackoverflow.com/questions/32705857/cant-kill-child-process-on-windows ?

Consider using [execa] ? 

[1]: https://github.com/sindresorhus/execa

jbolda:
This PR switches to cross-spawn. Currently we are stuck with nodejs emulation of signals on windows not killing nested processes gracefully.

This closes #182 (eventually).


packages/node/src/child_process.ts
```diff
@@ -1,52 +1,52 @@
...
 }
 
-// buffer will wait to resolve the promise until stdio is read, which works well for generators in effection
-// shell and detached will cover random piping errors, and windowsHide:true is the default
-// preferLocal runs node scripts local first
-const EXECA_DEFAULTS = {shell: process.env.shell || true, detached: true, preferLocal: true, buffer: false, stdio: 'inherit'}
+// using the shell that invokes will also hide the window on windows
+const PROCESS_DEFAULTS = {
+  shell: process.env.shell || true,
+  stdio: "inherit"
```
cowboyd: Why not let the default be what it is in node?

cowboyd:
I'm totally cool with the decision to switch to cross-spawn, which is what execa uses under the covers... did it make it easier to just stick with the existing semantics? One thing I know is that `kill()` on the process-group strategy did not work on windows. Is that something that cross-spawn can take care of for us?

jbolda:
for clarity, in discord I said:

> well, I think @effection/node really requires the ability to stream stdio as some uses in BigTest open a process, send messages to it and wait for the event rather than strictly waiting on close.
> The tests in effection imply that but digging into BigTest, it seems to be the case from my understanding. execa can't stream and it's built more around running a process and returning the output.
> It seems like it's worth using cross-spawn instead so we can still stream, and do the processing killing ourselves.

It made it so much easier to stick with the semantics. You are right about the `kill()` and I remember watching you struggle with that some back when you were working on this part of the code. `cross-spawn` is mostly a drop in replace of `child_process` spawn, so it doesn't seem to take care of it. I'm currently looking at the interesting bits of the `execa` kill, but there seems to still be bugs and orphaned processes that don't get cleaned up in windows under `execa`.

jbolda:
So I've tried a handful of considerations (and pushed a bunch to check on a clean build and make certain it wasn't bad processes hanging around on my end). From what I'm seeing and reading, the only way the child_process seems to close on forced exit `SIG*` is with `stdio` set to 'inherit'. The side effect is that with the stream on the parent, we can't observe it and `waitFor` fails on `undefined`. If we use `pipe`, it runs but won't seem to respond to any `SIG*` event. I think it's something to do with the stream on the parent vs the child.

The only other thing I can think to try is to create a separate stream on the parent, pass that too the child to pipe into and let the parent manage the data control. I haven't had a chance to try this though, and I think it might require a bit of refactoring if it does work.

Anything else that you can think of? /cc @cowboyd @taras

jbolda:
Well, it's hacky but this might work for windows:
`spawn("taskkill", ["/pid", child.pid, '/f', '/t']);`

Spawn that in response to a kill signal.

jbolda:
So I made a super minimal test case, and that proceeds and finishes as expected. So there must be something in BigTest / Effection that is keeping the process hung / open.

```js
const spawn = require("cross-spawn").spawn;

async function main() {
  const child = spawn("ping", ["www.google.com"], {
    shell: process.env.shell || true,
    detached: true,
    stdio: "pipe",
  });

  child.stdin.end();
  child.unref();
  console.log(child.pid, process.pid);

  child.stdout.on("data", function (data) {
    console.log(data.toString());
  });

  child.stderr.on("data", function (data) {
    console.error(data.toString());
  });

  setTimeout(function () {
    child.kill("SIGTERM");
  }, 500);

  child.on("exit", function (code, signal) {
    console.log("process exit " + code + " " + signal);
    child.stdout.end();
    child.stderr.end();
  });
}

main();
```

This hang only seems to happen on SIGTERM as far as I can tell, so a cheap win might be to just do this. I think everything in BigTest comes through wrapped with that. I am testing in BigTest so see if that holds valid.
image

cowboyd:
@jbolda awaiting the fruits of your research with baited breath....

jbolda:
I think the last bit of hangs are related to the socket streams not being cleaned up and Windows not shutting down because of that. Putting this after the `World.halt()` seems to let the processes close, and get green builds on Windows. It's feels bad and uses a private API, but I don't see any other way to close the connection through effection. Do you have any other ideas?

```js
    // @ts-ignore
    process._getActiveHandles().forEach((handle) => {
      try {
      handle.stdout.end()
      handle.stderr.end()
      } catch (e) {
        // no-op
      }
    });
```

This is the current diff (which has some formatting changes that I should probably switch back [ie it prettiered]):
https://github.com/thefrontside/bigtest/compare/v0...jbolda:fix/child_process_in_windows
and the PR I've been running CI on (but now I'm getting chrome binary errors after merging v0 in): 
https://github.com/jbolda/bigtest/pull/3

I think we get BigTest working (if even with patch-package), and then get the end result from BigTest working in the tests here, release and push that over to BigTest. Unless you think we should do it all directly in BigTest for now?

jbolda:
Still having troubles with the graceful shutdown for windows. I think we get this merged which prevents the extra windows and hanging at the very least (and I think we might need it in regardless even if something like Rust is being considered).

cowboyd:
@jbolda I'd like to sit down and have a complete re-think of what we need from a process library. We actually have some other work that's going on which will require some refactoring of the child process supervision resource. I'm going to sit down and document today all of our use-cases, error-handling, and what we need it to do. Then we can approach it from a "what we need" rather than "work with what we have" aspect.

jbolda:
Alright. I think it might make sense to get this merged in as a good first step at least. It takes care of a windows bugs that kill BigTest. We likely will have issues killing nested processes, but it would a) be nice to do a full repro to test and confirm, and b) I don't think it outweighs the benefits to getting this in BigTest as a decent step forward.

For the past week+, I haven't actually been having trying to solve any issues with child_process. It has been dealing with `ts-node` and it not picking up the config on windows. That is seemingly what is causing all of these tests to time out (since it compiles the generators and never ends up being able to yield properly). My hope was the resolve the last of that in BigTest that I am confident enough with what we have here isn't breaking anything else.

Does that sound like a reasonable path forward? I would like to get windows functioning at least partially before we undertake a (possibly big R&D) project especially if it involves rust.

packages/node/src/child_process.ts
```diff
@@ -1,12 +1,21 @@
-import { Operation, resource } from 'effection';
-import { once, throwOnErrorEvent } from '@effection/events';
+import { Operation, resource } from "effection";
+import { once, throwOnErrorEvent } from "@effection/events";
```
cowboyd, line 2: As a side note, we should avoid changes that simply reformat code unless the purpose of the change is to reformat code.

cowboyd:
@jbolda Ok, I think we have everything in place to un-block you. Let's get all the things unrelated to `child_process` merged in by breaking them each into the smallest set of changes that you think reasonable. Would it help if I migrated the `main.test.ts` to use `exec`?

cowboyd:
@jbolda I've migrated the `main.test.ts` file to use `exec`. The only requirement will be that you'll have to implement shutting down the process "the windows way" (this might involve invoking `taskill`for now) https://github.com/thefrontside/effection/blob/ec6635f1a115bf358748f807b9c572128720b5b3/packages/node/test/helpers.ts#L41-L62

This is a bit of duplication of the actual termination code between `posix.ts` and `helpers.ts`, and so there will be some duplication between `win32.ts` and `helpers.ts` but I think that's ok. What we're simulating in the `TestProcess` is what it's like for a user or the OS to shutdown a process from the outside in. Wheer as in our process resource implementation, we're requesting a process shutdown from the inside out.


So I think the process should be:

1. close this pr
1. get #199 and #200 merged
1. implement `exec/win32.ts`, `TestProcess#terminate()`, and `TestProcess#interrupt()` in a separate PR

If you want to get started, I'd base your changes off of the branch in #200

jbolda:
Sounds good. I started a branch a handful of days ago when I posted my comment, https://github.com/thefrontside/effection/pull/195#issuecomment-703206584, but I can merge in from your recent additions. I was sitting tight seeing how you were planning on dealing with the old APIs, and it looks like we are wholesale replacing and deprecating the old things.

RE: splitting things up, I will close this PR and open fresh based on your new APIs. Pretty minimal changes that are easy to port. For BigTest, I will have to check, but I think I've split out everything that isn't directly (or indirectly, eg ts-node not finding paths) related to child_process. I can start a new branch with the changes you have made to BigTest as well.

cowboyd (commented):
> I was sitting tight seeing how you were planning on dealing with the old APIs, and it looks like we are wholesale replacing and deprecating the old things.

Yeah, we were just trying to have `ChildProcess` be too many dang things: a one shot process, a daemon, cross platform and it was doing none of them well. This smaller abstraction was very easy to drop in and made things much, much simpler. I expect that there will be some duplication between `exec/win32.ts` and `exec/posix.ts` at first, but that 100% ok so long as we now have a completely separate adapter for windows that you can do whatever you want with.


> I can start a new branch with the changes you have made to BigTest as well.

Nice! I think that one is ready to go to so you should be able to base those changes of of that one https://github.com/thefrontside/bigtest/pull/591

cowboyd:
@jbolda thefrontside/bigtest#591 has been merged so the path should be completely cleared! Now we get to see how terribly our hypotheses perform IRL 🤣

jbolda:
Woo! Nice, I should be able to make some time this weekend to see if I can make some green windows builds.

packages/node/src/child_process.ts
jbolda: This was prettier running in vscode before I realized and turned it off. Completely agreed.

jbolda:
Closing in favor of #202.

PR #195, opened 2020-10-02
Introduce cross-platform process API

cowboyd:
Motivation
----------

We have had a lot of pain both in representing our processes in not only a cross platform way, but also trying to use a "one size fits all" approach for resource management. Sometimes you want to implicitly attach an error handler to a process and fail if it ever is triggered. Other times, you want to handle failure explicitly so you can do things like restart, etc.. Our existing process abstraction was too coupled to low-level node constructs that are already frustrating ad fiddly to begin with and so it seriously constrained what we could do.

Approach
----------

This takes a "less is more" approach of defining a very thin, effection-friendly API for process management that takes care of the following cases:

1. running a process to completion, and failing with an exception if that process does not complete.
2. running a process to completion and handling the output manually
3. running a process forever as you would a web server, and failing with an error if it exits for _any reason whatsoever_

For cases (1) and (2) this introduces the `exec()` operation which produces an instance of the `Process` api for finite length process that are expected to quit at some point. Key to the refactor is that errors in the process do not necessarily cause an error to be raised. Instead, they are collected into an `errors[]` array that is part of the process `ExitStatus` structure.

For Windows support, we take the strategy of allowing a completely parallel implementation of the `Process` API that is not (necessarily) shared between implementations. So when you call `exec()` windows implementation has complete freedom on how to manage the child process internally.

With `exec()` in place, we then implement a `daemon()` operation to handle case (3). For example the following code:

```ts
yield daemon("node http-server.js");
```

will spin up the command and will raise an error inside the operation context if the http server fails for whatever reason whatsoever. Naturally, processes created with both `exec()` and `daemon()` will be shut down if they ever pass out of scope.

TODOS & Open Questions
---------------------------

- [x] close stdout, stderr channels when a process exits. That way, any operations subscribed to those channels will complete
- [x] currently, we default `shell` to true so that we can pass whole commands. However this has a drawback of creating an extra bash or cmd.exe process which is annoying unix because you create two processes for every child you want to spawn, and downright debilitating on Windows since it opens a console window. By default we can parse the command with the [shellwords](https://www.npmjs.com/package/shellwords) package so that you can specify a command with a single string, but default to not having a child shell process.
- [x] The `ExitStatus` data structure has the `code` and `signal` properties which are unix-y. It's not possible for Windows to exit with a signal for example. Maybe we should just have an `ok` boolean on the `ExitStatus` interface, and then a `platformData` property that's like `PosixExitData | Win32ExitData`. That way, whatever platform-specific data is present to describe what happened on the exit is there for you to query, but in most people want to know if it was successful or not.
- [x] Different operations should be able to `join()` a process, but that won't work currently because of the way it is implemented. This PR currently contains a `Buffer` which is a more modern implementation of the `Mailbox` concept. However it's overkill in this case, and I think even incorrect. We can probably make do with a `Promise` instead. 
- [x] should we try this out in @bigtest/effection ?, put it in its own package like `@effection/exec` or just keep it in `@effection/node`


jbolda:
Couple thoughts on your own questions:

> It's not possible for Windows to exit with a signal for example.

Node does try to emulate them though so you will still get the signal.

> should we try this out in @bigtest/effection ?, put it in its own package like `@effection/exec` or just keep it in `@effection/node`

If we keep it in there, I am going to need to skip the graceful shutdowns on windows and convert the `main.test.ts` off `child_process` (in which case, do we switch to the new thing or update the old process spawning?

packages/node/src/daemon.ts
```diff
@@ -0,0 +1,41 @@
...
+  let stdin = new Channel<string>();
+  let stdout = new Channel<string>();
+  let stderr = new Channel<string>();
+
+  return yield resource({ stdin, stdout, stderr }, function*() {
+    let process: Process = yield exec(command, options);
+
+    yield spawn(subscribe(process.stdout).forEach(function*(chunk) {
+      stdout.send(chunk);
+    }));
```
jnicklas, line 23: We should make this a method on subscriptions, something like `pipe`.

jnicklas:
Regarding the questions above:

1) Using shellwords sounds like a good solution.
2) I think we need to build some kind of abstraction over the information returned from the process exit. As far as I know, Windows does support exit codes, right? So at least that part should be common across OS:es.
3) I agree that the `Buffer` solution is overengineered, and that a simpler solution would be applicable. Something like a Deferred might be enough.
4) The bad part about including this in `@bigtest/effection` is that that package is also used by the agent, and so packaging a bunch of node-specific things in there is not ideal. This is why `@bigtest/effection-express` is its own package. Developing this in the `effection` repo is easier, I think. One option is to add it to bigtest as `@bigtest/effection-process`, and then eventually move it to `effection` once we feel that we don't need to iterate as fast on it.

cowboyd:
@jbolda 
> Node does try to emulate them though so you will still get the signal.

I think it's best if we don't need to depend on that emulation at all. For example, if we end up implementing the `win32.ts` with a rust module, we would have to replicate that emulation in an API that's designed specifically for windows which feels weird. I'm leaning towards a boolean `ok` flag to indicate whether the exit was considered successful, and then a `platformData` which can contain _any_ platform specific exit data without the need to try and massage it into some uniform schema for what an exit entails. Most of the time, all you want to check is "did it work", but for the very rare cases where you need it, it'll be there.

> 
> If we keep it in there, I am going to need to skip the graceful shutdowns on windows and convert the `main.test.ts` off `child_process` (in which case, do we switch to the new thing or update the old process spawning?

My plan was to completely replace all instances of `ChildProcess` in bigtest with this new api, so it shouldn't be an issue. Theoretically, if you can add windows support via the `win32.ts` module during that time (it's currently implemented mostly as an alias for `posix.ts`) then it should be a drop-in upgrade.

cowboyd:
@jnicklas 
> 2. I think we need to build some kind of abstraction over the information returned from the process exit. As far as I know, Windows does support exit codes, right? So at least that part should be common across OS:es.

Even in this case it's a tad leaky because as far as I can tell, Windows _always_ has an exit code, whereas unix _maybe_ has an exit code because if the process terminates because of a signal, the exit code will be `null`. So really, the actual type of a Windows Exit is `{ code: number }`, and the type of a Unix exit is `{ code: number} | { signal: string}`

cowboyd:
@jnicklas Let me see if we can use it as-is here, and if we decide it's slowing down dev too much, I'll move it over to `@bigtest/effection-process`

cowboyd:
I've made the changes according to the suggestions. However, there is one key difference. If an `error` event is received it will actually cause the process to fail whenever it is `join()`d. This is because, at least on posix, the `error` event indicates that something is borked about the process at the `system` level, not the process level. In other words, the error is not from the process, but rather from OS trying to do something with the process.

From [the docs on the 'exit' event](https://nodejs.org/api/child_process.html#child_process_event_error):

```
The 'error' event is emitted whenever:

1. The process could not be spawned, or
2. The process could not be killed, or
3. Sending a message to the child process failed.
```

99% of the time the problem is (1). Thee process failed to even be created because there is an linkage error of a dynamic library, or the executable couldn't be found, etc.. In those cases, the `exit` is never fired at all which means that relying on it for join is not safe. You could be waiting forever on a process that was dead from the get go.

Therefore, if you are joining a process and it throws an error event, this will cause `join()` to throw an error. On the other hand, if the process decides that it is exiting in error, then `join()` will succeed with the error exit status.

packages/node/src/exec/api.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * The original command used to create the process
+   */
+  command: string;
+
+  /**
+   * The original options used to create the process.
+   */
+  options: ExecOptions;
```
jnicklas, line 79: How about moving these to the process, and instead providing a reference to the process in the exit status. So we can do `exitStatus.process.options`, this should allow even more flexibility.

packages/node/src/exec/buffer.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  signal = (value: T) => {
+    let next = this.waiters.pop();
+    if (next) {
+      next(value);
+    }
+  }
+
+  wait = () => new Promise<T>(resolve => this.waiters.push(resolve));
+}
```
jnicklas: Is this file used anymore?

packages/node/src/exec/deferred.ts
```diff
@@ -0,0 +1,21 @@
+export function Deferred<T>(): Deferred<T> {
+  let resolve: (value: T) => void = () => { return };
+  let reject: (error: Error) => void = () => {return };
+
+  let promise = new Promise<T>((_resolve, _reject) => {
+    resolve = _resolve;
+    reject = _reject;
+  }) ;
```
jnicklas: ✂️ extra space

packages/node/src/exec/posix.ts
```diff
@@ -0,0 +1,85 @@
...
+      }))
+
+      let value = yield once(childProcess, 'exit')
+      getResult.resolve({ type: 'status', value });
+    } finally {
+      try {
+        stdout.close();
+        stderr.close();
+        childProcess.off('error', onError);
+        process.kill(-childProcess.pid, "SIGTERM")
```
jnicklas, line 92: Can we add a comment here about why this is using `-pid`, since this isn't a commonly known thing.

packages/node/src/exec/buffer.ts
cowboyd: nope, torched it 🔥

packages/node/src/exec/posix.ts
cowboyd: Absolutely. I'll copy the comment from `ChildProcess` and also move the spawn into the resource body itself.

packages/node/src/exec/api.ts
cowboyd: The main issue I'd have with that is that it would convert what amounts to a simple value object into a holder for the stateful process object. I'm doubly nervous about hanging a live resource off of an object that could be handed around to other functions that could easily run outside the scope of the resource's context.

jnicklas: Fair point!

jnicklas (approved):

PR #197, opened 2020-10-07
🗑Deprecate ChildProcess `spawn()` and `spawn()`

cowboyd:
The new process management APIS `exec()` and `spawn()` provide a much more flexible interface for working with proceses that is both more effection-friendly as well as a better cross platform primitive. As such, the `ChildProcess` api is
redundant.

This uses the deprecation helper from `@effection/subscription` to indicate that this method will be going away soon, and helps to locate any place where it still resides in the code. Eventually, we may want to provide this from effection, but at this point I opted to just copy it.


taras (approved):
👍

PR #199, opened 2020-10-07
✨add pid to Process interface

cowboyd:
Motivation
-----------
In order to make our test cases cross platform, we need to be able to, from the outside in, send a signal to the process (or kill it using acustom windows module) In both of those cases, we need a process id.

Approach
----------
This adds the `pid` to the Process api and into the posix implementation. This did mean that the process had to be created outside of the resource body, but that should be ok.

### DIFF
--- a/.changeset/add-pid.md
+++ b/.changeset/add-pid.md
@@ -1,0 +1,4 @@
+---
+"@effection/node": minor
+---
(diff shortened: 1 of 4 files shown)


taras (approved):
👍

PR #202, opened 2020-10-12
✨Windows support for @effection/node 

### LINKED ISSUE #182: Problems spawning processes in Windows

jbolda:
This supersedes and pulls over much of what was learned in #186 on top of the new implementation of the process API.

This closes #182.

### DIFF
--- a/packages/node/src/exec/win32.ts
+++ b/packages/node/src/exec/win32.ts
@@ -1,7 +1,122 @@
-import { CreateOSProcess } from './api';
-import { createPosixProcess } from './posix';
+import { platform } from "os";
+import { Operation, resource, spawn } from "effection";
+import { Channel } from "@effection/channel";
+import { on, once } from "@effection/events";
+import { spawn as spawnProcess } from "cross-spawn";
+import { subscribe } from "@effection/subscription";
+import { ctrlc } from "ctrlc-windows";
+import { ExitStatus, CreateOSProcess, stringifyExitStatus } from "./api";
+import { Deferred } from "./deferred";
 
-//cheat to win!
-export const createWin32Process: CreateOSProcess = createPosixProcess;
+type Result =
+  | { type: "error"; value: unknown }
+  | { type: "status"; value: [number?, string?] };
 
-export const isWin32 = () => false;
+export const createWin32Process: CreateOSProcess = function*(command, options) {
+  let stdin = new Channel<string>();
+  let stdout = new Channel<string>();
+  let stderr = new Channel<string>();
+  let tail: string[] = [];
+
+  let getResult = Deferred<Result>();
+
+  function addToTail(chunk: string) {
+    if (tail.length < 100) {
+      tail.push(chunk);
+    }
+  }
+
+  function* join(): Operation<ExitStatus> {
+    let result: Result = yield getResult.promise;
+    if (result.type === "status") {
+      let [code, signal] = result.value;
+      return { command, options, code, signal, tail };
+    } else {
+      throw result.value;
+    }
+  }
+
+  function* expect(): Operation<ExitStatus> {
+    let status: ExitStatus = yield join();
+    if (status.code != 0) {
+      let error = new Error(stringifyExitStatus(status));
+      error.name = `ExecError`;
+      throw error;
+    } else {
+      return status;
+    }
+  }
+
+  let childProcess = spawnProcess(command, options.arguments || [], {
+    // We lose exit information and events if this is detached in windows
+    // and it opens a window in windows+powershell.
+    detached: false,
+    // When windows shell is true, it runs with cmd.exe by default, but
+    // node has trouble with PATHEXT and exe. It can't run exe directly for example.
+    // `cross-spawn` handles running it with the shell in windows if needed.
+    // Neither mac nor linux need shell and we run it detached.
+    shell: false,
+    // With stdio as pipe, windows gets stuck where neither the child nor the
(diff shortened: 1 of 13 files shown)


packages/node/src/exec/win32.ts
```diff
@@ -1,7 +1,119 @@
...
+          exec("taskkill /pid " + childProcess.pid + " /T /F");
+        } catch (e) {
+          console.error('error killing the task', e)
+        }
+      }
+    }
+  );
+};
+
+export const isWin32 = () => true;
```
cowboyd: We'll need this to be dynamic, not constant since this is used to determine which path to execute, posix, or win32. As-is, this will cause even posix systems to head down the windows path

packages/node/src/exec/win32.ts
```diff
@@ -1,7 +1,119 @@
...
+        childProcess.off("error", onError);
+        try {
+          // @ts-ignore
+          childProcess.stdout.end();
+          // @ts-ignore
+          childProcess.stderr.end();
+          // just in case, super kill the whole tree
+          // this isn't graceful but *shrug*
+          childProcess.kill('SIGTERM');
+          exec("taskkill /pid " + childProcess.pid + " /T /F");
```
cowboyd: If this rejects, it will cause an `UnhandledPromiseRejectionError`, so I think you'll need to just swallow the exception with a catch()

packages/node/test/exec.test.ts
```diff
@@ -83,9 +83,12 @@ describe('exec()', () => {
       await converge(() => expect(output).toContain("listening"));
     });
 
-    it('has a pid', () => {
+    it('has a pid', async () => {
       expect(typeof proc.pid).toBe('number');
       expect(proc.pid).not.toBeNaN();
+      // otherwise windows doesn't shut it down and we get
+      // EADDRINUSE: address already in use :::29000
+      await fetch('http://localhost:29000', { method: "POST", body: "exit" });
```
cowboyd: If windows doesn't shut it down, then doesn't that mean that it isn't working?

cowboyd (commented):
Nice! how is the new API feeling?

packages/node/test/exec.test.ts
jbolda: I believe it might be a race condition here?

packages/node/test/helpers.ts
```diff
@@ -54,9 +63,15 @@ export class TestProcess {
   // a users requests an interrupt via CTRL-C or equivalent.
   interrupt() {
     if (this.isWin32) {
-      //TODO:
+      // @ts-ignore
+      this.process.kill("SIGINT");
+      // @ts-ignore
+      this.process.stdout.end();
+      // @ts-ignore
+      this.process.stderr.end();
```
cowboyd: This is a f

cowboyd: ☝️  me trying to type a comment on my phone while lying in bed about 15 seconds after waking up and then falling back asleep 😂

packages/node/src/exec/win32.ts
```diff
@@ -1,7 +1,117 @@
...
+      } finally {
+        stdout.close();
+        stderr.close();
+        childProcess.off("error", onError);
+        try {
+          // @ts-ignore
+          childProcess.stdout.end();
+          // @ts-ignore
+          childProcess.stderr.end();
+          exec("taskkill /pid " + childProcess.pid + " /T");
```
cowboyd: If I understand correctly, `exec()` will return a promise here, and so we need to call `.catch()` on it so that in the event that it fails, we don't get a `UhandledPromiseRejection` warning.

packages/node/test/helpers.ts
```diff
@@ -43,9 +44,17 @@ export class TestProcess {
...
     if (this.isWin32) {
-      //TODO:
+      // @ts-ignore
+      this.process.stdout.end();
+      // @ts-ignore
+      this.process.stderr.end();
+      try {
+        ExecForTaskkill("taskkill /pid " + this.process.pid + " /T");
+      } catch (e) {
+        // swallow error
```
cowboyd: If we need to make terminate an async function so that you can await `ExectForTaskkill` so that you also swallow async errors, that's fine too.

cowboyd (commented):
Looking good! Remember to evict formatting changes from your commits. In a file with a bunch of changes, I usually review each hunk of a patch individually and explicitly stage it.

packages/node/src/exec/win32.ts
jbolda: `exec` takes a callback and returns a `<ChildProcess>`. We could use the promisified version, but I don't know that it would gain us anything.

> If this method is invoked as its util.promisify()ed version, it returns a Promise for an Object with stdout and stderr properties. The returned ChildProcess instance is attached to the Promise as a child property. In case of an error (including any error resulting in an exit code other than 0), a rejected promise is returned, with the same error object given in the callback, but with two additional properties stdout and stderr.

cowboyd: Cool. Just wanted to make sure we didn't generate a bunch of spam to `stderr`

cowboyd (approved):
🎉

PR #206, opened 2020-10-31
Add react support to effection

rauhryan:


cowboyd:
Hey, @rauhryan. I'm going to close this out because it's written against the `v1` API which is now obsolete. There is definitely paid work here if you are interested in bringing the v2 api into the React world, so get in touch if'n you're interested!

PR #208, opened 2020-11-02
Rename `main` to `run`

### LINKED ISSUE #130: Rename `main` method to `run` or `enter` (or something else)
Right now, there are two `main`s. One from `@effection/node` and another from `effection`, which is confusing. I think that `main` is the understood term as an entry point to an entire process, so `@effection/node` `main` should stay the same since that really is its purpose.

That means we really ought to come up with a better name for it in `effection` proper. I'm thinking that we should call it `run` or `enter`. It should convey that you are entering _an_ effection entry point, but not necessarily for the entire process.

jnicklas:
Closes #130


packages/effection/test.js
```diff
@@ -0,0 +1,6 @@
+let { main } = require('./dist/effection');
+
+main(function*() {
```
cowboyd: Was this file left in intentionally?

cowboyd (commented):
Looks good, barring what I think might be an accidental checkin for `effection/test.ts` (I didn't see a reference to it anywhere else) 👍 

We may want to consider extracting the deprecated function even further into `@frontside/javascript` at some later point.

packages/effection/test.js
cowboyd: @jnicklas ping

jnicklas: Nice catch 👌

cowboyd (approved):
👍

PR #209, opened 2020-11-03
Override exit handler for main

### LINKED ISSUE #203: promise returned by `main()`. reject on unexpected errors, resolve on expected errors
From https://github.com/thefrontside/bigtest/pull/627

>The way that this was implemented was to catch all non `MainError`s, but that felt not so great. It occurred to me that what might even be better would be to have the `Promise` returned from @effection/node` always resolve on both successful exits _and_ on `MainError` exits since these are still _expected_ outcomes. Only on an unexpected error would the promise actually reject. This would have the benefit of allowing custom error handling of unexpected exits, but baked in handling of all expected exits. In other words, we would have been able to write this as:

```ts
import { CLI } from './cli';
import { main } from '@effection/node';
import { unexpectedExit } from './unexpected-exit';

main(CLI(process.argv.slice(2)))
  .catch(unexpectedExit);
```

From the perspective of the `main()` operation, it would be consider a _successful_ exit when the operation either completes, or finishes with a `MainError` which indicates that the program raised the error. The only time the operation would fail would be on an unexpected error. Other things to consider:

- Don't return a raw `Context`, but a Promise implementor that has a reference to the context. That way, when you install a `catch` handler, it erases the default catch handler, or when you install a `then` handler, it erases the default tach handler. That way, there is full control about what to do when the process resolves.

jnicklas:
Makes it possible to set a custom error handler for `main`, allowing ultimate flexibility in how to handle errors.

I went with a slightly different approach here, instead of handling the error via `catch`, I added a callback instead. This felt simpler to me, and the behaviour is a little more predictable IMO.

Closes #203

### DIFF
--- a/.changeset/on-error.md
+++ b/.changeset/on-error.md
@@ -1,0 +1,5 @@
+---
+"@effection/node": "minor"
+---
+
+Can set onError handler for main, allowing maximum control over how to handle errors.
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -1,5 +1,5 @@
-export { main, MainError } from './main';
-export *  from './exec';
+export { main, exitOnError, isMainError, MainError } from './main';
+export * from './exec';
 export { daemon } from './daemon';
 
 export * as ChildProcess from './child_process';
--- a/packages/node/src/main.ts
+++ b/packages/node/src/main.ts
@@ -14,7 +14,26 @@ export class MainError extends Error {
   }
 }
 
-export function main<T>(operation: Operation<T>): Context<T> {
+export function isMainError(error: Error): error is MainError {
+  return error.name === 'EffectionMainError';
+}
+
+export function exitOnError(e: Error) {
+  if(isMainError(e)) {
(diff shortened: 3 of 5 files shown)


cowboyd (commented):
Hmm... I still think the promise api looks cleaner from the outside, even if it's a bit stranger to implement.

I think in the event that there is a `main` like convention in the JS ecosystem it is promise based. I see something like what you see in the `neon-cli` api script all the time:

https://github.com/neon-bindings/neon/blob/main/cli/bin/cli.js#L6-L8

Based on the history of Node, I think that when folks see a callback api there is the implicit assumption that it is a "errback" that is invoked always and `err` parameter is null on success. 

Even if we decided not to make a promise oriented api, I'd still prefer a chaining API:

`main(operation).onError(e => /* do stuff */)`

but at that point, might as well make it a promise?

cowboyd:
Closing for now. The main issue is still open, but we have a working solution with bigtest, so we can revisit it if we need to change that solution.

PR #216, opened 2020-11-17
Delete yarn.lock

taras:
## Motivation

image

I was going to fix the dependabot alerts and in the process discovered yarn.lock file in one of the workspaces.

## Approach

Remove the extra yarn.lock file and add `**/yarn.lock` to `.gitignore` so this doesn't happen again.


minkimcello (approved):

.gitignore
```diff
@@ -9,3 +9,4 @@ package-lock.json
 /scratchpad
 yarn-error.log
 .cache
+**/yarn.lock
```
cowboyd: This will match the root `yarn.lock` too, right? Maybe `*/**/yarn.lock` is what we want?

PR #217, opened 2020-11-18
Update tsdx to 0.14.1 and typescript 4.0.5

jnicklas:


jnicklas:
Well... this was more challenging than I expected 😅

cowboyd:
> Well... this was more challenging than I expected 😅

Yikes! Well if we're going to upgrade typescript, might as well use `@frontside/typescript`?

PR #222, opened 2021-02-26
Simplify sleep

jnicklas:
Makes the `sleep` operation a bit simpler by using a resolution primitive, rather than a promise.


taras (approved):
👍

PR #223, opened 2021-02-26
Update and fix packaging

jnicklas:
The `effection` package did not included the necessary files due to tsconfig files being missing. Also locks down the tsdx version to be identical and locked to 0.13.2 since we're using patch-package for that specific version.


taras (approved):
👍

PR #229, opened 2021-02-27
Tm/cancel timeout

taras:
## Motivation

@wKich noticed that timeoutId is not set when setTimeout is called. So, cancel is never called as a result.

## Approach

Assign timeoutId when setTimeout is called


jnicklas:
@taras see also #222

PR #231, opened 2021-03-01
Make generators reentrant

jnicklas:
A task may be currently running a generator, that is it has entered the generator and is currently inside the generator itself in between two yield points.

Since resume/reject and halting is now fundamentally synchronous, it could happen that the code in the generator itself causes the task to halt, either by explicitly halting itself, or by causing a sibling or parent task to fail and thereby being halted by the parent.

However, this poses a problem, since the iterator is already running, we cannot call `return()` on it. This will cause a `TypeError: Generator is already running`.

Effectively we cannot enter a generator if we have already entered the generator.

This fixes this problem by adding a cache of continuations and delaying their execution if we're already inside the generator.

### DIFF
--- a/packages/core/src/controller/iterator-controller.ts
+++ b/packages/core/src/controller/iterator-controller.ts
@@ -5,9 +5,14 @@ import { HaltError } from '../halt-error';
 import { Operation } from '../operation';
 import { Trapper } from '../trapper';
 
+type Continuation = () => IteratorResult<Operation<unknown>>;
+
 export class IteratorController<TOut> implements Controller<TOut>, Trapper {
   private didHalt = false;
+  private didEnter = false;
   private subTask?: Task;
+
+  private continuations: Continuation[] = [];
 
   constructor(private controls: Controls<TOut>, private iterator: OperationIterator<TOut>) {
   }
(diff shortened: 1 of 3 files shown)


jnicklas:
@cowboyd I ran into this issue, which manifested itself in `@effection/subscription`. I recall that we had this issue before, but I don't remember how we solved it. We should investigate this and figure out a good solution.

packages/core/src/controller/iterator-controller.ts
```diff
@@ -17,7 +22,30 @@ export class IteratorController<TOut> implements Controller<TOut>, Trapper {
...
+      this.step(iter);
+      this.didEnter = false; // release lock
+
+      // resume with stored continuation it happened while we were in the step
+      let continuation = this.continuations.shift();
+      if(continuation) {
+        this.resume(continuation)
+      }
+    }
+  }
```
jnicklas: This definitely makes the iterator controller a lot less "elegant", but it is the simplest fix that I could come up with.

cowboyd: @jnicklas I wonder if we can convert this to a loop that is just draining the continuations until there are none?

jnicklas: Yeah, making this recursive is maybe a bit strange. I did not want to use `forEach` to prevent problems with the collection being modified during iteration, but a while loop would totally work as well.

jnicklas: Ahh, I think I see what you're saying, basically you want something like this:

``` typescript
  resume(iter: Continuation) {
    this.continuations.push(iter);
    let continuation;
    while(continuation = this.continuations.shift()) {
      this.step(continuation);
    }
  }
```

Unfortunately this doesn't work, since we need to somehow ensure only one of these loops is running at a time.

Basically we need to build a really crappy mutex here 😆

jnicklas: This is an alternate implementation, I'm not sure if this is better, what do you think?

``` typescript
  resume(iter: Continuation) {
    this.continuations.push(iter);
    if(!this.didEnter) {
      this.didEnter = true; // acquire lock
      let continuation;
      while(continuation = this.continuations.shift()) {
        this.step(continuation);
      }
      this.didEnter = false; // release lock
    }
  }
```

jnicklas: I like that implementation better, let's go with that!

cowboyd (approved):

PR #233, opened 2021-03-02
Add Subscribable and use it in subpackages

jnicklas:
Depends on #231 

This adds a `Subscribable` class, and uses it for `on` and `Channel`. It also adapts the other packages to use this new interface.

## Motivation

We noticed that in effection v2, working with subscriptions directly is somewhat cumbersome. The normal case is that the subscriptions will be "drained" through either `forEach` or `first` or similar. Since `forEach` uses internal iteration, the scope of the iterator can be constrained to the task that the `forEach` runs into. However, creating an iterator requires a task. If we create the iterator and then drain it, we need to pass in a task. If we do this all in one step we don't need to pass in this task. This makes the API much cleaner and easier to understand.

## Approach

We add a `Subscribable` interface and a `createSubscribable` function which creates subscribables. The hypothesis is that manually creating subscribables is rare, and that it should usually only be done for primitive structures (e.g. Channel). Therefore the user will get a convenient interface without having to think too much about the intrecacies of this system.

Here's an example of what the current API looks like:

``` typescript
task.spawn(on<[string]>(task, childProcess.stdout, 'data').forEach(([data]) => function*() {
  // ...
}));
```

Note the `task` passed to `on`. Here is the same code with the new API:

``` typescript
task.spawn(on<[string]>(childProcess.stdout, 'data').forEach(([data]) => {
  // ...
}));
```

### Alternate Designs

The status quo is an alternative, where we double down on working with `Subscription` rather than `Subscribable`, but for reasons explained above this feels inconvenient.

Subscriptions are naturally race condition proof, whereas subscribables are not. Some care needs to be taken that no events/messages are lost before subscribing. However, this risk is lessened in effection v2, since task spawning no longer is an operation.

### TODOs and Open Questions

- [ ] Should we still have `Subscription` at all? That is should we have some object with higher level operations like `map` and `filter` for an actual subscription object, rather than a subscribable object? At the moment we feel that having such an object is probably *not* necessary, but more real-world experience is necessary.

### DIFF
--- a/.changeset/bright-terms-rescue.md
+++ b/.changeset/bright-terms-rescue.md
@@ -1,0 +1,6 @@
+---
+"@effection/events": minor
+"effection": minor
+---
+
+Change `on` to return subscribable, rather than taking scope
--- a/.changeset/sharp-penguins-check.md
+++ b/.changeset/sharp-penguins-check.md
@@ -1,0 +1,6 @@
+---
+"@effection/subscription": minor
+"effection": minor
+---
+
+Add `createSubscribable` and `Subscribable` interface
--- a/.changeset/shiny-terms-talk.md
+++ b/.changeset/shiny-terms-talk.md
@@ -1,0 +1,6 @@
+---
+"@effection/channel": minor
+"effection": minor
+---
+
+Make Channel subscribable and add all subscribable methods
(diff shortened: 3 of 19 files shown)


taras:
@jnicklas would it be possible to add before/after code example to the description?

jnicklas:
@taras I added an example to the description!

taras:
@jnicklas thank you. That helps a lot!

PR #235, opened 2021-03-02
Make channel ends splittable

jnicklas:
Depends on #233 

## Motivation

We have often had a need for a channel where the sending and receiving ends are separate. Often an API will want to expose the sending end but keep the receiving end private, or vice versa.

## Approach

A very minimal solution where we add a property called `receive` which only returns the subscribable. This allows us then to do this:

``` typescript
let { send, receive } = createChannel();
```

Giving us a very simple way of splitting the ends.

### Alternate Designs

We could add a separate thing for this e.g. `splitChannel`. We could also always split the ends like this:

``` typescript
let [send, receive] = createChannel();
```

But this solution feels both more general and more flexible.

None

### DIFF
--- a/.changeset/mighty-snakes-shop.md
+++ b/.changeset/mighty-snakes-shop.md
@@ -1,0 +1,6 @@
+---
+"@effection/channel": minor
+"effection": minor
+---
+
+Make channel splittable into send and stream ends
--- a/packages/channel/src/channel.ts
+++ b/packages/channel/src/channel.ts
@@ -9,6 +9,7 @@ export type ChannelOptions = {
 export interface Channel<T, TClose = undefined> extends Stream<T, TClose> {
   send(message: T): void;
   close(...args: TClose extends undefined ? [] : [TClose]): void;
+  stream: Stream<T, TClose>;
 }
 
 export function createChannel<T, TClose = undefined>(options: ChannelOptions = {}): Channel<T, TClose> {
@@ -30,7 +31,9 @@ export function createChannel<T, TClose = undefined>(options: ChannelOptions = {
     }
   });
 
-  return Object.assign(subscribable, {
+  return Object.assign({
+    stream: subscribable,
+
     send(message: T) {
       bus.emit('event', { done: false, value: message });
     },
@@ -38,5 +41,5 @@ export function createChannel<T, TClose = undefined>(options: ChannelOptions = {
     close(...args: TClose extends undefined ? [] : [TClose]) {
       bus.emit('event', { done: true, value: args[0] });
     }
-  });
+  }, subscribable);
 }
--- a/packages/channel/test/channel.test.ts
+++ b/packages/channel/test/channel.test.ts
@@ -45,6 +45,18 @@ describe('Channel', () => {
     });
   });
 
+  describe('with split ends', () => {
+    it('receives message on subscribable end', function*(task) {
+      let { send, stream } = createChannel();
+
+      let subscription = stream.subscribe(task);
+
+      send('hello');
+
+      expect(yield subscription.next()).toEqual({ done: false, value: 'hello' });
+    });
+  });
+
   describe('close', () => {
     describe('without argument', () => {
       it('closes subscriptions', function*(task) {


cowboyd:
We can actually have both:

```js
let channel = {
  send: new Object(),
  receive: new Object(),
  [Symbol.iterator]() { return [this.send,this.receive][Symbol.iterator]() }
}

let { send, receive } = channel;
let [send2, receive2] = channel;
send === send2 //=> true
receive === receive2 //=> true
```

This is a pretty common pattern that you see in the builtin  `Map` entries as both:

```js
for (let { key, value } of map.entries()) {};
```
and
```js
for (let [key,value] of map.entries()) {}
```
both work.

Also, I think that there are some advantages to using the array destructuring in the sense that you can name your channel different things without having to do cumbersome aliasing. Let's say that I want to call the broadcast portion of the channel `clientMessages`. We have to jump through some special hoops to re-alias if we have to do record destructuring:

```js
let { send, receive: clientMessages } = createChannel();
```

vs

```js
let [send, clientMessages] = createChannel();
```

The browser MessageChannel api uses this convention which is why you see:

```js
let [input, output] = MessageChannel();
```
More often than
```js
let {port1: input, port2: output} = MessageChannel();
```

cowboyd (approved):

jnicklas:
I realized that there is a big problem with this. Unfortunately, all of the array methods follow along with the array, so we end up with a whole bunch of array methods on the channel itself.

Screenshot 2021-03-03 at 19 03 05

This is especially confusing since some methods, like `forEach` and `map` actually exist on both.

Destructuring seems to only require `Iterable` rather than being an array, but as we saw today, the iterable protocol cannot be typed correctly.

Unfortunately this means that we cannot type this correctly in TypeScript, which leads me to think that we shouldn't provide an array split at all.

cowboyd:
nooooo

Well that sucks, but I concur. Were it not for the `close()` method, I think I would come down on the side of array splitting as _the_ way to go, and let those apis that needed to "restructure" channels. But since we do have the concept of closing a channel, it means we need more interface than just `Function` and `Subscribable`

I'm still curious though about ways to encourage preference for `Subscribable` over `Channel`.  Other alternatives that occur to me are:

```js
let [send, receiver, close] = createChannel<string>();
```

```js
let [{send, close}, receiver] = createChannel<string>();
```

Or, we could put the `close()` function on `send()` function:

```js
let [send, receiver] = createChannel<string>();
send.close()
```

Here's where we're using channels  currently https://github.com/thefrontside/bigtest/search?q=%22new+Channel%22 I'm curious how they feel with each. I didn't really find one where the entire channel was needed to be passed around, and shouldn't we be encouraging people to pass around Subscribables and OperationIterables rather than any other more complex interface? In other words, we should discourage accepting and returning the `Channel` interface from within user code. Which makes me wonder if in the event that you _do_ want to pass around all of the channel artifacts, maybe you should just have a custom, one-off interface or rely on TS structural typing.

cowboyd:
As an aside, it turns out that we are already using the "signal" case of no-arg channel publication for tracking file changes. Currently we just use `Channel<void>`  https://github.com/thefrontside/bigtest/blob/v0/packages/server/src/manifest-generator.ts#L80

I think it's fine to be forced to explicitly mark this however with the `void` type param however.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/bigtest">thefrontside/bigtest</a></strong></div><div>Tests that speed up development, not slow it down. Contribute to thefrontside/bigtest development by creating an account on GitHub.</div></blockquote>

jnicklas:
What I like about destructuring as records is that it allows the flexibility to do all of:

``` typescript
let myChannel = createChannel();
let { send, receive } = createChannel();
let { send: mySend, receive: myReceive } = createChannel();
let { send, receive, close } = createChannel();
```

Whereas some of these are not as easy to express with the array form.

jnicklas:
@cowboyd I updated this to use the new `Stream` naming and removed the array splitting so it becomes:

``` typescript
let { send, stream } = createChannel()
```

IMO this is much better than `receiver` and makes the record based API a little less awkward. It resolves the naming weakness of the record API while also preserving the following advantages:

1. Ability to keep the channel "together" when used internally.
2. A more natural API for `close`
3. Less awkward when used as a duplex channel/pipe since it doesn't require double array destructuring

The downside still remains that in many cases we want to encourage people to split the channel and to only hand out the stream ene.

cowboyd (approved):

PR #236, opened 2021-03-02
Add duplex channels

jnicklas:
Depends on #235 

## Motivation

It is pretty useful to have channel pairs that are tied to each other. We are already using this pattern inside bigtest, and it is pretty simple to implement on top of our new channels.

## Approach

Since we can destructure channels, we can create two channels and mix and match their methods.

### Alternate Designs

We could have separate `TClose` types for either ends? I can't see a usecase for that though.

Adds complexity and another thing to maintain

### DIFF
--- a/packages/duplex-channel/package.json
+++ b/packages/duplex-channel/package.json
@@ -1,0 +1,48 @@
+{
+  "name": "@effection/duplex-channel",
+  "version": "2.0.0-beta.1",
+  "description": "A bidirectional channel for effection",
+  "main": "dist-cjs/index.js",
+  "module": "dist-esm/index.js",
+  "types": "dist-esm/index.d.ts",
+  "typeDocEntry": "src/index.ts",
+  "sideEffects": false,
+  "homepage": "https://github.com/thefrontside/effection",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/thefrontside/effection.git",
+    "directory": "packages/channel"
+  },
+  "author": "Frontside Engineering <engineering@frontside.io>",
+  "license": "MIT",
+  "files": [
+    "README.md",
+    "CHANGELOG.md",
+    "dist-*/**/*",
+    "src/**/*"
+  ],
+  "scripts": {
+    "lint": "eslint '{src,tests}/**/*.ts'",
+    "test": "mocha -r ts-node/register test/**/*.test.ts",
+    "prepack": "tsc --build tsconfig.esm.json && tsc --build tsconfig.esm.json && tsc --build tsconfig.cjs.json",
+    "mocha": "mocha -r ts-node/register"
+  },
+  "devDependencies": {
+    "@effection/mocha": "^2.0.0-beta.9",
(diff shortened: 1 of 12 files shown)


cowboyd:
Let's pull out `WriteStream` interface on the hypothesis that it is the only primitive that we'll need. If it turns out that in fact, we do need it, then we can add it quite quickly because it's such a simple thing to build on top of write streams.

cowboyd:
This hasn't come up yet as being a hard requirement, so we're going to convert this into a draft.

jnicklas:
I've revived this PR, this time with duplex channel being a separate package. I think this is a better way forward, since it doesn't commit us as strongly to maintaining this in the future. The main reason to get rid of this is so that we can move the last remaining effection piece from bigtest to the effection repo.

packages/duplex-channel/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "expect": "^25.4.0",
+    "mocha": "^8.3.1",
+    "ts-node": "^10.1.0",
+    "typescript": "^4.3.5"
+  },
+  "dependencies": {
+    "@effection/core": "^2.0.0-beta.10",
+    "@effection/channel": "^2.0.0-beta.12",
+    "@effection/subscription": "^2.0.0-beta.12"
+  },
```
cowboyd, line 44: I'm beginning to feel like these "addon" packages like "atom", "fetch", "websocket-client", etc.. should have `effection` as a peer dependency, but that "core" packages have a pinned versions to each other in normal dependencies. With `npm` and `yarn2` now installing a matching peer dependency if it is not present, this should make managing your lock file simpler.

At the very least, I think it is safe to have it depend on a single `effection` version, since it isn't core.

packages/duplex-channel/src/duplex-channel.ts
```diff
@@ -0,0 +1,27 @@
+import { Stream } from '@effection/subscription';
+import { createChannel, ChannelOptions } from '@effection/channel';
+
+type Close<T> = (...args: T extends undefined ? [] : [T]) => void;
+
+export interface DuplexChannel<TReceive, TSend, TClose = undefined> extends Stream<TReceive, TClose> {
+  send(message: TSend): void;
+  close: Close<TClose>;
```
cowboyd, line 8: We'd talked about making these operations. Any reason why they can't be?

packages/duplex-channel/package.json
jnicklas: Hmm, that's a good point. Is this something we want to tackle for this PR, or something we want to adjust for all of these packages later on?

packages/duplex-channel/src/duplex-channel.ts
jnicklas: This just matches the current API of Channel. If we want to change this, then we should change it there first, I think.

packages/duplex-channel/package.json
cowboyd: I’m fine with making it follow-on.

cowboyd (approved):

PR #237, opened 2021-03-02
Rexport blessed packages

jnicklas:
This is a conservative rexport of the most useful packages, we can add further ones in the future.

### DIFF
--- a/.changeset/three-actors-search.md
+++ b/.changeset/three-actors-search.md
@@ -1,0 +1,5 @@
+---
+"effection": minor
+---
+
+Reexport blessed packages from `effection` package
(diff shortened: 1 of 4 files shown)


cowboyd:
The only package here that is a bit iffy in my mind is `@effection/channel`. I wonder if this one will survive the renaming and splitting apart.

jnicklas:
> The only package here that is a bit iffy in my mind is `@effection/channel`. I wonder if this one will survive the renaming and splitting apart.

We can remove this again later. Until we do the actual release of 2.0 this is not a firm commitment, IMO.

cowboyd (approved):

PR #245, opened 2021-03-04
Use shield style of status badge

jnicklas:
It looks nicer


cowboyd (approved):
I agree. I guess it's good for their brand, but shield really ought to be the default

PR #248, opened 2021-03-05
Rename Subscribable to Stream

jnicklas:
This is an interface which will appear in public interfaces a lot, let's give it a good name.


cowboyd (approved):
I'm in. Subscribable was always a source of confusion. I think `StreamIterator` has as nice ring to it too ✨

PR #253, opened 2021-03-09
Make event emitter api more convenient.

cowboyd:
## Motivation

99.9% of the time when you use the `once()` method, or the `on()` method, you're just looking to get the first argument passed to the `EventEmitter#emit()` method. This is because 99.9% of the time, there is only ever one argument. That means that almost everywhere in our codebase, you see something like:

```ts
on<string[]>(source, 'data').map(function*([chunk]) { return chunk }).forEach(...)`
````

or

```ts
let [event]: [Event] = yield once(source, 'event');
```

This current API requires insta-cruft be attached to  almost every single usage.

## Approach

This changes the `on()` stream and `once()` operation to only work on the first argument to `EventEmitter#emit()` which makes the mapping functions that appear everywhere and the de-structuring syntax that appears alongside every usage unnecessary. So the above examples become:

```ts
on<string>(source, 'data').forEach(...)`
````

and

```ts
let event: Event = yield once(source, 'event');
```

Not only is it less code, the type parameters are far less confusing.

For those cases where you do actually need the arguments (I actually can't think of any). There is the `onceEmit()` and `onEmit()`operations which behave the same as before.

### DIFF
--- a/.changeset/emmitter-stream.md
+++ b/.changeset/emmitter-stream.md
@@ -1,0 +1,11 @@
+---
+"@effection/events": patch
+"effection": patch
+---
+`once()` only yields the first argument passed to `emit()` which
+accounts for 99.9% of the use cases. For the cases where all the
+arguments are required, use `onceEmit()`
+
+`on()` produces a stream of the first arguments passed to `emit()`
+which accounts for 99.9% of the use cases. For the cases where all the
+arguments are required, use `onEmit()`.
--- a/packages/channel/src/channel.ts
+++ b/packages/channel/src/channel.ts
@@ -21,7 +21,7 @@ export function createChannel<T, TClose = undefined>(options: ChannelOptions = {
   let subscribable = createStream<T, TClose>((publish) => function*(task) {
     let subscription = on(bus, 'event').subscribe(task);
     while(true) {
-      let { value: [next] } = yield subscription.next();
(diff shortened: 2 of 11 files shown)


jnicklas (changes requested):
Looks good, and I agree, this is one of those things where it almost always annoying to do it the "right" way. I do like this solution of having a separate method for the rare case where you *do* need all the arguments. `onEmit` is not a terribly descriptive name, but I think that's okay.

I think it would be good to add some tests for this though.

cowboyd:
@jnicklas Added some tests (and also fixed the other breakages that happened elswhere around the codebase because of this change)

As for the name, I couldn't think of anything more descriptive that didn't also feel terribly cumbersome. E.g. `onEmitArgumentsAsArray()`.

🤷🏻‍♂️ 

If you have any suggestions, I'm definitely open to them.

jnicklas (approved):

jnicklas:
I'm fine with `onEmit`

cowboyd:
@jnicklas I added some pretty thorough comments to these methods since they're workhorses that have been with us a long time and will almost certainly going to stick with us. I figured if we can't come up with a terribly descriptive name for `onEmit()` and `onceEmit()` we can at least describe exhaustively what they do.

I've used the same documentation style here as in `@bigtest/interactor`

PR #254, opened 2021-03-09
Add `buffer` method to Stream

jnicklas:
This method can be used to buffer a stream and returns a new stream which replays all messages from the point in time that the stream was created. This is useful for ensuring that messages are not missed if the moment of subscription is not tightly controller. In BigTest we're aiming to replace the mailbox pattern used in the result aggregator with this.


cowboyd (approved):
That method is just pleasant to look at. Perhaps at some later time we can expand it to take parameters setting a max length, etc..

PR #255, opened 2021-03-09
Add `stringBuffer` method to `Stream`

jnicklas:
Depends on #254 

This method retains a string which concatenates all previously emitted values to each other. This should normally be used on streams of strings or node buffers, however we cannot constrain this method to only work on these types (unless we make it a free function instead). This type of stream is useful for use with node io streams, to aggregate their results in memory. We will be able to use this with the io streams in `@effection/node` to make working with them far easier.

### DIFF
--- a/packages/subscription/src/stream.ts
+++ b/packages/subscription/src/stream.ts
@@ -18,6 +18,7 @@ export interface Stream<T, TReturn = undefined> extends OperationIterable<T, TRe
   toArray(): Operation<T[]>;
   subscribe(scope: Task): OperationIterator<T, TReturn>;
   buffer(scope: Task): Stream<T, TReturn>;
+  stringBuffer(scope: Task): Stream<string, TReturn>;
 }
 
 export function createStream<T, TReturn = undefined>(callback: Callback<T, TReturn>): Stream<T, TReturn> {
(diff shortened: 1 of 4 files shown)


packages/subscription/src/stream.ts
```diff
@@ -114,6 +119,28 @@ export function createStream<T, TReturn = undefined>(callback: Callback<T, TRetu
...
+          internalBuffer += `${m}`;
+          publish(internalBuffer);
+        });
+      });
+
+      return {
+        ...stream,
+        get value(): string {
+          return buffer;
+        }
```
cowboyd, line 145: How do we even use this from TypeScript?

jnicklas: It uses an extended interface, see here: https://github.com/thefrontside/effection/pull/255/files#diff-0fd53fb3d32144b53a5c547ff0713daaee9dd8428e8502bfb9a8f0b720f7e573R23-R25

I have some thoughts about this though, I think there might be a more comprehensive/powerful solution available here. We could have something like a `ValueStream<T>` which is more general than a type specific for strings. Some methods on this, like `map` could then return a new value stream. We could also add a `unique` method on this type of stream. We would be able to implement this interface not just for buffered streams, but also for Atom for example.

jnicklas: Interestingly this leads back to an idea I had a long, long time ago about client side frameworks. This concept is similar to Ember Octanes change tracking, and even before that came out I was tinkering with that. If you squint a little bit it's also similar to React's `useState` hook, except a lot less weird ;)

cowboyd: Right, but the `Stream#stringBuffer` interface returns a `Stream`. The test itself never uses the `value` property, only the accumulated value on the stream iterator. It's not a biggie, but I don't see a way that we can actually get a reference to a `StringBufferStream` without down-casting.

cowboyd (approved):

packages/subscription/src/stream.ts
jnicklas: Oh duh! I totally missed that, and yes we should totally have a test for this too!

PR #256, opened 2021-03-10
Add Atom package from bigtest

jnicklas:
This imports the atom package from BigTest into the repo and makes it work with v2 effection. The ability to reset an atom has been removed.

### DIFF
--- a/packages/atom/src/types.ts
+++ b/packages/atom/src/types.ts
@@ -1,0 +1,112 @@
+import { Stream } from '@effection/subscription';
+import { Operation } from 'effection';
+
+export interface Slice<S> extends Stream<S> {
+  get(): S;
+  set(value: S): void;
+  update(fn: (state: S) => S): void;
+  slice: MakeSlice<S>;
+  once(predicate: (state: S) => boolean): Operation<S>;
+  remove(): void;
+  stream: Stream<S>;
+}
+
+export interface MakeSlice<S> {
+  /*
+  * Brute foce overload to allow strong typing of the string path syntax
+  * for every level deep we want to go in the slice there will need to be an overload
+  * with an argument for each of the string paths
+  *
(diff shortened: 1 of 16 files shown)


cowboyd:
Funny, looking at it now, if we aliased the `set()` method to `send()`, `Slice` would also be a `WritableStream` 🤔

packages/atom/src/types.ts
```diff
@@ -0,0 +1,112 @@
+import { Stream } from '@effection/subscription';
```
cowboyd, line 1: Even though this is a core package, should we import from `effection`? Feels like when we deprecate subscription, it'll be one less point of coupling. However, I can also see that in order to avoid circular dependencies we could have a policy that core package depend on core packages explicitly.

Or maybe the rule is "blessed packages should depend directly on other blessed packages", which in this case, means that atom could pull from the omnibus `effection` package since it is not a blessed package (and is unlikely to be one)

jnicklas: Actually in this PR I did add `atom` as a blessed package, but maybe it shouldn't be?

IMO all packages in the monorepo should depend on the individual subpackages, regardless of whether they are blessed or not.

jnicklas:
@cowboyd is this good to go? Should `atom` be a blessed package?

packages/atom/src/types.ts
cowboyd: I'd say that this is not a blessed package as you will only need it a certain amount of time that is likely to be < 50%

👍🏻 on the policy of importing from individual packages within the repo.

cowboyd:
I think this is pretty much good to go, but I think atom should _not_ be a blessed package.

jnicklas:
@cowboyd I removed it as blessed package

cowboyd (approved):
😍

PR #260, opened 2021-03-11
Simplify Atom implementation and remove dependencies

jnicklas:
Our Atom has always been based on lenses and functional programming. However, with the advent of Streams in effection, I have been thinking about a simpler implementation of Atom. I realized that we could make an implementation which stacks slices on top of each other, rather than accessing the state directly. Effectively our slice becomes sort of like a lens itself, except with the addition of a stream.

Since there were some chicken-and-the-egg problems with making slices, I introduced `Container` as a simplified slice, this also makes it easy to attach all of the additional method for slice which are not needed for the traversal.

In my opinion the resulting implementation is easier to understand, and also means we do not need to depend on any external packages in `@effection/atom`.

The resulting code does make rather judicious use of `any` to skip around the inherent difficulty in making a type safe lens-like thing in TypeScript. However, since the type safety of Slices is provided by the `MakeSlice` type, there is no loss in external type safety.

### DIFF
--- a/packages/atom/test/slice.test.ts
+++ b/packages/atom/test/slice.test.ts
@@ -1,0 +1,222 @@
+import { describe, it, beforeEach } from '@effection/mocha';
+import * as expect from 'expect';
+import { createAtom } from '../src/atom';
+import { Stream, OperationIterator } from '@effection/subscription';
+import { Slice } from '../src/types';
+
+type Data = { data: string };
+
+describe('@bigtest/atom Slice', () => {
+  describe('with no data', () => {
+    let atom: Slice<{ outer: Data }>;
+    let slice: Slice<string>;
+
+    beforeEach(function*() {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      atom = createAtom(undefined as any);
+      slice = atom.slice('outer', 'data');
+    });
+
+    it('should not blow up with no state and get', function*() {
+       expect(slice.get()).toBeUndefined();
+    });
+
+    it('should not blow up with no state and set', function*() {
+      slice.set('houston we have a problem');
+
+      expect(slice.get()).toBeUndefined();
+    });
+  });
+
+  describe('with data', () => {
+    let atom: Slice<{ outer: Data }>;
+    let slice: Slice<Data>;
+
+    beforeEach(function*() {
+      atom = createAtom({ outer: { data: "baz" } });
+      slice = atom.slice('outer');
+    });
+
+    it('should return the slice data', function*() {
+      expect(slice.get()).toEqual({ data: 'baz' });
+    });
+
+    it('should set the slice and atom', function*() {
+      slice.set({ data: 'bar' });
+
+      expect(slice.get()).toEqual({ data: 'bar' });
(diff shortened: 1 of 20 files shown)


packages/atom/test/slice.test.ts
```diff
@@ -24,7 +24,7 @@ describe('@bigtest/atom Slice', () => {
     it('should not blow up with no state and set', function*() {
       slice.set('houston we have a problem');
 
-      expect(slice.get()).toBeUndefined();
+      expect(slice.get()).toEqual('houston we have a problem');
```
jnicklas: Why should this be `undefined`? That doesn't seem right to me?

packages/atom/test/slice.test.ts
```diff
@@ -216,7 +216,7 @@ describe('@bigtest/atom Slice', () => {
 
       expect(atom.get()).toEqual(['foo', 'bar']);
       slice.remove();
-      expect(atom.get()).toEqual([undefined, 'bar']); // TODO: shouldn't this remove the item entirely?!
+      expect(atom.get()).toEqual(['bar']);
```
jnicklas: There is a case to be made for either behaviour here. The value of the slice will now have changed to `bar` which is a bit odd.

taras (commented):
I can the see the value of removing the external dependency. How much are we giving up by the *liberal* use of `any` inside of the implementation?

cowboyd (commented):
I can see why we could opt to go with a simpler, hand-rolled solution, but my guess is that we'd have to go back to lenses once we start to do things like composing references from different locations in the atom and presenting them as a single slice. This is something likely to come up sooner rather than later in the simulator. One example is a people table, and  as devices table, and I want to create a location slice that pairs people and devices with a lat-long data.

While `fp-ts` can be quite inscrutable, there is nevertheless a lot of prior art on how to do these things with lenses, and taking ownership of these complexities in a way that remain performant is not something I'm keen to do.

That said, I do feel like there are certain things about our implementation that do not make sense though and the awkwardness in typing might be indicative of that. For example, `Slice<T>.get()` returning `undefined` is an indication of type incoherence in my book. While we don't use arrays now, we do need some way of having creatable and removeable data, even if that is represented by a id-value pairs.

One thing that occurs is that if a record is deleted, the stream for it should end.

jnicklas:
@cowboyd I see your point on this. Composing slices is something I hadn't even thought about, and that definitely sounds like a can of worms that we may not want to eat ourselves.

Let's hold off on this for a while and go with the fp-ts based implementation for now.

jnicklas:
> I can the see the value of removing the external dependency. How much are we giving up by the _liberal_ use of `any` inside of the implementation?

There are actually some uses of `any` in the existing implementation as well, and they work mostly around the same issues. There are some ways to increase the type safety of the `makeArrayContainer` and `makeRecordContainer` functions, but I don't actually think that they help that much in practice. Basically instead of using `Container<any>` we'd use a  `Container<Array<any>` instead, there's still an `any` in there but it's slightly better type safety inside, unfortunately this also leads to a ton of casts everywhere, which made the code much harder to read for marginal gain, IMO.

PR #267, opened 2021-03-15
Publish with beta tag for v2 releases

minkimcello:
## Motivation

v2 changes are being published with the latest tag so anyone that adds effection to their package will get v2 unless they specify otherwise.

## Approach

- ~Disabled release workflow for v2 by commenting out the whole workflow~
- Specify `yarn publish --tag v2-beta` command for the release workflow.

## TODOs

- [x] ~We need to do a release of v1 to take back the latest tag.~ Push the latest tags on all of the current versions of v1 effection packages.
  - [x] @effection/channel @ 1.0.0
  - [x] effection @ 1.0.0
  - [x] @effection/events @ 1.0.0
  - [x] @effection/fetch @ 1.0.0
  - [x] @effection/node @ 1.0.1
  - [x] @effection/subscription @ 1.0.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,21 +1,21 @@
-# name: Release
+name: Release
 
-# on:
-#   push:
-#     branches:
-#       - 'v[0-9]+'
+on:
+  push:
+    branches:
+      - 'v[0-9]+'
 
-# jobs:
-#   publish-releases:
-#     name: Publish Releases
-#     runs-on: ubuntu-latest
-#     steps:
-#     - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


cowboyd (changes requested):
Rather than comment out, let's delete it. If we need it, it's in the git history.

minkimcello:
> Rather than comment out, let's delete it. If we need it, it's in the git history.

Jonas pointed out that disabling the release workflow would stop the pre-release publishing. So I enabled the release workflow and added a tag flag to the publish command.

cowboyd (approved):

PR #269, opened 2021-03-16
New, more ergonomic `exec` API

jnicklas:
This changes the `exec` API to not take a `scope` anymore. Instead there are functions to either spawn the process in a scope via `run`, spawn it as a daemon which is expected to be kept running via `daemon`, or run to completion via operations `join` and `expect`. The latter two do not require a scope at all. This makes use of the same kind of pattern we have used for streams, where we create an abstract thing which can either be spawned in the current context, or run to completion as an operation. The API looks like this:

``` typescript
run(function*(scope) {
  let process = exec('node some-server.js').run(scope); // run in scope
  
  console.log('started process:', process.pid);
  scope.spawn(process.stdout.forEach((m) => ...));
  yield process.join();
});

run(function*(scope) {
  let process = daemon('node some-server.js').run(scope); // run in scope, throw on early exit

  console.log('started daemon:', process.pid);
});

run(function*(scope) {
  let result = yield exec('node some-server.js').join(); // run to completion
  
  console.log('process completed', result.code, result.stdout);
});

run(function*(scope) {
  let result = yield exec('node some-server.js').expect(); // run to completion, throw on non-zero exit code
});
```

Both `join` and `expect` return `stdout` and `stderr` as strings, rather than streams. This makes it easy to run a process to completion and grab the output, which is a pretty common thing to need to do.

### DIFF
--- a/packages/node/src/daemon.ts
+++ b/packages/node/src/daemon.ts
@@ -1,6 +1,6 @@
 import { Task } from '@effection/core';
 
-import { exec, Process, ExecOptions, StdIO, ExitStatus, DaemonExitError } from './exec';
+import { exec, Process, ExecOptions, ExitStatus, DaemonExitError } from './exec';
 
 /**
  * Start a long-running process, like a web server that run perpetually.
@@ -8,14 +8,21 @@ import { exec, Process, ExecOptions, StdIO, ExitStatus, DaemonExitError } from '
  * before the operation containing them passes out of scope it raises an error.
  */
 
-export function daemon(scope: Task, command: string, options: ExecOptions = {}): StdIO {
-  let process: Process = exec(scope, command, options);
+export interface Daemon {
+  run(scope: Task): Process;
+}
 
-  scope.spawn(function*() {
-    let status: ExitStatus = yield process.join();
+export function daemon(command: string, options: ExecOptions = {}): Daemon {
+  return {
+    run(scope) {
+      let process = exec(command, options).run(scope);
 
-    throw new DaemonExitError(status, command, options);
-  });
+      scope.spawn(function*() {
+        let status: ExitStatus = yield process.join();
+        throw new DaemonExitError(status, command, options);
+      });
 
-  return process;
+      return process;
+    }
+  }
 }
--- a/packages/node/src/daemon.ts
+++ b/packages/node/src/daemon.ts
@@ -1,6 +1,6 @@
 import { Task } from '@effection/core';
 
-import { exec, Process, ExecOptions, StdIO, ExitStatus, stringifyExitStatus  } from './exec';
+import { exec, Process, ExecOptions, StdIO, ExitStatus, DaemonExitError } from './exec';
 
 /**
  * Start a long-running process, like a web server that run perpetually.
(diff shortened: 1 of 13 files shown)


packages/node/src/daemon.ts
```diff
@@ -1,25 +1,28 @@
...
-export function daemon(scope: Task, command: string, options: ExecOptions = {}): StdIO {
-  let process: Process = exec(scope, command, options);
-
-  scope.spawn(function*() {
-    let status: ExitStatus = yield process.join();
-
-    let error = new Error(`daemon process quit unexpectedly\n${stringifyExitStatus(status)}`);
+export interface Daemon {
+  run(scope: Task): Process;
+}
```
cowboyd, line 13: I see this cropping up a lot. I use it here: https://github.com/thefrontside/simulacrum/blob/v0/packages/server/src/interfaces.ts#L4-L6

Maybe we've found the resource equivalent of `v2` E.g.

```ts
export interface Resource<T> {
  create(scope: Task): T;
}
```

E.g.

```ts
export declare function daemon(cmd, opts): Resource<Process>;
export declare function exec(cmd, opts): Resource<Process>;
```

to use:

```ts
daemon('node server.js').create(scope);
let { stdout } = yield exec('echo hello world').create(scope).join();
```

packages/node/src/exec.ts
```diff
@@ -1,26 +1,63 @@
...
   let opts = { ...options, arguments: args.concat(options.arguments || []) }
 
-  if (isWin32()) {
-    return createWin32Process(scope, cmd, opts);
-  } else {
-    return createPosixProcess(scope, cmd, opts);
+  return {
+    run(scope: Task) {
+      return createProcess(scope, cmd, opts);
+    },
```
cowboyd: It feels more consistent to follow the established pattern here, even when we're delegating:

```ts
run(scope: Task) {
  return createProcess(cmd, opts).run(scope);
}
```

cowboyd (commented):
This feels much, much better. 

I'm a little worried about the potential foot-gun that the `buffered` option presents as a public API since it means that you have the potential to blow out your memory if you don't know to drain it. Do we need it to implement `join()` and `expect()`.

The other thing I noticed is that the tail is no longer in the error message. The main reason that a tail was added was to diagnose problems in a situation where you are _not_ necessarily tracking the output. This is a very common case when something goes wrong during daemon startup, or when the daemon quits unexpectedly. Knowing what happened just beforehand is tremendously helpful during debugging.

packages/node/src/daemon.ts
jnicklas: Maybe, but I'd personally like to hold off on formalizing this pattern until we have a little more experience.

I could foresee something like `task.create(daemon(...))` or `task.resource(daemon(...))` or `task.allocate(daemon(...))` maybe?

packages/node/src/exec.ts
jnicklas: I chose not to change this API because it is internal, and I wanted to keep the diff smaller, but I can totally change that if you prefer.

packages/node/src/exec/win32.ts
```diff
@@ -1,119 +1,110 @@
...
+      };
+
+      scope.spawn(function*(task) {
+        task.spawn(function*() {
+          let value: Error = yield once(childProcess, 'error');
+          getResult.resolve({ type: 'error', value });
+        });
+
+        task.spawn(on<Buffer>(childProcess.stdout, 'data').map((c) => c.toString()).forEach(stdoutChannel.send));
+        task.spawn(on<Buffer>(childProcess.stderr, 'data').map((c) => c.toString()).forEach(stderrChannel.send));
```
cowboyd, line 75: It's nice that we have bound functions now for all instances so that we can use `channel.send` as a value.

cowboyd (approved):

PR #275, opened 2021-03-22
Add value streams

jnicklas:
A value stream is the combination of a stream and a current value. Some streams make it possible to extract a value at any point in time, and this can often be useful. For example the `stringBuffer` stream introduced in #255. Rather than adding ad-hoc solutions for this, we introduce a more formal concept. Another example of a value stream could be `Atom`.

I have chosen to implement `map` for value streams, such that the mapped stream is also a value stream, however, I did not choose to do the same for `filter`. The reason being that a value stream should be well behaved such that the current value is always the last value that was emitted. This property can be guaranteed to transfer to a mapped value stream in most cases (unless it does something particularly silly like adding a random number), but it cannot be guaranteed for `filter`. Therefore to be on the conservative side, `valueStream.filter(...)` returns a `Stream`, not a `ValueStream`.


jnicklas:
Closing this, since it's very out of date, and likely not the implementation we'd choose.

PR #276, opened 2021-03-23
Website usecases

jorgelainfiesta:
## Motivation

We had a workshop in which we came up with a few use cases that represent Effection's power. Having this documentation will help us grow our internal knowledge of Effection and consolidate the APIs. 

## Approach

I'm setting up a docusaurus website in which we can put together everything related to the use cases. I want each use case to be understandable for people that encounter Effection (and structured concurrency) for the first time. That's why I'm proposing that we structure each use case with the following sections:

- Introduction: what is this use case in a nutshell?
- Common challenges: why is this use case difficult or interesting? 
- Approach: how do we address this problem using structured concurrency? try to explain it without code or API references, just natural language.
- Criteria: what are the _must_ haves for an adequate solution to this use case?
- Implementations: a solution in Effection is a must, and I think it'd be great to maintain solutions using other popular tools.

### TODOs

I fleshed out more content about type-ahead because I've implemented it in the past so I have some more context. But I don't really know enough about the other use cases. I propose that we finish the sections that go before "Implementations" for all use cases as a first step. Maybe we can each take one?

[Rendered site](https://deploy-preview-276--effection.netlify.app/effection/docs/use-cases/typeahead)


cowboyd:
@jorgelainfiesta Another case that should be here that I think everybody can understand is managing TIMEOUTS

omg how difficult they are to do without effection.

jnicklas:
I've rebased this branch on `v2`. While all the copy here is just placeholder test, having a skeleton of the website to start with is a great starting point. Let's roll with this for now, and we can improve it from here.

jnicklas (approved):

PR #281, opened 2021-03-25
Filter with a type predicate can narrow type of stream

### LINKED ISSUE #274: Ability to narrow the type of a filtered stream
If I have a discriminated union

```ts
type Person = {
  type: 'person';
  name: string;
};

type Planet {
  type: 'planet';
  orbit: number;
}

type Thing = Person | Planet;
```

We should be able to have some way to narrow the type of a stream with a filter:

```ts
let stream: Stream<Thing> = getStream();
stream.filter(isPlanet).foreEach(({ orbit }) => console.log(`planetary orbit is ${orbit}`));
```

jnicklas:
Closes #274

### DIFF
--- a/packages/subscription/src/stream.ts
+++ b/packages/subscription/src/stream.ts
@@ -6,6 +6,8 @@ import { SymbolOperationIterable } from './symbol-operation-iterable';
 import { Callback, createOperationIterator } from './create-operation-iterator';
 
 export interface Stream<T, TReturn = undefined> extends OperationIterable<T, TReturn> {
+  filter<R extends T>(predicate: (value: T) => value is R): Stream<R, TReturn>;
+  filter(predicate: (value: T) => boolean): Stream<T, TReturn>;
   filter(predicate: (value: T) => boolean): Stream<T, TReturn>;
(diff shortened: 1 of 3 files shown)


packages/subscription/src/stream.ts
```diff
@@ -28,16 +30,20 @@ export interface StringBufferStream<TReturn = undefined> extends Stream<string,
...
-    filter(predicate: (value: T) => boolean): Stream<T, TReturn> {
-      return createStream((publish) => {
-        return subscribable.forEach((value) => function*() {
-          if(predicate(value)) {
-            publish(value);
-          }
-        });
+  function filter<R extends T>(predicate: (value: T) => value is R): Stream<T, TReturn>
+  function filter(predicate: (value: T) => boolean): Stream<T, TReturn>
+  function filter(predicate: (value: T) => boolean): Stream<T, TReturn> {
```
jnicklas, line 35: Would be really nice not to have to write out this ridiculous type, but I couldn't find a way to do that.

cowboyd (approved):
Nice!

PR #287, opened 2021-03-31
Improve reentrancy on atom

jnicklas:
This makes state changes appear in the correct order if a state change is itself triggered by another state change. While this improves the status quo, there are still some pretty significant traps to fall into here. Most significantly the state received by a subscription might not actually match the current state of the atom. This is sort of a pest/cholera situation, where we have to choose in favour of correctness in some way, and neither option is particularly appealing.

There is a case to be made that this entire situation would be best avoided entirely. Maybe we should throw an error if the user tried to set the state from within a state change?

### DIFF
--- a/.changeset/calm-points-arrive.md
+++ b/.changeset/calm-points-arrive.md
@@ -1,0 +1,6 @@
+---
+"@effection/atom": patch
+"effection": patch
+---
+
+Make atom more reentrant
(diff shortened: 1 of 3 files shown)


cowboyd (approved):
If feel like we're encountering an "iron dipole" here (as opposed to an iron triangle) where exactly one of either `get` or `set` maybe synchronous, but not both.

jnicklas:
> where exactly one of either get or set maybe synchronous, but not both.

In that case it is definitely `set` that should be async, but it's worth keeping in mind that it only needs to be async if you're calling `set` from within a `set` itself. I wonder: should we even encourage such a thing? If we're talking about the sort of effect/reactive architecture that you're experimenting with, it should seem feasible to always delay the effect until the next tick. I seem to remember you having this `useEffect` hook, which would add an effect mapping, this hook could insert the tick for you.

If `set` should wait for all effects to complete, then we really need a more thorough integration between effects and the atom itself.

So I'm a bit weary of this, this PR somewhat improves the situation, but there are still a lot of footguns and odd behaviour here.

PR #288, opened 2021-03-31
Improved subscriptions and first class queues

jnicklas:
Since adding streams, we have been unsure of what to do with subscriptions. Should we double down on streams and get rid of subscriptions all together? Should we improve the subscription interface? What are the fundamental primitives.

This is an experiment in a new direction, based on some experience gained in porting code inside BigTest.

A few key insights:

- For many interfaces, using a Stream is not a good choice, since the moment of subscription is crucial and missing any messages is unacceptable. Using streams places a high burden on the consumer to subscribe (or buffer) at the right point in time. As such, subscriptions/iterators are indispensible, and relying on streams only is not tenable.
- Higher order operations (filter/map/match/reduce) are useful on streams, but not so much on subscriptions. There is some very limited utility for `map`, but I'm interested in seeing if we can get by without *any* higher order operations, since it seems odd to provide `filter` but not `map`.
- The fundamental primitive that implements Subscription is a queue. Currently, the `createOperationIterator` function uses a queue internally, and this is no accident. The ability to push in values on one end and receive them on the other is the fundamendal ability that defines a Subscription. However while such a primitive exists for `Stream` in the form of `Channel`, we currently do not have an equivalent for `Subscription`.

This PR attempts to implement these insights.

We turn `Subscription` into an interface, and subscriptions can be created via `createSubscription`. The `createOperationIterator, `Subscription.of`, `Subscription.create` as well as `subscribe()` are removed. Queue is made a first class citizen which implements `Subscription` with an expanded interface, and forms the fundamental primitive of subscriptions.

OperationIterator and OperationIterable are on the kill list but will stick around for now.


cowboyd (approved):
👍🏻

Issue #257, opened 2021-03-10
Current Task as default argument and Task elision


jnicklas:
I am resurrecting an old idea to solve an old problem in a new way. This was a thought after a discussion that @cowboyd and I had yesterday around how passing in a `Task` as an argument is not intuitive.

This was the suggested new API for fetch:

``` typescript
run(function*(task) {
  let request = createFetch(task, 'some/url');
  yield request.send();
  let data = yield request.json();
});
```

What if we could write it like this:

``` typescript
run(function*() {
  let request = createFetch('some/url');
  yield request.send();
  let data = yield request.json();
});
```

No strange `task` variable and `createFetch` is just a regular old function!

The key to making this work is that in 99% of cases, the task that we want to send in is the task of the currently running function. It's a very rare thing that this is not the case! Effection used to have an API for getting a handle to the currently running task/context, but we removed this at some point. What if we brought it back, and then the signature of `createFetch` would look like this:

``` typescript
import { Effection } from 'effection';

function createFetch(url: string, options?: FetchOptions, scope = Effection.currentTask) {
  // ...
  scope.spawn(...)
  // ...
}
```

In the case where you do need to "widen" the scope of a created resource, you can manually pass the task in.

We could even provide a `spawn` free function, so that you don't have to pass in the task just for spawning:

``` typescript
import { run, spawn } from 'effection';

run(function*() {
  spawn(function*() {
    // ... do whatever ...
  });
  // continue here
});
```

The downside to this is that in the case where you *should* widen the scope, it might not be immediately obvious how to do it. We're adding some implicit magic here which could be a bit confusing.

One cool trick we can do is that the default for `Effection.currentTask` is the `root` task, this way we can actually call a function like `createFetch` outside of an operation! Another neat thing is that it makes `run` a subset of `spawn`, so we could get rid of `run` entirely and just do this:

``` typescript
import { spawn } from 'effection';

spawn(function*() {
  spawn(function*() {
    // ... do whatever ...
  });
  // continue here
});
```

jnicklas:
I've had some thoughts about this and how it could relate to resource-like things.

To recap: a resource is a thing which needs to spawn tasks in the background. In `v1`, this was a formalized concept with the `resource` function, which did some very dark magic to make things work. In `v2` we have opted for a simpler approach, where resources are simply functions which take a task as an argument.

The proposal is to define a resource like this:

``` typescript
interface Resource<T> {
  use(scope: Task): Operation<T>;
}
```

[...]

PR #293, opened 2021-04-07
Add resource interface and make resources spawnable

jnicklas:
See https://github.com/thefrontside/effection/issues/257#issuecomment-814808843


jnicklas:
@cowboyd I created this to get a quick feel for what this would look like, take a look and see what you think!

cowboyd:
We're going for the moonshot of just being able to make Resource an operation type. 🚀

PR #294, opened 2021-04-08
Add Task `use` to formalize resources

jnicklas:
Another, more conservative approach to adding resource-like things. An alternative to #293. This time by adding `task.use` which allocates a resource and returns a pseudo-task like thing. The advantage to this is that we can cleanly halt the entire resource and everything it allocates. The interface for defining a resource is the same.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4,6 +4,7 @@ import { Effection } from './effection';
 
 export { Task, TaskOptions } from './task';
 export { Operation } from './operation';
+export { Resource, ResourceTask } from './resource';
 export { sleep } from './sleep';
 export { Effection } from './effection';
 export { deprecated } from './deprecated';
--- a/packages/core/src/resource.ts
+++ b/packages/core/src/resource.ts
@@ -1,0 +1,11 @@
+import { Task, Operation } from './index';
+
+export interface Resource<TOut> {
+  use(scope: Task): Operation<TOut>;
+}
+
+export interface ResourceTask<TOut> extends Promise<TOut> {
+  task: Task;
+  initTask: Task<TOut>;
+  halt(): Promise<void>;
+}
--- a/packages/core/src/task.ts
+++ b/packages/core/src/task.ts
@@ -10,6 +10,7 @@ import { swallowHalt } from './halt-error';
 import { EventEmitter } from 'events';
 import { StateMachine, State } from './state-machine';
 import { HaltError } from './halt-error';
+import { Resource, ResourceTask } from './resource';
 
 let COUNTER = 0;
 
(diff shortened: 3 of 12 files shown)


cowboyd (commented):
I really like the concept of a resource as a "holder of tasks". There's something that feels both clean and fundamental about the abstraction.

However, seeing this PR makes me wonder why we need a separate `ResourceTask` API to begin with? I mean I understand why we need it from an implementation perspective, but should we need it? It's a "task-like" thing in the sense that you can do with it the same things you _always_ do with a task when you spawn it.  Specifically, you either `await` it, or you `halt()` it, and in that sense it's really just an enhanced `Promise`. It seems a shame that we need a separate interface just to use it the same way we use every task.

But of course, we have to have a separate interface because there are other things that a Task does like be able to `spawn` and `ensure` and so we have to have a narrower API. But what occurs to me is that we use `Task` _completely_ differently in practice depending on if it is on the left side or the right side of a `spawn()`.

```ts
let left = scope.spawn(function*(right) { /* ... */ });
```

For the left side, the only thing I have ever done with this is either `then/catch/finally` it, or `halt()` it. Is it ever a healthy idea to say `left.spawn()`? and spawn something in the scope of your child and not in your own scope? This feels like a pretty shady gray area to draw in and still be considered as maintaining structural concurrency guarantees; like a call stack that allocates a variable in a child frame and uses it for storage. Sure the child frame probably won't see it, but why not use your own frame? That's what its for.

As for things like `left.ensure()`, they can be better expressed with `left.finally()`

Now let's look at the right side. In practice, we never call the same methods on `Task` when it occurs on the right side of a spawn point that we do on the left.

`then/catch/finally` 
--------------------
Using these methods from a child are nonsense at the best, and harmful at the worst and anything you could do with them is trivially accomplished by other means. For example, you could call `finally()` on your scope.... or you could just have a `finally` block yourself. Calling `catch()` on the other hand, what would that even?

`halt()`
--------

I could maybe see a use case for this, where, for example you wanted to bind a callback to a "Cancel" button, but even so, in order that the structural guarantees should hold, a task halting itself must require a yield point:

```ts
run(function*(self) {
  self.halt();
  console.log('nice try! I'm still going!!!')
})
```

Again, a task halting itself feels dubious to begin with, and it's no surprise that we've never actually encountered it in the wild. After all, if you do want to "halt" yourself, why not just return like a normal function would by yielding to a signal and exiting gracefully. Better yet, if you must call `halt()`, do it from the the task that spawned you, which is not problematic in the slightest.

This is all to say, that in our actual usage, and in our intuition, we have two _separate_ interfaces for using task that happen to be met in a single implementation: those that occur to the left of a spawn point, and those that occur to the right. In fact, we've already been representing this dichotomy in our coding  conventions where we refer to the left side as `task` and the right side as `scope`. What if we made this explict by splitting the current class into two interfaces: `Task` and `Scope`

```ts
interface Task<TResult> extends Promise<TResult> {
  halt(): Promise<void>;
}

interface Scope {
  spawn<TResult>(operation: Operation<TResult>): Task<TResult>;
  use<THandle>(resource: Resource<THandle>): Task<THandle>;
  ensure(callback: () => void): void;
}
```

Not only would this allow us to introduce a "Resource Task" without actually introducing any new interfaces (other than `Resource` itself), but I think it would reinforce the structural concurrency paradigm in the minds of our users.

cowboyd:
I think we can keep the same basic implementation with what is currently `Task` doing all the heavy lifting. It would just have two thin interfaces that it could present to its caller and callee, and that we could use the thin `Task` interface for both operation tasks and resource tasks

cowboyd:
Another thing that occurs to me: the basic difference of when you want to use a resource and when you want to use a simple operation is that resources are meant to be persistent. Should we warn if a resource finishes early? After all, this is something that is actually a _very_ common pitfall when using Effection; especially when you're getting started. Things like forgetting to put a `yield` at the end of your operation, or forgetting to loop in a `while`

We could detect this case and warn or even fail when it occurs, similar to how we do with `dameon()`.

cowboyd:
We're going for the moonshot of just being able to make Resource an operation type. 🚀

PR #299, opened 2021-04-14
Unify promises, iterators, and resolutions as "atomic" operations

### LINKED ISSUE #270: Consider bringing back bare iterators as an operation
One of the biggest differences between `v1` and `v2` is that generators are no longer an operation, only generator functions. This was done because unlike promises which only have a trinary state (pending, resolved, rejected) that once settled never changes, an iterator represents a more complex set of state and one that can be incoherent if multiple tasks are yielding to it at the same time. Promises don't have this problem because no matter where it is resolved, all consumers of the promise will see the same value forever and always.

While this makes the implementation simpler, it means that there are odd asymmetries in the api that can be difficult to explain. For example, to refactor the following code from an async function to a generator:

```js
return {
  async listening() {
    await bound.promise;
    return server.address();
  }
```

You would think, that you could just substitute `*` for `async` in the function declaration and then replace every instance of `await` with `yield`. 

```js
return {
  *listening() {
    yield bound.promise;
    return server.address();
  }
}
```

However, this is not the case. The equivalent generator based operation is:

```js
return {
  listening: () => function*() {
    yield bound.promise;
    return server.address();
  }
}
```

This is confusing and it doesn't allow us to leverage people's understanding of async/await as cleanly. In other words, an async function is an operation, and the return value of an async operation is also an operation, and the same is not true for generator functions.

I think that it is possible to add this in safe way if we say that only a single task may be driving an iterator at a given time so even if we are yielded the iterator, we first resolve it to a promise and then yield on that.

```js
let gen = genFunction();
yield gen; //=> sets up task and controller and attaches a promise to `gen` which resolves when it is driven to completion
yield gen; //=> produces the same value as above because it is keying off the promise associated with `gen`, not `gen` itself.
```

My gut tells me that allowing async functions and generators to be nearly swappable will lower the bar to entry into the ecosystem considerably, especially when today they are _often_ swappable, but not always. It means that we have to expend extra work explaining why this asymmetry must exist and how to think about it.

cowboyd:
## Motivation

A major part of the motivation is to see what the properties of a solution would be that resolves #270

As further motivation, I was hoping not just to achieve a closer parity with async/await code, but also try and reduce the ceremony required to do every day tasks. For example, @pittst3r and @minkimcello recently ran afoul of this while [trying to refactor the bigtest example install scripts](https://github.com/thefrontside/bigtest/pull/913/files#diff-567ca1c36c415906092e9ef8da5c5fbf84266fdc9cf53f3b1fd31fcce1315f1f). It's just cumbersome to have to always wrap your operations in the ceremony of a higher order function, especially when you're not going to do anything with the task parameter to widen the scope or spawn things.

## Approach

The fundamental approach was to see what would be required to simplify the model of what constitutes an operation into either a) a "continuable" thing or b) a function that takes a task and returns a continuable thing. That's it. Those are the only two things that an operation can be. What I found was that not only was this possible, it actually was quite easy to implement with the current architecture (taking just a few hours), and that with the exception of the very rarely used `OperationResolution` form, was completely and totally backwards compatible with all `v2` code. For example, since `simulacrum` does not use the `OperationResolution` form at all, it was a 100% seamless upgrade. 

Furthermore, I think that it makes upgrade from `v1` code much simpler since any `v1` operation that does not use resources or `v1` `Controls` is also automatically a valid `v2` operation.

The end result is that there are four atomic operation types or "yieldable" things

- Iterator
- Promise
- Resolution
- `undefined`

And then there is a function operation which just takes a task and returns one of those atomic things. Again, what's nice about this is that there is nothing special about a generator function. It's completely and totally interchangeable conceptually with an `async` function for those acquanting themselves with effection. And that's because like an `async` function that returns a promise, It's just a function that returns an atomic operation.

Not only does this result in a simpler conceptual model, but it simplifies the implementation since there is no longer the need for a `FunctionController` and all of the controller selection logic is in a single place.

**Operation Resolution**

The only breaking change here was that `OperationResolution` had to be changed from a function into an interface. I tried to do this without adding extra interface, but typescript was not happy. In the end though, I think that it results in a more ergonomic API. This is because, as with generators, you always have the option of using a single function to define your operation instead of two.

To make this work though, there was one more change. In all cases, the only thing that `task` was ever being used for was to add `ensure()` hooks, but that can be accomplished by returning your teardown from your setup which is a fairly common pattern.

before:
```ts
export function sleep(duration: number): Operation<void> {
  return (task) => {
    return (resolve) => {
      let timeoutId = setTimeout(resolve, duration);
      task.ensure(() => clearTimeout(timeoutId));
    }
}
```
after:

```ts
export function sleep(duration: number): Operation<void> {
  return {
    perform(resolve) {
      let timeoutId = setTiemout(resolve, duration);
      return () => clearTimeout(timeoutId);
    }
  } 
}
```

In the event that you want to spawn some stuff, that's fine too. You can always define an operation function that takes a task and returns a resolution as before.

**Iterators**

The changes to the resolution API represent by far and away the bulk of the proposed changes. In fact, I believe there were no changes outside of `@effection/core` that were not related to this api change. Once those were in place, the changes related to actually making iterators an operation were actually pretty minimal.  If you take a way the testcase, it works out to around 15 lines of code https://github.com/thefrontside/effection/commit/5917539833ed6accab723782b1e56a27de79c63c

The strategy is to just store the task on the iterator using a private symbol the first time it is yielded to, and treating it as a promise if it's ever seen again after that.

**Alternatives**

There are several different variations of the `OperationResolution` API we could go with aside from having the teardown be the return. Alternatively, we could pass in an `ensure` function as a third parameter:

```ts
function perform(resolve, reject, ensure) {
 /*...*/
}
```
This is nice, but is annoying if you are only using `resolve` and not `reject`

Or maybe make it a destructurable:

```ts
function perform({ resolve, reject, ensure }) {
  /*...*/
}
```

  - [x] any time there is a symbol introduced, it needs to be done very carefully, and in a way that won't bork things up with different versions of the library finding different things. **With the _fail fast_ approach, we're no longer doing anything with a symbol except maintaining a sanity check**
  - [x] There could be hidden typing gotchas that aren't accounted for, although it does compile all internal packages _and_ simulacrum. _does typecheck on core repo as well as simulacrum_
  - [x] There are potential gotchas if you are exporting atomic operations at the module level. It's hard to imagine such a scenario but it does exist. Suppose we added an API to stream all the environment variables from node everytime they changed:
  
```ts
import { ENV } from '@effection/node';

export const logENV: Operation<void> = ENV.forEach(vars => console.log(vars))
```

If `forEach` were implemented with a generator function (and returned an iterator), this would be akin to exporting a `Promise` rather than a stateless operation. And it would "bind" to the very first task that ran it. The solution would be to have have `forEach` return an `OperationFunction` (as it does now) so that a new iterator is created every time. The difference is that you are not _required_ to do this.

### TODOs and Open Questions

- [x] I didn't do anything to the `Task` when adding it to the iterator since it is a promise, but we may want to proxy it to handle edge cases like if it is halted? Although it seems actually pretty safe, which is testament to how nice `v2` really is. **With _fail fast_ we're never proxying or delegating anything, so this is not an issue**
- [x] rename `OperationResolution` something like `Action` since now it is an atomic and isn't required to appear in an `OperationFunction`. _Maybe, but not now. If it's on the kill list, we may not need it anyway._
- [x] Go even farther and make `Iterator`, not `Generator` the operation type? **nope**
- [x] How does this work in conjuction with the Task Ellision proposal #257? _It does not seem incompatible, and so what we'll do is make a change to have the `Effection.current` Task available and experiment in user space._
- [x] How does this work in conjunction with the v2 resources proposals in #293 and #294 _resources will be a fundamental operation type that can be yielded to_


packages/core/test/generator.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('can run as operations', async () => {
+    expect(await run(add(2, 2))).toEqual({ sum: 4 });
+  });
+
+  it('yields the same thing when run twice in the same operation', async() => {
+    let result = add(2, 2);
+    await run(function*() {
+      expect(yield result).toBe(yield result);
+    })
```
jnicklas: I still see this as somewhat of an anti-pattern and would kind of prefer this to throw an error, instead of using the cached promise. I feel like if `add` would have any side-effects, especially if those side effects are dependent on the context that it runs in, then this behaviour becomes a bit of a foot-gun.

cowboyd: Yeah, that makes sense. I was also thinking that we can make a linting rule that would complain if you were doing anything but yielding to an iterator or a promise.

cowboyd: @jnicklas Updated with a "fail fast" strategy

taras:
I was just reading Paul's blog post [JavaScript generators: The superior async/await](https://blog.logrocket.com/javascript-generators-the-superior-async-await/) and there is something powerful about being able to say "it's just better async await". It sounds like this proposal takes caveats out of that statement by removing the extra ceremony.

cowboyd:
I added a few more questions in the "Open Questions and todos", specifically how this would interact with other proposals

jnicklas (approved):

packages/core/src/controller/function-controller.ts
```diff
@@ -1,42 +1,25 @@
...
-  async halt() {
-    if(this.controller) {
-      this.controller.halt();
-    } else {
-      throw new Error('INTERNAL ERROR: halt called before start, this should never happen');
+  halt() {
+    if (!this.delegate) {
+      throw new Error(`EFFECTION INTERNAL ERROR halt() called before start()`);
     }
+    this.delegate.halt();
```
jnicklas, line 23: There is some churn here, which is a bit unnecessary, but it's not really important.

PR #301, opened 2021-04-16
Turn Task into an interface

jnicklas:
Using an interface allows us more flexibility. The interface is kept intentionally minimal, with a separate interface for the internal controls.

## Motivation

There are a couple of reasons to prefer interfaces over classes:

1. bound functions are nice
2. interfaces allow tighter control over the public API
3. interfaces allow different underlying implementations

Especially the last point is important. For the case of resources we would like to return something which acts mostly like a task, but which actually consists of two tasks meshed together. This will be much easier to achieve when we can fake the interface, rather than having to create a bogus class.

## Approach

We create a very minimal Task interface, which only exposes those operations which should be public, and expand the controls interface to hide all the things which are internal. Some things might move to the public interface at some point.

### Alternate Designs

We could preserve `Task` as a class (something like `TaskInternal`) and have it implement the `Task` interface, as well as the controls, or maybe have a separate class for the controls. This would allow us to 

Allocating a Task should be cheap. In theory, classes make better use of memory, since they do not require an allocation beyond the instance and its instance variables, whereas a bound object would need to allocate both every single bound function, as well as the object itself. However, assuming a sufficiently smart JIT, maybe this is fine?

### DIFF
--- a/packages/core/src/task.ts
+++ b/packages/core/src/task.ts
@@ -27,8 +27,6 @@ type WithControls<TOut> = { [CONTROLS]?: Controls<TOut> }
 export interface Task<TOut = unknown> extends Promise<TOut> {
   readonly id: number;
   readonly state: State;
-  readonly result?: TOut;
-  readonly error?: Error;
   catchHalt(): Promise<TOut | undefined>;
   spawn<R>(operation?: Operation<R>, options?: TaskOptions): Task<R>;
   ensure(fn: EnsureHandler): void;
@@ -36,8 +34,10 @@ export interface Task<TOut = unknown> extends Promise<TOut> {
 }
 
 export interface Controls<TOut = unknown> extends Trapper {
-  readonly options: TaskOptions;
-  readonly children: Set<Task>;
+  options: TaskOptions;
+  children: Set<Task>;
+  result?: TOut;
+  error?: Error;
   start(): void;
   halted(): void;
   resolve(value: TOut): void;
@@ -63,8 +63,6 @@ export function createTask<TOut = unknown>(operation: Operation<TOut>, options: 
   let trappers = new Set<Trapper>();
   let ensureHandlers = new Set<EnsureHandler>();
   let emitter = new EventEmitter();
-  let _result: TOut | undefined;
-  let _error: Error | undefined;
 
   let stateMachine = new StateMachine(emitter);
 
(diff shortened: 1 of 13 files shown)


packages/core/src/task.ts
```diff
@@ -26,180 +22,221 @@ export interface TaskOptions {
...
 
-export class Task<TOut = unknown> extends EventEmitter implements Promise<TOut>, Trapper {
-  public id = ++COUNTER;
+type WithControls<TOut> = { [CONTROLS]?: Controls<TOut> }
+
+export interface Task<TOut = unknown> extends Promise<TOut> {
+  readonly id: number;
+  readonly state: State;
+  readonly result?: TOut;
+  readonly error?: Error;
```
cowboyd: I'm wondering what the use case is for making these part of the public API? Shouldn't these values be consumed via `then()` and `catch()`? There's no way to consume them synchronously, so if I were coming to effection with fresh eyese, I'd be confused as to what they represent.

packages/core/src/task.ts
```diff
@@ -26,180 +22,221 @@ export interface TaskOptions {
...
+  catchHalt(): Promise<TOut | undefined>;
```
cowboyd, line 30: I've never needed to use this in public code, but feels like if we are going to make it public, then the return type should be`Promise<MaybeHalted<T>>` because if this is a `Task<undefined>` then there is no way to distinguish between the halt and resolve case. E.g.

```ts
type MaybeHalted<T> = { didHalt: true, value: T } | { didHalt: false, value: undefined };

interface Task<T> {
  /*..../*
  maybe: Promise<MaybeHalted<T>>;
```

Does this need to be part of the public API for now?

jnicklas: I personally find this method really useful. The point of it isn't so much to provide a useful return value, and more for halts to not blow anything up. Since `HaltError` is not exported, there isn't any other really safe way of preventing this.

packages/core/src/task.ts
jnicklas: I've moved these to the private `controls` interface.

packages/core/src/task.ts
cowboyd: I can definitely see why it's useful, especially when you're consuming a task from outside effection. But what about the case though where you have a `Task<undefined>`. Is it useful to disambiguate between `undefined` because successful completion, or `undefined` because of halt?

cowboyd (approved):

PR #302, opened 2021-04-16
✨ Allow for pending `it()` blocks

cowboyd:
Motivation
-----------

Mocha allows for `it()` blocks that do not yet have a body. This is so that you can write out the flow of a testcase before you actually get around to making it work. To support this, the second argument to `mocha.it()` is actually optional.

```js
describe('a thing', () => {
  it('will do this');
  it('will do that');
})
```

Approach
----------
This allows the same construction in `@effection/mocha` by relaxing the requirement for the second argument to `it()`

### DIFF
--- a/.changeset/pending-mocha-it-blocks.md
+++ b/.changeset/pending-mocha-it-blocks.md
@@ -1,0 +1,5 @@
+---
+"@effection/mocha": patch
+---
+Allow "pending" it blocks that don't yet have a body. E.g.
+`it('willb do this')`;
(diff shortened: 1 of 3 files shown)


.changeset/pending-mocha-it-blocks.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effection/mocha": patch
+---
+Allow "pending" it blocks that don't yet have a body. E.g.
+`it('willb do this')`;
```
cowboyd:
```suggestion
`it('will do this')`;
```

jnicklas (approved):

Issue #292, opened 2021-04-07
Add operation combinators


jnicklas:
There are a few combinators that we could add that come up sometimes which are useful. We should provide all methods available on Promise for intuitive interop:

- [x] `all`: completes when all the given operations complete, returns an array of results
- [ ] `allSettled`: like `all`, except it will not short circuit if any operation fails, and instead return an aggregate result
- [x] `race`: completes when any of the given operations complete, returns its result and halts all the other operations
- [x] `any`: like `race`, but some of the given may fail as long as at least one of them completes

Additionally, we might consider the following:

- [x] `timeout`: throws an error after the given amount of time (alt names: `throwAfter`, `failAfter`, etc...)
- [x] `timebox`: runs the given operation and throws an error if it has not completed after given amount of time (alt names: `withTimeout`, `withTimeLimit`, `withFailAfter`, etc...), this is essentially `race(op, timeout(x))`
- [x] `ensure`: runs the given operation on halt, basically `try { yield } finally { yield op }`, we could replace `task.ensure` with this, I think.

Anything else?

cowboyd:
I've often wanted a general operation mapping function.

Also something to consider:

a form of `all` except instead of an array, it is a hash of named promises.

```ts
let { services, url } = all({
  services: getServices(),
  url: bindPort()
});
```

I used this quite a bit and preferred it to the array form of all in `RSVP` https://github.com/tildeio/rsvp.js#hash-of-promises

They actually called the method `hash()` but I think that's confusing since a hashing function is already a well established concept

cowboyd:
@jnicklas another useful one might be a `capture()` combinator for when you want to programatically deal with error cases:

```ts
declare function capture<T, TErr = Error>(operation: Operation<T>): Operation<T | TErr>;
```

jnicklas:
@cowboyd `captureError` is something we already have in the `@effection/mocha` package, and for writing tests it is extremely useful. I would be very open to moving this to core.

jnicklas:
We're still missing `allSettled`, but I question the utility of that method. Let's close this out for now.

PR #304, opened 2021-04-21
✨Add a race() combinator

cowboyd:
Motivation
----------

Called for in https://github.com/thefrontside/effection/issues/292

Approach
----------

This adds a `race()` combinator that runs an array of operations in parallel where the first one to complete becomes the result of the race operation.

If any of the operations fail, then the race operation fails.

The API is designed to mirror the functionality of [Promise.race][1] as closely as possible

### Alternate Designs

The `race()` function as implemented is simple in its implementation, but it _will_ spawn all operations _always_. However, if the operation is a resolution function, then there is the possibility that it operation _could_ complete or fail synchronously, and so we might not need even spawn the remaining operations because we have our answer. However, by wrapping each operation in a generator function with `try/catch`, we have eliminated the possibility of detecting this case and short-circuiting. In practice, it's hard to imagine this being an issue unless you are racing thousands and thousands of operations against each other, but I did want to bring it up.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race

### DIFF
--- a/.changeset/race-combinator.md
+++ b/.changeset/race-combinator.md
@@ -1,0 +1,4 @@
+---
+"@effection/core": minor
+---
+add race() combinator
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -9,6 +9,7 @@ export { sleep } from './sleep';
 export { Effection } from './effection';
 export { deprecated } from './deprecated';
 export { Deferred } from './deferred';
+export { race } from './race';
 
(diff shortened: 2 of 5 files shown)


jnicklas:
I did some experiments with this branch, and like I mentioned when we discussed this before, there are some problems with this implementation when the operations resolve synchronously. We need to somehow insert a yield point, so that iteration of the operations does not proceed if the task has already failed. Actually implementing this correctly is surprisingly tricky though.

cowboyd:
Yeah, let's leave it for now.

PR #309, opened 2021-04-30
Add resources 👊

jnicklas:
## Motivation

Oftentimes we want to have an operation both continue to keep running, but also be able to interact with it in some fashion. For example, a process might want to be kept running, while we are able to interact with its IO-streams, or send signals to it. There is currently no good solution for this. In v2 we opted for a simple solution whereby we did not add any framework support for this pattern. However, we discovered that we keep using the same pattern over and over, and by adding some framework support for it, we can greatly simplify it and remove a lot of boilerplate.  

## Approach

There is a new interface called `Resource`, implementing a resource looks like this:

``` typescript
let myResource: Resource<string> {
  *init(scope) {
    scope.spawn(function*() { ... }) // keeps running when generator function finishes
    yield sleep(10);
    return "result";
  }
}
```

Resource can be consumed like this:

``` typescript
main(function*(task) {
  yield myResource;
  yield task.spawn(myResource); // alternatively
});
```

### Alternate Designs

We have explored many, many alternate designs to this approach.

In this PR I chose to not create a wrapping task as the scope for the resource. This way tasks spawned by the resource are direct children of its parent. I felt that this wrapping task while *neat* is also simply not necessary, and might get in the way of future ideas, such as a `spawn` resource.

Additionally, this PR has a trick up its sleeve compared to versions we've discussed earlier: yielding to a resource in a resource actually works! This allows one to write resources without a deeper understanding of the scope of things, at the cost of being a bit magical.

It adds complexity, both from an implementation perspective and from a cognitive perspective.

- [ ] Is the magical yield thing to magical?

### DIFF
--- a/.changeset/wicked-melons-bake.md
+++ b/.changeset/wicked-melons-bake.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Add Resource to create an option for spawning tasks in the background in an operation
--- a/packages/core/src/controller/controller.ts
+++ b/packages/core/src/controller/controller.ts
@@ -1,14 +1,15 @@
 import type { Task } from '../task';
 import type { Operation } from '../operation';
-import { isResolution, isPromise, isGenerator } from '../predicates';
+import { isResource, isResolution, isPromise, isGenerator } from '../predicates';
 import { createFunctionController } from './function-controller';
 import { createSuspendController } from './suspend-controller';
 import { createPromiseController } from './promise-controller';
 import { createIteratorController } from './iterator-controller';
 import { createResolutionController } from './resolution-controller';
+import { createResourceController } from './resource-controller';
 
 export interface Controller<TOut> {
-  start(task: Task<TOut>): void;
+  start(): void;
   halt(): void;
 }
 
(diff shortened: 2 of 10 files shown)


taras:
>  Is the magical yield thing too magical?

Why do you think it's magical? Isn't it just a regular generator?

jnicklas:
> Why do you think it's magical? Isn't it just a regular generator?

@taras it's somewhat magical in that a resource that is yielded to in another resource automatically has its lifetime extended past the end of the generator funcion it is in. This is probably what you want to happen, but it is nevertheless a bit surprising, since it breaks the normal rules of effection.

taras:
Does the resource’s life time extend to match callsite or callsite extends to match resource?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jnicklas:
@taras 

Maybe this will clarify:

With no magic, using another resource, such as `exec` from within a resource, needs to be written like this:

``` typescript
let myResource: Resource<string> {
  *init(scope) {
    yield scope.spawn(exec("yarn start"));
  }
}
```

With Magic™ it can be written like this:

``` typescript
let myResource: Resource<string> {
  *init() {
    yield exec("yarn start");
  }
}
```

The key here is that if we were to write it like in the second example, normally we would expect the resource to be shut down at the end of the generator function, and therefore the `yarn start` command would be terminated. But instead, we do some special magic which makes the resource of the `exec` have the same lifetime as `myResource`.

This is not normally how effection works, and it is a special rule that only works within the definition of resources. I think this behaviour does the thing that the user probably wants to do, and therefore, writing resources requires less deep knowledge of effection, but on the other hand, once the user *does* understand the rules, it might seem strange why this particular case doesn't follow them.

cowboyd (approved):

PR #310, opened 2021-05-03
Use resources in subpackages

jnicklas:
Depends on #309

Now that we have resources, let's use them where appropriate!

### DIFF
--- a/packages/fetch/src/fetch.ts
+++ b/packages/fetch/src/fetch.ts
@@ -1,9 +1,17 @@
-import { Task, Operation } from '@effection/core';
+import { Task, Operation, Resource } from '@effection/core';
 import { fetch as nativeFetch } from 'cross-fetch';
 import { AbortController } from 'abort-controller';
 
-export function fetch(scope: Task, info: RequestInfo, init: RequestInit = {}): Operation<Response> {
-  return function*() {
+export interface Fetch extends Resource<Response> {
+  arrayBuffer(): Operation<ArrayBuffer>,
+  blob(): Operation<Blob>,
+  formData(): Operation<FormData>,
+  json(): Operation<unknown>,
+  text(): Operation<string>,
+}
+
+export function fetch(info: RequestInfo, requestInit: RequestInit = {}): Fetch {
+  let init = function*(scope: Task) {
     let controller = new AbortController();
 
     scope.spawn(function*() {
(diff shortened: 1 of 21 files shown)


packages/fetch/src/fetch.ts
```diff
@@ -1,9 +1,17 @@
...
+export interface Fetch extends Resource<Response> {
+  arrayBuffer(): Operation<ArrayBuffer>;
+  blob(): Operation<Blob>;
+  formData(): Operation<FormData>;
+  json(): Operation<unknown>;
+  text(): Operation<string>;
+}
+
+export function fetch(info: RequestInfo, requestInit: RequestInit = {}): Fetch {
+  let init = function*(scope: Task) {
```
cowboyd:
```suggestion
  function* init(scope: Task) {
```

packages/fetch/src/fetch.ts
```diff
@@ -14,8 +22,33 @@ export function fetch(scope: Task, info: RequestInfo, init: RequestInit = {}): O
...
+      return yield response.formData();
+    },
+    json: function*() {
+      let response = yield { init };
+      return yield response.json();
+    },
+    text: function*() {
+      let response = yield { init };
+      return yield response.text();
+    },
```
cowboyd: For these operations, would it make sense to curry so they are stateless? E.g.

```js
{
  json: () => function*() {}
}
```

That way you can consume it as a value:

```ts
export const status =  fetch('api.github.com/v2/status').json()
```

Otherwise, I'd just write them using method syntax:

```ts
{
  *json() {
     let response = yield { init };
     return yield response.json();
  }
}
```

packages/mocha/src/index.ts
```diff
@@ -26,7 +26,7 @@ mocha.afterEach(async function() {
 
 function runInWorld(fn: TestFunction) {
   return async function(this: mocha.Context) {
-    await run((task) => fn.call(this, world!, task));
+    await run({ init: fn.bind(this) }, { resourceScope: world! });
```
cowboyd, line 29: Why not just spawn this as a resource in world?

packages/subscription/src/stream.ts
```diff
@@ -56,6 +56,10 @@ export function createStream<T, TReturn = undefined>(callback: Callback<T, TRetu
 
     filter,
 
+    init: function*(task: Task) {
+      return subscribe(task);
+    },
```
cowboyd: Well that's kinda cool!
```suggestion
    *init(task: Task) {
      return subscribe(task);
    },
```

packages/fetch/src/fetch.ts
jnicklas:
``` typescript
export const status =  fetch('api.github.com/v2/status').json()
```

This is not a pattern I'm a fan of encouraging, given that it depends on the implementation of `json` whether it'll work. Generally we should not be encouraging creating an operation which is shared among multiple spawn/yield points, even in those cases where that technically works.

packages/mocha/src/index.ts
jnicklas: I actually investigated this, it makes the error messages read a little bit worse, which is why I didn't do this at the moment. There might be ways around this, but this was the easier implementation.

cowboyd (approved):
🥳 fetch is now quite fetching.

PR #312, opened 2021-05-04
Add foundational operations

jnicklas:
## Motivation

There are a few operations that fill the criteria of being simple to implement, and generally useful. It makes sense to ship these as part of the effection package.

See #292 

## Approach

We include these operations in `@effection/core` in a subfolder.

The operations added are:

- ensure: runs the given operation on shutdown
- timeout: throws an error after the given amount of time
- withTimeout: runs an operation with a timeout, so it effectively races the timeout.
- spawn: simply spawn a task in the parent process, just like in effection v1
- race: race multiple operations against each other
- all: wait for all the given operations to complete

### Alternate Designs

Put these into a separate package, like `@effection/operations` or something?

Put them into individual packages, `@effection/ensure` etc... seems excessive.

None


cowboyd:
How would we consume timeout bounded code programmatically?

```js
for (i = 0; i < maxAttempts; i++) {
  try {
    return yield withTimeout(op, 2000)
  } catch (error) {
    if (error.name === 'TimeoutError') {
      continue;
    }
 }
}
```

vs

```js
for (let i = 0; i < maxAttempts; i++) {
  let result = yield withTimeout(op, 2000);
  if (result.didComplete) {
    return result.value;
  }
}
```

jnicklas:
@cowboyd my thinking was that these operations are most useful in the case where you just want to wrap an operation in a timeout. I think writing something like a retry operation with backoff would be really interesting, but that's not the usecase we should be optimizing for. Most of the time, wrapping the returned value would be inconvenient, so I'd personally go for the first type of implementation, or even not use the timeout operations at all for implementing this type of thing.

jnicklas:
@cowboyd I've rebased this PR and also added a spawn operation. Whether we want to encourage or not, it will make it much easier to port bigtest to v2.

packages/core/src/operations/ensure.ts
```diff
@@ -0,0 +1,9 @@
+import { Operation } from '../operation';
+
+export function *ensure<T>(operation?: Operation<T>): Operation<void> {
```
cowboyd: Should we allow undef here? Seems like a guaranteed hang?

cowboyd (commented):
Part of me thinks that `ensure()` should be a resource. Most of the times I've felt the need for an ensure are for when you have some synchronous, effectful code that is not part of an operation, but that you need to nevertheless add teardown for like this:

https://github.com/thefrontside/simulacrum/blob/49e2363bf0ab6da0bc89f70f9a2d877a62711b7e/packages/server/src/http.ts#L28-L34

I could write this with an ensure operation, but it would be like:

```js
let server = app.listen(options.port);

yield spawn(ensure(function*() {
  server.close();
});
```

But even better would be:

```js
let server = app.listen(options.port);

yield ensure(() => server.close());
```

This would imply `ensure()` being both a resource _and_ being able to accept synchronous `() => void` functions similar to what happens in [`forEach`](https://github.com/thefrontside/effection/blob/v2/packages/subscription/src/stream.ts#L83)

packages/core/src/operations/ensure.ts
jnicklas: True. Seems a bit useless to allow it.

jnicklas:
> Part of me thinks that `ensure()` should be a resource.

I like it! Let's go for it :)

packages/core/src/operations/spawn.ts
```diff
@@ -0,0 +1,10 @@
+import { Operation, Resource } from '../operation';
+import { Task } from '../task';
+
+export function spawn<T>(operation?: Operation<T>): Resource<Task<T>> {
+  return {
+    *init(scope) {
+      return scope.spawn(operation);
+    }
+  }
+}
```
cowboyd: What about adding a `within()` function to shift the target of the spawn? 

```ts
interface Spawn<T> extends Resource<Task<T>> {
  within(scope: Task): Resource<Task<T>>;
}
export function spawn<T>(operation?: Operation<T>): Resource<Task<T>> {
  function* init(scope) {
    return scope.spawn(operation);
  }

  function within(scope: Task) {
    return {
      init: () => init(scope)
    }
  }

  return { init, within };
}

```

This gives us a way to re-target spawns and yet still achieving the benefits of an operation.

cowboyd (dismissed):

packages/core/src/operations/spawn.ts
cowboyd: For example, this lets me write our file watcher example in the mini-guide as:

```ts
main(function*(scope) {
  let watcher = watch('./src/**/*.ts', { ignoreInitial: true, ignored: 'dist' });
  try {
    let process = yield spawn(start());

    yield on(watcher, 'all').forEach(function*() {
      yield process.halt();
      process = yield spawn(start()).within(scope);
    });
  } finally {
    watcher.close();
  }
```

cowboyd (approved):
These are some pretty nice upgrades 👍🏻 

I had an idea for another operation, but we don't want to clog up what is already a pretty big PR.

something that would just let us call a synchronous function for its side-effects within the context of an operation:

```js
function call<T>(fn: () => T): Operation<T> {
  return { 
    perform(resolve, reject) {
      try {
        resolve(fn());
      } catch (error) {
        reject(error);
      }
    }
  }
}
```

This would give flexibility about how to structure an api, and also a very fast out in the event of a race condition.

```js
yield call(() => socket.send(message));
```

PR #323, opened 2021-05-17
Sleeping for zero ms should not suspend indefinitely

jnicklas:
The correct behaviour is to sleep just until the next event loop tick. This is a bug introduced by #292.


cowboyd (commented):
Good 'ol JavaScript. Always good for a laugh.

cowboyd (approved):

PR #327, opened 2021-05-24
Add express package from @bigtest

jnicklas:
This extracts the `effection-express` package from bigtest and adds it here under `@effection/express`. The main purpose for this is the websocket code, which I want to use for the debug package. Maybe the WS code doesn't need to use `express` though, and we don't need all this other express stuff?

@cowboyd you did some work in this area for simulacrum, right?

IMO the express wrapper is somewhat questionable, for a few reasons, which is why we have so far resisted putting it in effection itself. Things I don't love about it:

- Questionable error handling. Should an error in a request really kill the whole server?!
- It doesn't properly wrap the request/response objects for requests, yet does so for websocket requests. So it operates at multiple levels of abstraction.

Anyway, putting this here as a starting point for further discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
       "packages/atom",
       "packages/node",
       "packages/fetch",
+      "packages/express",
       "packages/effection"
     ]
   },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -4,6 +4,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "license": "MIT",
+  "author": "Frontside Engineering <engineering@frontside.com>",
   "scripts": {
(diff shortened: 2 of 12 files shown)


cowboyd:
@jnicklas This was the only integration that I needed. Basically, you just stream connections and that's it. You only need to use `createServer` from node's `http` package.

https://github.com/thefrontside/simulacrum/blob/v0/packages/server/src/websocket-transport.ts#L42-L56


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/simulacrum">thefrontside/simulacrum</a></strong></div><div>A simulation platform for use during testing, during development and for high-fidelity application previews - thefrontside/simulacrum</div></blockquote>

jnicklas:
Going with #337 instead of this

PR #333, opened 2021-05-26
Add websocket client package

jnicklas:
This is a websocket client using the `isomorphic-ws` package, which works on both node and the browser (in theory). So it should provide an implementation which works in either environment.

### DIFF
--- a/packages/websocket-client/README.md
+++ b/packages/websocket-client/README.md
@@ -1,0 +1,23 @@
+# @effection/websocket-client
+
+A basic websocket client for sending and receiving messages over a websocket
+connection.  Works both in the browser and in node. The client is oppinionated
+in that it assumes that the messages are serialized as JSON.
+
+## Usage
+
+``` typescript
+import { createWebSocketClient, WebSocketClient } from '@effection/websocket-client';
+import { main } from '@effection/node';
+
+type Request = { value: string };
+type Response = { value: number };
+
+main(function* () {
+  let client: WebSocketClient<Request, Response> = yield createWebSocketClient('ws://localhost:1234');
+
+  yield client.forEach(({ value }) {
+    client.send({ value: parseInt(value) });
+  });
+});
+```
--- a/packages/websocket-client/src/index.ts
+++ b/packages/websocket-client/src/index.ts
@@ -1,0 +1,43 @@
+import * as WebSocket from 'isomorphic-ws';
+
+import { spawn, ensure, Resource, Operation } from '@effection/core'
+import { createQueue, Subscription } from '@effection/subscription';
+import { on, once } from '@effection/events';
+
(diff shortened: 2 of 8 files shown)


packages/websocket-client/CHANGELOG.md
```diff
@@ -0,0 +1,274 @@
+# Changelog
```
cowboyd: We should probably delete this file so that the changesets can generate it

packages/websocket-client/README.md
```diff
@@ -0,0 +1,59 @@
+# @effection/events
```
cowboyd: Readme should be stripped down to remove the copy-paste.

packages/websocket-client/src/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      return {
+        ...queue.subscription,
+        send(message: TOutgoing) {
+          socket.send(JSON.stringify(message));
+        }
+      }
+    }
+  }
+}
```
cowboyd: If we create a single top level "resource task", then it will work well with our visual debugger in the future, right?

```ts
export function createWebSocketClient<TIncoming = unknown, TOutgoing = TIncoming>(url: string): Resource<WebsocketClient<TIncoming, TOutgoing>> {
   return {
     *init() {
       let socket = new WebSocket(url);
  
       yield spawn(function*() {
         // not yet, but soon!
         // yield setLabels({ name: `websocket client`, url })
         try {
           let queue = createQueue<TIncoming>();
           yield spawn(function*() {
             let { wasClean, code, reason } = yield once(socket, 'close');
             if(wasClean) {
               queue.close();
             } else {
               throw new Error(`websocket server closed connection unexpectedly: [${code}] ${reason}`);
             }
           });
  
           yield on<{ data: string }>(socket, 'message').forEach((message) => { queue.send(JSON.parse(message.data)) });
         } finally {
           socket.close();
         }
       });

       yield once(socket, 'open');

       return {
         ...queue.subscription,
         send(message: TOutgoing) {
           socket.send(JSON.stringify(message));
         }
       }
     }
   }
 }
```

packages/websocket-client/src/index.ts
```diff
@@ -0,0 +1,39 @@
+import * as WebSocket from 'isomorphic-ws';
+
+import { spawn, ensure, Resource } from '@effection/core'
+import { createQueue, Subscription } from '@effection/subscription';
+import { on, once } from '@effection/events';
+
+interface WebsocketClient<TIncoming = unknown, TOutgoing = TIncoming> extends Subscription<TIncoming> {
+  send(message: TOutgoing): void;
```
cowboyd: Are there cases where having `send()`  be synchronous can cause the execution of illegal instructions?

jnicklas: Not sure what you mean by this?

cowboyd (approved):

PR #338, opened 2021-05-27
Bump @changesets/cli to 2.14.1

minkimcello:
## Motivation

[This](https://github.com/thefrontside/effection/runs/2685563148?check_suite_focus=true) is failing but [this](https://github.com/thefrontside/effection/actions/runs/882463135) passed. We noticed `v2` has `@changesets/cli@2.14.1` while `v1` is on `2.11.1` so we suspect that might be the reason why it's failing:
```
🦋  error TypeError: git.getCommitsThatAddFiles is not a function
```

## Approach

- Update `@changets/cli` version in monorepo devDependency to match what we have for `v2`
- No changeset file necessary

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,6 @@
     "patch-package": "6.2.2"
   },
   "dependencies": {
-    "@changesets/cli": "2.11.1"
+    "@changesets/cli": "^2.14.1"
   }
 }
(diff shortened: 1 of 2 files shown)


jbolda:
Are you able to run it locally and see that it makes a difference?

minkimcello:
> Are you able to run it locally and see that it makes a difference?

I could create a workflow that runs the same script to see if we get that error?

Or try running the changeset workflow on this branch because there's no harm in having it create a PR we can just close right away

jnicklas (approved):

PR #339, opened 2021-05-27
Add @effection/main package

jnicklas:
This package has separate entry points for node and the browser, and will work fine within both environments. This means we can reexport this from the `effection` package, which means we can finally do this:

``` typescript
import { main } from 'effection';
```

Which removes a really ugly wart from effection. It will also be useful to have this be part of the meta package, since we will want to use this in for example the CLI.


packages/main/src/browser.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+        if(e.options.verbose) {
+          console.error(e.stack);
+        }
+      } else {
+        console.error(e);
+        throw e;
+      }
+    } finally {
+      window.removeEventListener('beforeunload', interrupt);
```
cowboyd:
```suggestion
      window.removeEventListener('unload', interrupt);
```

cowboyd (dismissed):
Looks good, just one event name I saw that needs to be renamed.

cowboyd (approved):

cowboyd (approved):

PR #347, opened 2021-05-30
Improve Error output with Effection Trace and Colors

jnicklas:
## Motivation

We would really like to have more readable and useful error output. When things go wrong in Effection it can be really hard to figure out what's going on. Stack traces are decidedly less useful with Effection.

## Approach

When an error is thrown in an Effection context, we wrap this error in an EffectionError. When errors percolate upwards, we continually expand the trace property of the EffectionError, so it keeps track of the stack of tasks that it percolated through. This way we get a sort of stack trace of operations. With the addion of #330 and #331, we can actually print some useful information here.

We use this wrapped error in @effection/main to print a nice colorized stacktrace and Effection trace. By putting this in @effection/main, we avoid pulling in terminal colorization packages in the browser, where it wouldn't make sense. In this PR  we don't yet to anything to print any diagnostic infomation in the browser, but node environments seem like a higher priority and easier to iterate on.

### Alternate Designs

We could set a property on the error, instead of wrapping them, this would mean you would get the original exception and wouldn't have to unwrap it.

Wrapping the error in an EffectionError, could be a little confusing, since when using `catch` on a task, you wouldn't get the original error.

## Screenshots

Before:

Screenshot 2021-05-28 at 17 31 01

After:

Screenshot 2021-05-28 at 17 32 04

### DIFF
--- a/.changeset/seven-ducks-attend.md
+++ b/.changeset/seven-ducks-attend.md
@@ -3,4 +3,4 @@
 "effection": minor
 ---
 
-Add EffectionError and wrap errors in this error automatically
+Collect trace of effection operations and propagate them along with the raised error
--- a/.changeset/seven-ducks-attend.md
+++ b/.changeset/seven-ducks-attend.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Add EffectionError and wrap errors in this error automatically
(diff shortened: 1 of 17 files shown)


.changeset/seven-ducks-attend.md
```diff
@@ -0,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Add EffectionError and wrap errors in this error automatically
```
cowboyd:
```suggestion
Collect trace of effection operations and propagate them along with the raised error
```

cowboyd (approved):

PR #348, opened 2021-06-01
Add @effection/react package

jnicklas:
This currently only contains a hook called `useSubscription`, making it possible to subscribe to a subscription from within react.

## Motivation

React is by far the most popular client-side framework. Integrating with React would be great.

## Approach

It seems that hooks are the standard way of writing React apps these days, so this PR adds a hook. The most natural effection primitive to map to React is a subscription, so that's where we start out. This has come up as a usecase in building the debug UI.

### Alternate Designs

One question is how we obtain an Effection task to spawn off of, there are a few options for this:

1. Explicitly pass it in and down as a prop
2. Use `Effection.root`
3. Set it as a React context and have it be passed down implicitly

This PR chooses the third approach, since it seems like a nice combination of low ceremony and flexibility.

This is not a full-fledged react integration and it might be a bit of a red-herring for someone looking to integrate Effection and React. On the other hand we do need to start somewhere.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -1,0 +1,1 @@
+# @effection/debug-ui
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -1,1 +1,1 @@
-# @effection/debug-ui
+# @effection/react
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,0 +1,44 @@
+{
+  "name": "@effection/react",
+  "version": "2.0.0-preview.1",
+  "description": "Hooks for integrating effection into react applications",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "module": "dist/debug.esm.js",
+  "repository": "https://github.com/thefrontside/bigtest.git",
+  "author": "Frontside Engineering <engineering@frontside.com>",
+  "license": "MIT",
+  "files": [
+    "README.md",
+    "CHANGELOG.md",
+    "dist/**/*",
+    "src/**/*"
+  ],
+  "scripts": {
+    "lint": "eslint '{src,tests}/**/*.ts'",
+    "test": "mocha -r ts-node/register 'test/**/*.test.{ts,tsx}'",
+    "prepack": "tsdx build --tsconfig tsconfig.dist.json",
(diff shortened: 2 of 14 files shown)


packages/react/src/use-subscription.ts
```diff
@@ -0,0 +1,15 @@
+import { useState, useEffect, useContext } from 'react';
+import { Subscription } from '@effection/subscription';
+import { EffectionContext } from './context';
+
+export function useSubscription<T>(subscription: Subscription<T>): T | undefined {
+  let scope = useContext(EffectionContext);
+  let [state, setState] = useState<T>();
+
+  useEffect(() => {
```
dagda1, line 9: apologies for the driveby comment but a `useEffect` without a dependency array will run on every render.

There is an eslint plugin that is pretty good for what is missing in the weird world of react hooks and it is so easy to screw up it really is worth having them like this

```json
  "rules": {
    // ...
    "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks
    "react-hooks/exhaustive-deps": "error" // Checks effect dependencies
  }
  ```

Comment (jnicklas) [reply packages/react/src/use-subscription.ts:9]: Great catch! I totally forgot about this.

Comment (jnicklas) [reply packages/react/src/use-subscription.ts:9]: I've added a fix and added the eslint rules

File: packages/react/README.md [hunk 2]
```diff
@@ -0,0 +1 @@
+# @effection/debug-ui
```
cowboyd:
```suggestion
# @effection/react
```

packages/react/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@effection/react",
+  "version": "2.0.0-preview.1",
+  "description": "Hooks for integrating effection into react applications",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "module": "dist/debug.esm.js",
+  "repository": "https://github.com/thefrontside/bigtest.git",
```
cowboyd:
```suggestion
  "repository": "https://github.com/thefrontside/effection.git",
```

cowboyd (approved):

PR #350, opened 2021-06-04
Refactor internals to use futures and run loop

jnicklas:
There are a lot of problems which occur with re-entrancy in Effection. Additionally, the code around task shutdown is hard to understand and the relationship between the controller and the task is unclear. This change attempts to clean up some of these internals by introducing a `Continuation` a sort of synchronously resolving promise. Additionally we introduce a run loop, a sort of synchronous mutex. This is used internally by the continuation, as well as by other components directly.

### DIFF
--- a/.changeset/cool-brooms-bathe.md
+++ b/.changeset/cool-brooms-bathe.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Rewrite internals to using futures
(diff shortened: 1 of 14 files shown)


jnicklas:
Interestingly this also provides a path forward towards getting rid of `Controls` which would be nice, since it's a very confusing part of the codebase. The controller no longer needs it, and with the slimmed down interface left like this:

``` typescript
export interface Controls<TOut = unknown> {
  continuation: Continuation<TOut>;
  children: Set<Task>;
  start(): void;
  setLabels(labels: Labels): void;
  on: EventEmitter['on'];
  off: EventEmitter['off'];
}
```

We can get rid of `start` entirely if we want, since `Continuation` lessens the impact of race conditions with linking and adding trappers and such. If we want to keep it, we can make it public.

`children` can be made a readonly accessor which returns an array. This makes it impossible to modify the set of children from the outside, but still makes the list readable.

All other methods could just as well be public, and in fact I think would be better as public.

cowboyd (approved):

PR #352, opened 2021-06-06
Add visual inspector

jnicklas:
## Motivation

We want to gain more insight into Effection processes. Having a visual representation of what is going on inside of an Effection tree would be huge in helping people understand what is going on inside of their Effection code.

## Approach

We implement a websocket server which is embedded in a node process and that a debug UI can connect to.

### Alternate Designs

We had a thought about building a separate CLI and have the debugged process connect to this CLI process. But this turned out much more complicated in practice.

With this design, a web app will need to be debugged through devtools. This PR does not include any devtools yet, but #343 provides a path forward for adding these, so that problem is already pretty well understood.

### TODOs and Open Questions

- [x] ~~Should we dogfood and use BigTest to test the UI? It would be cool, but also potentially a bit risky in terms of bootstrapping~~ **no for now, to avoid complexity, the UI is very simple, and testing it is not essential yet**
- [x] ~~The communication protocol between the server and UI is *very* ineffcient, serializing the entire tree for every single change of state. It would be better to send only any changes to the client, so we don't have to serialize/unserialize all the time.~~ **We punt on this for now**
- [x] Should this package be part of the `effection` meta package? Should it even be imported by default? In that case how do we feel about its dependencies, such as the `@effection/websocket-server` and `@effection/atom` packages. **no for now**
- [x] Shutting down the debugger needs to clean up its resources properly.
- [x] Add more tests
- [ ] Should we randomize the debugger port? Technically this works fine, but a fixed port has the advantage of being "stable" so a simple refresh in a browser window reconnects. On the other hand, collisions with other processes are a problem. A random port avoids this problem.

## Screenshots

Screenshot 2021-06-05 at 14 59 52


jnicklas:
This is finally ready to go, but unfortunately it is failing on windows:

```
error @parcel/source-map@2.0.0-rc.1.0: The engine "node" is incompatible with this module. Expected version "^12.18.3 || >=14". Got "12.11.1"
```

Any ideas on how to solve this? cc @cowboyd @jbolda

cowboyd (commented):
So this is going to be a really annoying change to make, but I feel like it is a really important one, and that is that we should brand this as an "inspector" and not a "debugger". Not only is that the norm in the JavaScript community and so we'd connect with prior experience, but I see this tool being very much a tool of inquiry and education and not just something to use when you encounter a problem. 

For example, my hope is that a developers could use this to show each other how the system works. I can also see adding memory and performance profiling capabilities to this as well.

cowboyd:
> Should we randomize the debugger port? Technically this works fine, but a fixed port has the advantage of being "stable" so a simple refresh in a browser window reconnects. On the other hand, collisions with other processes are a problem. A random port avoids this problem.

One option here is the concept of port "affinity" where we try and select the same port consistently. It's a bit more complicated, but could provide for a really nice experience. It might look like:

1. try a fixed port (default 8800 or something constant)
2. If fixed port is not available, see if `.effection-port` is available in the current working directory. If so, read the port contained therein, delete it and goto (1)
3. If port is different to default, write the used port to `.effection-port`, otherwise, delete `.effection-port`
4. at some interval check if the original default port is available, and if so, delete `.effection-port` if it exists

I'm also happy to punt on this and try to figure out as an enhancement.

jnicklas:
Really love the idea of naming this `inspector`. I think `debugger` can be a bit intimidating. It sort of evokes `gdb` for me.

It actually wasn't so bad to make this change.

jnicklas:
I did this as a new commit, rather than amending so that #343 becomes easier to rebase later on.

cowboyd (approved):

PR #353, opened 2021-06-07
Make `Future` an operation type

jnicklas:
Depends on #355 

## Motivation

Resolution functions where introduced because writing leaf nodes as generators is not always easy, for example formulating `sleep` as a generator is basically impossible. And while it might be possible to write an iterator by hand this is not convenient. The alternative leaf node is a Promise, but promises have the problem that they always resolve asynchronously. This is a problem for example for `@effection/events`, where synchronous resolution is very important to prevent race conditions.

However, resolution functions are awkward. Not only do they require this rather odd `perform` function, but they just feel like a thing which was tacked on to solve a need, without really being excellent in their design.

## Approach

Enter: Futures. Futures were introduced in #350 as a new way of structuring the internals of Effection. A future is in effect something similar to a synchronous promise. Doesn't this sound pretty similar to our resolution functions? What if we just got rid of resolution functions entirely and used only futures instead. This would mean our leaf nodes are always either a promise, or a future, which is in effect kinda-sorta a promise.

In this PR we add futures as an alternative to resolution functions, with an eye towards possibly removing resolution functions in the future.

### Alternate Designs

None

Futures are a bit hard to understand, and their API is a little awkward, so it might be difficult for people to understand these? Then again having to work with them should be rare for user code.

### TODOs and Open Questions

<!-- OPTIONAL
- [ ] Is this a good idea? Should we do this?

### DIFF
--- a/.changeset/happy-elephants-rush.md
+++ b/.changeset/happy-elephants-rush.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Allow futures as an operation type
--- a/.changeset/ten-ants-relax.md
+++ b/.changeset/ten-ants-relax.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Enable tasks to resolve synchronously when used as an operation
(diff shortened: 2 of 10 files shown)


cowboyd:
Can we make _both_ futures _and_ resolutions operations for now?

jnicklas:
> Can we make both futures and resolutions operations for now?

Yes sure. Should we deprecate resolutions, or leave them alone entirely?

cowboyd (approved):

PR #355, opened 2021-06-09
Add FutureLike interface and implement it for Task

jnicklas:
This adds a minimal interface, similar to the `PromiseLike` interface. This skips the `state` property, as well as the requirement of extending promise. This interface is implemented for `Task` so that `Task` can act as a future. We cannot implement the full `Future` interface, since the `state` properties conflict.

### DIFF
--- a/.changeset/selfish-mangos-repeat.md
+++ b/.changeset/selfish-mangos-repeat.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Add `FutureLike` interface and implement it for `Task`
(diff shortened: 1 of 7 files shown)


cowboyd (commented):
Another option is to not have the `state` property on future. Do we actually use it?

jnicklas:
@cowboyd there is currently only one spot where we use it, which is in `finalize`, but that one is annoyingly hard to work around. https://github.com/thefrontside/effection/blob/v2/packages/core/src/task.ts#L179-L180

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection">thefrontside/effection</a></strong></div><div>Effortlessly composable structured concurrency primitive for JavaScript - thefrontside/effection</div></blockquote>

cowboyd:
Oh right. Yeah that is annoying. Hmmm, there is one option available to us. We could add an `isPending` property to the future. But that does mean the task would have to have that property as well, although I can see that being usefull too.

cowboyd (approved):

PR #357, opened 2021-06-09
Rename @effection/node package to @effection/process

jnicklas:
The @effection/node package started off as a sort of catch-all package for things which were dependent on the node runtime. This is no longer really the case, so we should call this package by what it's really for, which is managing processes.


jnicklas:
No changeset needed here, since merging this will publish the `process` package as a new version. I maintained the same version numbering, so that the changelog makes sense.

cowboyd (commented):
There are enough things out there using this that we should make it a pass through and deprecate it to give folks a chance to upgrade. Then, when we release 2.0 we can remove it.

cowboyd (approved):

PR #368, opened 2021-06-10
Update project README

jnicklas:
I started writing this a while ago, and added some more text just now. Maybe it can provide a starting point for a new README.

[Rendered](https://github.com/thefrontside/effection/blob/v2-docs/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,173 +7,151 @@
 
 # Effection
 
-A framework for building concurrent systems in JavaScript and TypeScript.
-
-JavaScript has gone through multiple evolutionary steps in how to deal with
-concurrency. From callbacks and events, to promises and finally to async/await.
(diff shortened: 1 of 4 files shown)


cowboyd (approved):

cowboyd:
@jnicklas I went ahead and tried to package this into a minimal README that would at least have something that wasn't completely stale (our v1 README) since we have released 2.0.0-beta.1 as the default install on npm, and the `v2` branch is the default branch on github.

I was trying to get a feel for the examples at the end and it had the feeling of a tutorial. That's something we will definitely need, but since there wasn't time to really finish it properly, I just condensed it with the other examples to try and make it more of "taste" of the API. We should revisit it, but just wanted to make sure that we're starting from a position of "everything is on the same page" with respect to npm, default branch, and README.

PR #371, opened 2021-06-10
Begin release 2 beta series

cowboyd:
> depends on #369 and #370
## Motivation
We've made `v2` the default branch, the README is almost in place, and so we want the npm package that you download and install to be compatible with what you see when you visit the page and the documentation.

## Approach
This changes the prelease tag to be "beta", and when we publish, now the beta release gets the 'latest' tag, which is what will be selected when somebody just says `yarn add effection`

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,21 +1,7 @@
 {
   "mode": "pre",
-  "tag": "preview",
-  "initialVersions": {
-    "@effection/channel": "1.0.0",
-    "effection": "1.0.0",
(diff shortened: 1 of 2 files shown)


minkimcello (approved):
👍 wasn't sure what was going to happen if we have a new pre-release tag on top of what we have currently (if `2.0.0-preview.1` was going to turn into `2.0.0-preview.1-beta.1`) but it seems changesets already accounts for that so it'll go from `preview.8` to `beta.9`.

PR #372, opened 2021-06-11
Start beta at the beginning

cowboyd:
Motivation
-----------

Changesets was just swapping out the tag, but keeping the version point numbers in place which means that the beta tag was not really starting from anywhere meaningful.

Approach
----------

This levels out all packages to start at 2.0.0-beta.0 by removing the existing changsets. These are already in the changelogs, and so it's ok if we drop them.


jbolda (commented):
Since we don't have any change files now, I think it is going to try to publish. Did we want changelog entries?

It doesn't appear there is a way to go from prerelease to prerelease as all the docs expect an `exit` followed by a `version` which pulls it out of prerelease mode. Perhaps if we ran `exit`, `version`, `enter`, and then `version` again it would create the changelog entries, but I imagine it will still use all the content from every file in the entries.

cowboyd:
@jbolda I added a changeset that will just go ahead and cause a full `beta.1` release for everything.

jbolda (approved):

PR #376, opened 2021-06-14
Add yieldingTo property to Task

jnicklas:
## Motivation

The task an iterator controller is currently yielding to is an important piece of information about a task. This property will be used by the debugger to read the currently yielding to task.

## Approach

Since we want to implement this without coupling the task and controller and without adding a method to the task to set the Task, we implement it via a callback. This isn't super elegant, but I'm unsure of a better approach.

### Alternate Designs

~~I'd really like to bikeshed the name of this, since it starts to leak out into public APIs and `subTask` isn't a great name. I've been racking my brain trying to come up with a better name, but I haven't come up with a decent alternative.~~

It further expands the scope of the Task API

### DIFF
--- a/.changeset/nasty-forks-raise.md
+++ b/.changeset/nasty-forks-raise.md
@@ -1,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Add `yieldingTo` property to task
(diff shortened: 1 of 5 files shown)


.changeset/nasty-forks-raise.md
```diff
@@ -0,0 +1,6 @@
+---
+"@effection/core": minor
+"effection": minor
+---
+
+Add `subTask` property to task
```
cowboyd:
```suggestion
Add `yieldingTo` property to task
```

cowboyd (approved):

PR #378, opened 2021-06-14
WebsocketServer can accept HTTP server and port can be omitted

jnicklas:
## Motivation

Rather than the websocket server always supplying its own http server, we can also send in our own, this makes it possible to use both websocket connections and other HTTP services on the same server.

### DIFF
--- a/.changeset/heavy-walls-stare.md
+++ b/.changeset/heavy-walls-stare.md
@@ -1,0 +1,5 @@
+---
+"@effection/websocket-server": minor
+---
+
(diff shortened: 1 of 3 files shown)


cowboyd (commented):
`server` and `port` should be mutually exclusive, right? Maybe we should have a validation that both are not defined.

jnicklas:
We could change the type of `Options` to:

``` typescript
type Options = { http: http.Server; } | {  port?: number;  };
```

cowboyd (approved):

PR #379, opened 2021-06-14
Add name label to root task

jnicklas:
Makes it look nicer in the debugger

### DIFF
--- a/packages/core/src/effection.ts
+++ b/packages/core/src/effection.ts
@@ -6,7 +6,7 @@ const MAJOR_VERSION = version.split('.')[0];
 const GLOBAL_PROPERTY = `__effectionV${MAJOR_VERSION}`;
 
 function createRootTask() {
-  let task = createTask(undefined, { ignoreChildErrors: true });
+  let task = createTask(undefined, { ignoreChildErrors: true, labels: { name: 'root' } });
   task.start();
   return task;
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/effection.ts
```diff
@@ -6,7 +6,7 @@ const MAJOR_VERSION = version.split('.')[0];
 const GLOBAL_PROPERTY = `__effectionV${MAJOR_VERSION}`;
 
 function createRootTask() {
-  let task = createTask(undefined, { ignoreChildErrors: true });
+  let task = createTask(undefined, { ignoreChildErrors: true, labels: { name: 'root' } });
```
cowboyd, line 9: Two hunches:

From a purely branding perspective, `effection` feels hotter

```suggestion
  let task = createTask(undefined, { ignoreChildErrors: true, labels: { name: 'effection' } });
```

Also, shouldn't we just omit the root task from the visualization?

jnicklas: > From a purely branding perspective, `effection` feels hotter

Given that the root task is accessed through `Effection.root`, I feel like `root` should be in there somewhere. Maybe something like `effectionRoot`, but that feels awkward to me.

> Also, shouldn't we just omit the root task from the visualization?

That is an option too, but it might be instructive to see it, I think?

cowboyd: After discussion, we can leave as-is, in other word, we label it after its name in the API, and we add an option to the debugger to show the root task.

cowboyd (approved):

PR #394, opened 2021-06-15
Update website

jnicklas:
This updates the website and makes it more current. It adopts "Guides"|"Api Reference" as the main navigation and further devides the guides into three sections: "Getting started"|"Advanced"|"Cookbook"

Here "Cookbook" contains the use case examples that were added in #276

I have started filling out the first three guides, and the rest are currently marked "Coming soon"

[Deployed](https://deploy-preview-394--effection.netlify.app/effection/docs/guides/introduction)


website/docs/guides/introduction.md
```diff
@@ -0,0 +1,141 @@
...
+Effection allows you to compose concurrent code so that you can reason
+about its behavior.
+
+
+## Replacing async/await
+
+If you know how to use `async/await`, then you're already familiar with most of
+what you need to know to use Effection. The only difference is that instead
+of [async functions][async functions], you use
+[generators][generators] and replace:
```
jorgelainfiesta: Maybe we can elaborate a bit on what a Generator is in this context? like what do they mean/do in Effection

website/docs/guides/tasks.md
```diff
@@ -0,0 +1,140 @@
...
+sayHello();
+```
+
+### Tasks
+
+We need something to "drive" the iteration of the Generator, and in Effection this is
+a [`Task`][task]. Tasks have other responsibilities as well, as we'll see when we talk about
+[`spawn`][spawn].
+
+We can use the [`run`][run] function from Effection which takes a generator function
```
jorgelainfiesta, line 40: Could we make a diagram to describe de relationship between Generators, Tasks, Operations, and methods like `run` or `main`?

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+import { fetchWeekDay } from './fetch-week-day';
+
+main(function*() {
+  let dayUS = yield fetchWeekDay('est');
+  let daySweden = yield fetchWeekDay('cet');
+  console.log(`It is ${dayUS}, in the US and ${daySweden} in Sweden!`);
+});
+```
+
+This works, but it slightly inefficient. How can we run both `fetch` operations concurrently?
```
jorgelainfiesta: can we elaborate a bit on why it is inefficient?

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Using `async/await`
+
+If we were using `async/await` we might do something like this:
```
jorgelainfiesta: I got lost, are we talking about Effection still or saying that we'd use this format if we didn't use Effection?

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+that the more complex the operations we're trying to combine, the more opportunity
+for problems there are.
+
+We are calling these situations "dangling promises", and most significantly complex
+JavaScript applications suffer from this problem. `async/await` fundamentally does
+not handle cancellation very well when running multiple operations concurrently.
+
+### Effection
+
+How does Effection deal with this situation? If we translate this directly into
```
jorgelainfiesta: If we translate the await solution? what is it that we expected and what did we get instead?

minkimcello (approved):
Read through this last night 👍👍

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+main(function*() {
+  let dayUS = fetchWeekDay('est');
+  let daySweden = fetchWeekDay('cet');
+  console.log(`It is ${yield dayUS}, in the US and ${yield daySweden} in Sweden!`);
+});
+```
+
+This is still serial, and does not run both `fetch` operations concurrently! To
+understand why, remember that calling a generator function does not do
+anything, only by passing the generator to `yield` or `run` do we actually run
```
jorgelainfiesta: I'm afraid I really don't follow the logic or what you're explaining here

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+
+main(function*() {
+  let dayUS = yield spawn(fetchWeekDay('est'));
+  let daySweden = yield spawn(fetchWeekDay('cet'));
+  console.log(`It is ${yield dayUS}, in the US and ${yield daySweden} in Sweden!`);
+});
+```
+
+Like `run` and `main`, `spawn` takes an `Operation` and returns a `Task`. The
+difference is that this `Task` becomes a child of the current `Task`. This
```
jorgelainfiesta, line 119: does this mean that only when using `spawn` will we get the 'well behaved' things mentioned above?

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+```
++-- main [FAILED]
+  |
+  + -- fetchWeekDay('est') [HALTED]
+  |
+  + -- fetchWeekDay('cet') [FAILED]
+```
+
+Since `fetchWeekDay` was implemented using the `@effection/fetch` package, and
+that package knows how to clean up after itself, the `fetchWeekDay('est')`
```
jorgelainfiesta: is it only that package that does that?

website/docs/guides/spawn.md
```diff
@@ -0,0 +1,145 @@
...
+request will be cancelled for you.
+
+Idiomatic Effection code creates tasks that are tied to the lifetime of their
```
jorgelainfiesta: Is there another way of saying this that doesn't include 'Idiomatic'? and also, is it non-idiomatic something we'd like users to use?

jorgelainfiesta (commented):
Great content! I've added comments where I didn't understand something or wanted more information. We can edit these pages for style etc once you're gone so I didn't comment on anything from that realm.

website/docs/guides/tasks.md
jnicklas: I expanded the explanation a bit here instead. Maybe some kind of illustration would still be helpful.

website/docs/guides/spawn.md
jnicklas: Done

website/docs/guides/introduction.md
jnicklas: Sure!

website/docs/guides/spawn.md
jnicklas: It's if we didn't use Effection. I've tried to clarify it a bit.

website/docs/guides/spawn.md
jnicklas: I've rewritten this so hopefully it's clearer.

website/docs/guides/spawn.md
jnicklas: Yes it does!

website/docs/guides/spawn.md
jnicklas: No, on reading this again, I just removed this entire paragraph, it felt more confusing than helpful.

website/docs/guides/spawn.md
jnicklas: In programming-parlance, "idiomatic" has a similar meaning to "the recommended way", but this is a bit unnecessarily jargon-y so I rewrote it without using that word.

jnicklas:
GREAT feedback @jorgelainfiesta. 

I've gone over everything and tried to incorporate your suggestions. I've also added three more guides. These were mucgh harder to write, and I'm not as happy with these. We're getting into explaining difficult nuances, and so explaining it clearly and easily is very hard. Nevertheless, let me know what you think.

website/docs/guides/lifecycle.md
```diff
@@ -0,0 +1,172 @@
...
+
+``` javascript
+main(function*() {
+  let task = yield spawn();
+  yield task.halt();
+});
+```
+
+Halting a Task means that the task itself is cancelled, it also causes any Task
+that has been spawned from the Task to be halted.
```
minkimcello, line 37: Would it be worth mentioning that halting a task won't shut down the parent task the way an error will?

website/docs/guides/lifecycle.md
```diff
@@ -0,0 +1,172 @@
...
+``` javascript
+import { main } from 'effection';
+
+let task = main(function*() {
+  yield // we will `return` from here
+  console.log('we will never get here');
+});
+
+task.halt();
+```
```
minkimcello: I'm guessing the `return()` you refer to is an internal thing?

Regarding the point you're making here, doesn't this apply to normal generator functions too (and not an effection-specific thing)?

Also do we need the `task.halt()` in this particular example?

website/docs/guides/resources.md
```diff
@@ -0,0 +1,215 @@
...
+  socket.write('hello'); // this works
+  // once `main` finishes, the socket is closed
+});
+```
+
+## How resources work
+
+Let's take a look at the implementation of `createSocket`. Effection considers
+any object which has an `init` function a Resource, and so the object returned
+from `createSocket` is a Resource.
```
minkimcello: I think we should mention this _before_ the example on line 88.

website/docs/guides/lifecycle.md
jnicklas: This is a good point, and an important mechanism we haven't talked about yet, I'll try to incorporate this.

website/docs/guides/lifecycle.md
jnicklas: I was struggling to explain this in a good way. The end effect is similar to if you'd have replaced the `yield` with a `return` statement. So in the end, when we call `halt`, the function behaves somewhat similarly to this:

``` javascript
main(function*() {
  return;
  console.log('we will never get here');
});
```

So the reason we don't get to the console log is because we have "returned" from the generator function at the yield point.

If we don't halt the task then it's not very illustrative of the behaviour I'm trying to explain here.

I would love some ideas on how to explain this better though.

website/docs/guides/resources.md
jnicklas: Good call, I've rewritten this slightly

website/docs/guides/lifecycle.md
jnicklas: I've rewritten this section slightly in an attempt to clarify. Let me know if it's more understandable this way.

website/docs/guides/collections.md
```diff
@@ -0,0 +1,336 @@
...
+  queue.send('hello');
+  queue.send('world');
+
+  console.log(yield queue.expect()); // logs 'hello'
+  console.log(yield queue.expect()); // logs 'world'
+});
+```
+
+As you can see, we can push values into the queue using `send`, and then we can
+fetch values from the queue using `expect`.
```
cowboyd, line 35: Queues are a very low-level mechanism that you basically never, ever use; even if you're a really advanced Effection developer. As such, it feels as though it should be introduced at the very end. I see that we need some way of introducing subscription first, but maybe we can use a practical example instead? Either way, it feels like Subscription should get the top billing.

website/docs/guides/collections.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Subscription
+
+You have already met a Subscription, because `Queue` is in fact a
+`Subscription`! Or rather you can think of the *receiving* end of the queue as
+a subscription.
+
+A subscription is a basically an iterator over a stream of values. Taking the
```
cowboyd, line 43:
```suggestion
A subscription is an iterator over a stream of values. Taking the
```

website/docs/guides/debugging.md
```diff
@@ -0,0 +1,203 @@
+---
```
cowboyd: It might be worth first starting on how to interpret an effection stack trace as a first step.

website/docs/guides/debugging.md
```diff
@@ -0,0 +1,203 @@
...
+import { withLabels } from 'effection';
+import { fetch } from '@effection/fetch';
+
+export function *fetchWeekDay(timezone) {
+  let response = yield fetch(`http://worldclockapi.com/api/json/${timezone}/now`);
+  let time = yield response.json();
+  return time.dayOfTheWeek;
+}
+```
+
```
cowboyd: Worth including a note that while we do our best to label a generator function (and async function), it's not foolproof, so when in doubt, just use explicit labels to improve the visualization and debugging experience for you and for any consumers of code?

website/docs/guides/events.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Recurring events
+
+If you've been following the chapter on streams and subscriptions, you may
+already have a feeling that it is not a good idea to repeatedly call
+`once(socket, 'message')` to grab the messages sent to a WebSocket. The risk
+here is that we miss messages if we're not very careful.
+
+Instead we can use `on`. `on` is a very convenient function which takes an
+event emitter and the name of an event, and returns a Stream of values.
```
cowboyd: Technically it takes either an `EventEmitter` (node) or an `EventTarget` (dom). We should probably even point that out since it is actually really handy because it lets you treat all event callbacks on all platforms uniformly.

website/docs/guides/futures.md
```diff
@@ -0,0 +1,99 @@
+---
+id: futures
+title: Futures
+---
+
+We mentioned in the [introduction][] that Effection is fundamentally
```
cowboyd, line 6: missing link?

website/docs/guides/futures.md
```diff
@@ -0,0 +1,99 @@
...
+synchronous. What we mean by this is that any `Task` *could* complete
+synchronously. Most of the time, tasks do take a while to complete, but
+sometimes they do not, and can in fact complete instantaneously.
+
+This is different from Promises and `async/await`. The Promise specification
+requires that promises *always* complete asynchronously. And while they might
+complete *very quickly*, they always have to release the event loop to do so.
+
+There are good reasons for this behvaiour in promises, mostly related to error
```
cowboyd:
```suggestion
There are good reasons for this behavior in promises, mostly related to error
```

website/docs/guides/futures.md
```diff
@@ -0,0 +1,99 @@
...
+main(function*() {
+  let event = yield once(document.body, 'click');
+  event.preventDefault();
+  console.log('you clicked!');
+});
+```
+
+If our `yield` point released the event loop after the click occurred, then the
+default action has already happened by the time we call `preventDefault`. This
+is no good, we need `once` to resolve *synchronously*.
```
cowboyd, line 35: This is a _fantastic_ example.

website/docs/guides/collections.md
jnicklas: Hmm, I've actually used `Queue` a lot, I disagree that it's so low level.

This section was pretty challenging to write. Especially the order of things. You can't really talk about subscriptions without talking about some way of creating them. There are currently two principal ways of creating a subscription: from a stream, or from a queue. The former requires us to talk about streams first, but we can't really talk about streams without also mentioning how to consume them... which requires a subscription. So that's how I landed on talking about queues first.

website/docs/guides/futures.md
jnicklas: I don't think so? Seems to work fine.

cowboyd: TIL 🤯🤯

cowboyd (approved):

PR #396, opened 2021-06-15
Configure CI for API docs

### LINKED ISSUE #383: Add workflow to generate and publish documentation
We want to have an automated workflow which generates and uploads the documentation.

- [ ] When should this be done? On release of a new version, or for each commit?
- [ ] Should we use GitHub actions or CircleCI, since we use both on this repo
- [ ] How should the docs be hosted?
- [ ] Should we keep old versions of docs around?

minkimcello:
## Motivation
To resolve #383

### Typedoc stuff
- Added typedoc as dependency to monorepo instead of individual package
  - Upgraded to Typescript v4 because of `typedoc` (Rebase after ~#419~ #434)
  - To take advantage of the yarn workspaces config for the new monorepo support
  - Added patch to typedoc:
    - Skip ts error check (because we use tsdx)
    - Specify separate entrypoint (because typedoc wants src/ and we use dist/ for publishing)
      - With the patch typedoc will use `typeDocEntry` instead of `main` from each package.json
    - Remove `@effection` scoping from each package in generated website
  - Had to add each individual package in `typedoc.json` to omit the new `inspect` packages
- Added default theme from `typedoc` to local repo so it's easier to style it later. @jorgelainfiesta
- ✨ Removed `core`, `main`, `channel`, `events,` `subscription`, and `node` from typedoc config because they're included in `effection` and `process`

###  Netlify stuff
- Netlify runs the build scripts in their own VM and it's using an older, incompatible version of node (with typedocs) so I added env variables to specify the version we need
- ❗It seems it's necessary to `prepack` before running `typedocs`?
- Redirect set up as follows:
  - `frontside.com/effection/*` => `effection.netlify.app/:splash`
  - `effection.netlify.app/api/*` => `v2--effection-api.netlify.app/:splash`

### ✨CI stuff
- I created a new job _within_ the `version-or-publish` workflow. This is to ensure that we deploy new api docs only _after_ a `version-packages` PR is merged and when packages have been published.
  - Note: It's important that we run the deploy job based on the output of `successfulPublish` instead of `command == 'publish'` because the covector command defaults to `publish` if there are no change files. This means if any non-version-packages PR is merged while there are no change files in the main branch, it will run `publish` and exit when it detects there are no new versions.
  - The new job, which triggers when `successfulPublish` is `true`, is to generate api docs in github actions in order to hash and check the cache to see if there are any new changes to the api docs. If it detects that the docs is different, that's when we trigger build/deploy on netlify via build hook.
    - The reason why we build and check for cache hits on github actions before netlify is for managing resources. We have a lot more free minutes in github than we do in netlify. We can remove the caching steps if we need more github actions minutes.`Doesn't netlify has its own caching system?` - It does but building the api docs is pretty costly because we need to `prepack` beforehand which is why we're offloading that responsibility over to github actions instead.

Screen Shot 2021-07-15 at 8 07 40 AM

<details>
<summary>Old PR</summary>
<br>

### outdated CI stuff
- The original motive for #420 was to use its output to determine if we should deploy API docs. Now instead of doing that, we're building typedocs in github actions and caching it to detect if there are any new changes. However, this doesn't solve the problem of prematurely updating the api docs before they're actually published. (When a new feature branch is merged, the workflow would detect there are new changes to the API docs and deploy it right away). So once #420 is merged, I will incorporate the output from the covector action to determine whether or not we should check cache (because it requires resources to `yarn && yarn prepack && yarn docs`).
  - See TODOs.

Screen Shot 2021-06-30 at 1 17 11 PM

### API Docs

Right now we're generating API docs using `typedoc` and the guides part of the website with `docusaurus`. 

Docusaurus has [plugins](https://www.npmjs.com/package/docusaurus-plugin-typedoc) that are supposed to work with typedoc by generating the typedoc in markdown. The upside to this approach is docusaurus does all the building for both the guides and API docs in one go. The downside is that the styling of docusaurus overrides typedoc's default layout. In the two screenshots below, the website on the left is when we use the docusaurus-typedoc plugin and on the right is the default styling of typedocs:
interface-slice
modules

In our discussions we talked about how we could have two separate websites and route `.com/effection/api/v2` to the second website. So then our first todo would be to generate typedocs for our API. Typedocs doesn't offer too much support for monorepos right out of the box. However, there is [this](https://github.com/unstubbable/typedoc-plugin-monorepo) plugin. I haven't spent too much time with it but seeing as how typedoc is configured as a peerdependency from a range of >0.20 to <1.0 and typedoc is currently at 0.21, it makes me hopeful that it could actually work. I'll invest more time into it.

However, the main concern of generating the API website separately is styling. Maybe there's an easy way of maintaining consistency between both sites? But my instinct is that keeping all styles in one ecosystem would be easier to manage.
Screen Shot 2021-06-18 at 7 13 48 AM
_Red line indicates that we somehow need to apply Docusarus' theme/style to Typedoc_

I'm not too familiar with how docusaurus works (yet) but if there's a way to override and apply a separate style (by putting like a template at the root of `/api/v2`) for the typedoc pages, using the docusaurus-typedoc plugin would be most robust and I think the easiest to maintain? Thoughs? @jorgelainfiesta 

### The actual CI part
Our netlify setup for bigtest (and also effection) triggers builds/deploys whenever we push commits to the main branch. This is actually a little problematic for the bigtest website too. If I make changes or introduce new features, I update the docs in the same PR. When that PR gets merged, the actual features aren't published until we merge the subsequent changesets' `Version Packages` PR but the website gets published with the first PR. So there's a small gap between when we merged a Version Packages PR and when the feature PR is merged.

I think the solution to this is to create a workflow that detects if there are changeset files or not. If there are no changesets, we call a [`build hook`](https://docs.netlify.com/configure-builds/build-hooks/) to trigger a new build/deploy. If not, we assume it's a feature PR and don't deploy a new site.

One downside to this is that if we need to fix a small typo, we would submit the correct new version of the website and it won't be updated until we merge a `Version Packages` PR.

I haven't checked yet but if the changeset action outputs the [`published`](https://github.com/changesets/action#outputs) boolean regardless of if we utilize their publish command, then we could use that as a variable to decide if we should run the build hook for the website. And on top of that we could attach a `or` conditional to check for a `website` PR label so that we can force a deploy even if there are changset files. 🤷‍♂️
</details>

## TODOs
- [x] ~actually implement api docs build/deploy to the release PR and configure it so it only runs if there are no changeset files~
  - [x] ~confirm changeset output in actions and use it to add a conditional~
    - [x] ~need to merge #420~
  - [x] put netlify build hooks in secrets
    - [x] hash and cache the docs and trigger build hook only when hash is different
- [x] redirect needs tweaking
- [x] ~#419~ #434
- [x] #420
  - [x] Incorporate output from covector to decide whether or not we should deploy new API docs.
- [x] When ready to finalize, I need to modify the workflows to run on `v2` and update the redirect link (website/netlify.toml).


jorgelainfiesta:
About the styles, I think it's very unlikely that we'll be able to reuse Docusaurus styles over Typedocs. Right now, unfortunately, all of our websites have different styling mechanisms so it's a cost I think we have to take.

Docusaurus styles are built on top of an experimental Facebook styles framework ([Infima](https://github.com/facebookincubator/infima)), which seems to be a re-write of Bootstrap. I don't know Typedocs styles system but I'm fairly confident it's not the same.

That said, I'll be happy to add some style sprinkles to the API site. I wouldn't aim for full integration with docusaurus at the moment.

minkimcello:
@jorgelainfiesta Oh I see. I assumed that would be the case. But what I was suggesting is that with the plugins, typedoc gets converted to markdown and effectively is basically a docusaurus website. So what I was getting at was... if we could give the docusaurus-version of typedoc a specific template to style, wouldn't that be easier for you than having to maintain two separate tools for styling of one website?
```
+ docusaurus
|  public/
|    apply main theme (docusaurus)
|      guides
|      guides
|  docs/api/v2
|    apply custom api theme for pages in this directory 
|          // (using docusaurus template (idk if this mechanism exists or not))
|      api
|      api
```
versus
```
+ docusaurus
|  theme (docusaurus)
|    guides
|    guides

+ typedoc
|  separate theme, different tool
|    api
|    api
```

jorgelainfiesta:
Yeah that would be indeed easier to style. We can try how far we can get in that approach. I think the problem lies in the translation from typedoc to markdown, because it doesn't have as many resources to express styles. I might be wrong, these are just the conclusions I got to when we tried this last time (using that plugin to translate typedoc to markdown, the outcome was completely rejected by Jonas and Charles; I didn't think it was _that_ bad, but they preferred having no API for BigTest than having that so maybe it really was not great)

cowboyd:
The problem is that the linkability and navigation structure is the most important part of API docs, and that aspect of it just did not work very well with the translation. While it looked better, it was cumbersome to navigate.

@minkimcello Is there a way to generate a cache key from the typedocs? That would seem like a more general solution. I wonder if anybody has tried it.

minkimcello:
> The problem is that the linkability and navigation structure is the most important part of API docs, and that aspect of it just did not work very well with the translation. While it looked better, it was cumbersome to navigate.
> 
> @minkimcello Is there a way to generate a cache key from the typedocs? That would seem like a more general solution. I wonder if anybody has tried it.

I wasn't able to find anything on caching for netlify nor typedocs. :/

cowboyd:
> I wasn't able to find anything on caching for netlify nor typedocs. :/

I think we'll have to generate our own caching scheme then. We can hash the files that contribute to the api docs, and every time we publish store the hash somewhere. Then, before each publish, we can check and see if the published hash is the same as the local hash, and only publish when they differ.

cowboyd:
@minkimcello we might have some inspiration on building subdirectories here https://nx.dev/

<blockquote><div><strong><a href="https://nx.dev">Nx: Smart, Extensible Build Framework</a></strong></div><div>Nx is a smart and extensible build framework to help you architect, test, and build at any scale — integrating seamlessly with modern technologies and libraries while providing a distributed graph-based task execution, a robust CLI, computation caching, dependency management, and more.</div></blockquote>

.github/workflows/covector-version-or-publish.yml
```diff
@@ -48,3 +47,22 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - run: |
+          yarn
+          yarn prepack
+          yarn docs
+      - uses: actions/cache@v2
+        id: typedoc
+        with:
+          path: docs/api/v2
+          key: typedoc-cache
```
cowboyd: I'm not quite following how the cache will work. To me it looks like it is generating the docs every time on a successful publish, and only trying to restore the cache _after_ the docs are already generated. Also, because the cache key is static, it will only ever work once.

Is successful publish enough here to only have the api docs publish on a release? If I understand correctly, this will only be true when covector _actually_ publishes new versions to NPM? If that's the case, then yes, we have the possibility that the docs will publish for cases where we make a release that only has internal changes. However, that's still a fraction of releases which is a fraction of all PRS which sounds like we'd be avoiding the vast majority of the unnecessary work.

If that's not the case, then I think we'll need to:

1. restore the cache _before_ generating the docs
2. only generate the docs on a cache miss
3. make the cache key a derivative of all the public version numbers of the packages

minkimcello: Yes, the point of the workflow is to prevent publishing new docs until we know the new changes have been published to npm. Otherwise we might have new changes in docs while the version-bump is still stuck in version-packages pr.

I'm not sure how we can check for cache hits without actually building the docs. Alternatively we could just omit the caching and building in github actions and let netlify take care of it.

cowboyd: > I'm not sure how we can check for cache hits without actually building the docs

This is what the cache key is all about. You generate the cache key from the sources which is presumed to be inexpensive. For example, we could concatenate all the versions together, so the cache key might be:

```
@effection/core-2.0.0-beta.5;@effection/subscription-2.0.0.beta.6;@effection/main-2.0.0-beta.1;effection-2.0.0-beta.4;
```

If that changes, then you get a cache miss and need to regenerate the docs. As it stands here, the key `typedoc-cache` will only ever miss once, and will forever hit after that.

minkimcello: if i'm understanding you correctly, basing it on the package versions, we won't need the cache mechanism because covector will tell us if new packages have been published

> If that changes, then you get a cache miss and need to regenerate the docs. As it stands here, the key typedoc-cache will only ever miss once, and will forever hit after that.

```
- uses: actions/cache@v2
  id: typedoc
  with:
    path: docs/api/v2          <---- 👈 
    key: typedoc-cache
```
We're caching the directory of the generated docs (`./api/docs/v2/`) so it should miss every time the docs' contents or styles have changed... ?

cowboyd: > Alternatively we could just omit the caching and building in github actions and let netlify take care of it.

As long as this only ever happens after an actual publish to NPM, and not just every time we check a publish, then we should be good.

minkimcello: 👌 i removed the caching but left the netlify build hook in its own job so it shows as a separate check (and easier to see if it was skipped or not).

cowboyd: > We're caching the directory of the generated docs (./api/docs/v2/) so it should miss every time the docs' contents or styles have changed... ?

`path` (and the files therein) is the value that is cached, `key` is what determines if there is a miss.

You can think of it as a key->value map. If the key exists in the cache, then you have a hit and pull the value. If the key does not exist then you have a miss and you need to generate (and then save) the value. In this case, the key is `typedoc-cache` which never changes, so you will always hit (except for the first time)

minkimcello: ohh I see, I misused the key

cowboyd (approved):

PR #408, opened 2021-06-25
Adjust netlify setup for effection website for effection-api site

minkimcello:
## Motivation
We are trying to build two websites from one repository (#396) and we need to make sure the configurations don't overlap. Netlify describes how you can setup multiple netlify sites in a monorepo [here](https://docs.netlify.com/configure-builds/common-configurations/monorepos/#build-multiple-sites-from-a-monorepo); it's pretty straight forward when we have separate packages for each site.

But at the moment we have the `effection-api` setup to run from the root of the monorepo which requires us to configure at root and inside `./website` and that gets a little funky because any `netlify.toml` in the repo will override the netlify UI settings. This becomes an issue if we have `netlify.toml` at root for `effection-api` and one in `./website` for the main website because when we need to build for the main website, the one at root will override any settings we specify in netlify UI.

Long story short, we need to move the original netlify config file elsewhere and set some of the configs in the netlify UI.

## Approach
- Just moved the `netlify.toml` to `website/` and removed the base property as it's already in the base directory.
  - I will specify the base property in the UI (listed in the TODO below)
  - Added a redirect placeholder for `/effection/api/*` to see if it conflicts with the first redirect.
- Simplified docusaurus build script
- For the `effection-api` site, I'll be configuring its base, command, and publish directory in the UI so that its config doesn't obstruct the `./website`'s site.

## TODOs
- [x] Once this PR is approved, I will change the Netlify configuration (in their UI) so that the base directory for the Effection website will be set to `./website/` before merging.


website/netlify.toml
```diff
@@ -12,3 +11,9 @@
 [[redirects]]
   from = "/*"
   to = "/effection/:splat"
+
+[[redirects]]
+  force = true
+  from = "/effection/api/*"
+  status = 301
```
cowboyd: Why is it that we use 200 elsewhere, but 301 here?

cowboyd (commented):
Looks good, just have the one question

website/netlify.toml
minkimcello: oh oops, that's supposed to be 200

cowboyd (approved):

PR #412, opened 2021-06-30
Allow anything labellable.

cowboyd:
Motivation
----------

As part of making everything labelled, it became obvious that there are certain things which are _not_ opeartions that need to be labelled in order to make the operations that they give rise to be labelled. For example, a `Stream` is not an operation, but `stream.forEach` is. Therefore, we want to make sure we can name the stream, and then label that stream's operations. E.g.

```
on('message').forEach()
```

Approach
----------

This relaxes the constraint that the only thing that can be used with `withLabels` is an Operation, and instead says, if it is labelable, then you can label it. This will allow us to label things like streams, subscriptions and channels.

### DIFF
--- a/.changeset/with-label-any-labelled.md
+++ b/.changeset/with-label-any-labelled.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": patch
+"effection": patch
+---
+label anything with the `Labelled` interface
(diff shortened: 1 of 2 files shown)


jbolda (approved):
✨

PR #413, opened 2021-06-30
Label the suspend operation

cowboyd:
Motivation
-----------

Currently in the inspector, whenever you are suspended indefinitely, it just shows us as 'task'. This is not particularly helpful since it's unclear what exactly is going on. Instead, we'd like a clear indicator that the operation is a suspend operation and will never, ever return by itself.

Approach
----------

We handle this in the create task and check if the operation is undefined, if so, we assign it the `suspend` label.

### DIFF
--- a/.changeset/label-suspend.md
+++ b/.changeset/label-suspend.md
@@ -1,0 +1,5 @@
+---
+"effection": patch
+"@effection/core": patch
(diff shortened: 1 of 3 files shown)


jbolda (approved):
✨

PR #414, opened 2021-06-30
label the main entry point

cowboyd:
Motivation
-----------

It is very common to use an anonymous generate function as the entry point for a main operation. For example, let's take a program that run an http server and a websocket socket server. It might look like:

```js
main(function*() {
  yield createHttpServer();
  yield createWebSocketServer();
  yield sleep();
})
```

Currently, this would be visualized roughly in the inspector as:

```
- main platform=node
  - task
    - http server
    - websocket server
    - sleep
```

But what is that anonymous task? It's silly to have it just sitting there not saying what it is, when it is the entry point to the program.

Approach
----------

This change, implicitly labels the entry point as such when it does not have a label itself. If the entry point is labelled, then it defaults to that.

After:

```
- main platform=node
  - task
    - http server
    - websocket server
    - sleep
```

### DIFF
--- a/.changeset/label-main-entry-point.md
+++ b/.changeset/label-main-entry-point.md
@@ -1,0 +1,4 @@
+---
+"@effection/main": patch
+---
+label the main entry point for use in the debugger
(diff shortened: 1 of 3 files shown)


jbolda (approved):
✨

PR #420, opened 2021-07-01
Add Covector

minkimcello:
## Motivation
Changesets' action does not provide helpful outputs for what we need to set up API docs' CI (#396) so I figured it's as good as time as any to adopt covector.

## Approach
- Made sure to use Jack's github token and NPM token with the covector setup
- Turned off github releases
- Initiated the pre-release mode with `beta` tag that's currently configured with changesets' pre-mode
- Removed changesets dependency but left the workflow for a smoother transition
  - Anyone that adds changesets files and merges their PR after _this_ PR is merged will still go through the normal process. The only difference is that covector will be responsible for publishing now instead of the release workflow.

## TODOs
- [x] Merge [#217](https://github.com/jbolda/covector/pull/217) in covector and update covector dependency
- [x] ~Should we have the version packages PR automatically publish previews? @cowboyd~
- [x] Will covector automatically detect that the packages are already in prerelease mode? Or will there be a conflict there? (Because we're "entering" pre-release mode from package versions that are already in pre-release mode (from changesets) so I guess what I'm asking is what's the purpose of the [`changes` property in `pre.json`](https://github.com/thefrontside/material-ui-interactors/blob/v4/.changes/pre.json)).


jbolda:
I think I would recommend also adding in the status workflow to help as an error catch.

So this first step is using changesets to handle the version PR, and covector to publish? I suspect we will have to have some kind of coordination or rebasing of existing PRs to switch over the publish in the future. We have a lot of draft PRs currently, but since they are all draft, it might not be the worst timing to switch publish as well.

The `pre.json` in `.change` will tell covector that it is in prerelease mode. The `changes` property in `pre.json` helps to determine what bumps have already been applied. If a major change is added, it bumps the major and adds the prerelease. If there is another major added, it needs to only apply the next prerelease number. So to switch `version` over from changesets, I think we can copy all the change files and the [changsets](https://github.com/thefrontside/effection/blob/22a6d0e70997e31d4f8c80e6ed1bda3884e110d9/.changeset/pre.json#L4-L19) value over and covector can pick up from there.

minkimcello:
@jbolda

> So this first step is using changesets to handle the version PR, and covector to publish? I suspect we will have to have some kind of coordination or rebasing of existing PRs to switch over the publish in the future. We have a lot of draft PRs currently, but since they are all draft, it might not be the worst timing to switch publish as well.

The rebase won't be necessary. The migration to covector process I had in mind will allow developers to continue using changesets in current PRs and be forced to use covector in new PRs.

> The `pre.json` in `.change` will tell covector that it is in prerelease mode. The `changes` property in `pre.json` helps to determine what bumps have already been applied. If a major change is added, it bumps the major and adds the prerelease. If there is another major added, it needs to only apply the next prerelease number. So to switch `version` over from changesets, I think we can copy all the change files and the [changsets](https://github.com/thefrontside/effection/blob/22a6d0e70997e31d4f8c80e6ed1bda3884e110d9/.changeset/pre.json#L4-L19) value over and covector can pick up from there.

So I take it it's okay to enter into covector's prerelease mode from package version that's already in pre-release mode? Or do we need to exit out of changeset's prerelease mode before moving to covector's prerelease mode?

I don't think we should convert the current changeset files to covector change files. That forces us to make sure each PR is rebased properly and if someone accidentally reintroduces changeset files, we'll end up messing up the changelog and double bump.

jbolda:
> > So this first step is using changesets to handle the version PR, and covector to publish? I suspect we will have to have some kind of coordination or rebasing of existing PRs to switch over the publish in the future. We have a lot of draft PRs currently, but since they are all draft, it might not be the worst timing to switch publish as well.
> 
> The rebase won't be necessary. The migration to covector process I had in mind will allow developers to continue using changesets in current PRs and be forced to use covector in new PRs.

If changesets is still making the versions PR, new PRs that use covector won't be picked up in the changeset versions PR though. Or am I missing something?

> So I take it it's okay to enter into covector's prerelease mode from package version that's already in pre-release mode? Or do we need to exit out of changeset's prerelease mode before moving to covector's prerelease mode?

It should be OK as long as covector is aware of what has been bumped. In this case, I think it is `major` everything. I believe that only matters if it is running `version` though, so it shouldn't matter either way in this case I don't believe.

minkimcello:
> If changesets is still making the versions PR, new PRs that use covector won't be picked up in the changeset versions PR though. Or am I missing something?

During the transition phase, new PRs created and merged will generated covector's version PR, and the pre-existing PRs with changesets will generate changesets' version PR.

Neither the changesets nor change files get "consumed" until their version PRs are merged. So say if we had two PRs with a major change and a major changeset for package-A, the moment one of those PR is merged, the other PR would get rebased with the new version, no? Since the version PRs get updated on push to main branch.

minkimcello:
@jbolda any idea why the status could be failing? The error message in console doesn't say much.

jbolda:
@minkimcello I am going to go out on a limb and guess that I didn't error check on the `pre.json` load, and it expects a `changes: []` to exist.

minkimcello:
> @minkimcello I am going to go out on a limb and guess that I didn't error check on the `pre.json` load, and it expects a `changes: []` to exist.

So I'll take out the status workflow until that gets fixed?

jbolda:
Try adding it into the `pre.json`. I believe that if `status` errors, then `version-or-publish` likely won't work either.

`pre.json`:
```json
{
  "tag": "beta",
  "changes": []
}
```

minkimcello:
@jbolda So this one is good to go if the config looks good to you.

jbolda (approved):
Looks good to me!

Issue #360, opened 2021-06-09
task.spawn as operation


cowboyd:
Working in code, and writing explainers, it's clear that whenever you `spawn`, it is safest as an operation. In fact, there is the standalone operation `spawn(operation)` and `spawn(operation).within(scope)`. But there is a cognitive dissonance between method `spawn` and operation `spawn`. Method spawn is needed for integration from outside an effection tree, but when working from within an operation, you always want to use operation `spawn` because it is safer.

I propose renaming `task.spawn` to `task.evaluate`, and introducing `task.spawn` as an operation. That way instead of saying

```ts
yield spawn(operation).within(scope);
```

you can instead just say:

```ts
yield scope.spawn(operation);
```

Because let's face it, when you find out at the end of reading an operation which scope it will be in, it's going to be a nasty suprise. E.g.

```ts
yield spawn(function*() {
  yield doThis();
  yield doThat();

  yield spawn(function*() {
    yield doOtherThing();
    yield doYetSomethingElse();
 });
}).within(scope) // psyche! It's in the parent. I hope you were paying attention!!!
```

by making `spawn()` mean the same thing in every context, it cuts down on potential confusions. In other words, `spawn` is `spawn`, and it's what you use within operations to start other concurrently executing operations. `task.evaluate(operation)` is something else entirely and what you use _outside_ of operations.

jnicklas:
I'm not entirely opposed to this, it makes sense semantically, IMO.

We should be clear that this is a pretty significant API change at the 11th hour, but maybe that is justified.

We'd need to bikeshed the name of the current spawn method though. I feel like `evaluate` is confusing, it sounds like the operation is evaluated within the context of the task, which is not the case, it runs as a child. Last time around we discussed `run`, but it has the same issue.

cowboyd:
I'll throw together a PR today.

PR #426, opened 2021-07-06
rename task.spawn -> task.run()

cowboyd:
## Motivation

There is a more complete writeup and discussion for this change in #360 but the long and the short are that `task.spawn()` the function and standalone `spawn()` the operation have different semantics and should be named different things.

## Approach

This renames `task.spawn()` to `task.run()` which is in keeping with the static definition of `run()`. Global `run()` is a function that turns an operation into a task as the child of the root task, there fore `task.run()` is a function that turns an operation into a task as the child of `task`.

`task.spawn()` is deprecated in favor of `task.run()`.

### DIFF
--- a/.changeset/spawn-is-run.md
+++ b/.changeset/spawn-is-run.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": patch
+"effection": patch
+---
+rename Task.spawn() -> Task.run()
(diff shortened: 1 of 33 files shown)


dagda1:
You could add the [jsdoc @deprecated comment](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#deprecated-support) on `spawn`.

```ts
  /** @deprecated Use run instead */
  spawn<R>(operation?: Operation<R>, options?: TaskOptions): Task<R>;
```

Editors like vscode will put a line through on anything marked with it

dep

dagda1 (approved):

PR #430, opened 2021-07-13
Fix some typos in docs

wKich:
## Motivation

Fix some typos in docs


website/docs/guides/collections.md
```diff
@@ -124,7 +124,7 @@ main(function*() {
 ```
 
 If we unpack this a bit, we can see that we're creating a new `Stream` called
-`textStream` using the method `map` on the stream. This stream gets the `text`
+`channel` using the method `map` on the stream. This stream gets the `text`
```
cowboyd: I think what we want here is 

```
If we unpack this a bit, we can see that we're creating a new `Stream` called 
`textStream` using the method `map` on `channel`. This stream gets the `text`
property from each value in the stream
```

website/docs/guides/spawn.md
```diff
@@ -109,8 +109,8 @@ import { main, spawn } from 'effection';
 import { fetchWeekDay } from './fetch-week-day';
 
 main(function*() {
-  let dayUS = yield spawn(fetchWeekDay('est'));
-  let daySweden = yield spawn(fetchWeekDay('cet'));
+  let dayUS = spawn(fetchWeekDay('est'));
+  let daySweden = spawn(fetchWeekDay('cet'));
```
cowboyd: Operations don't actually do anything until they are evaluated, and in order to do that, they must be yielded to. 

`spawn()` is an "higher order" operation. It takes an operation and returns an operation that evaluates that operation concurrently. But, like every other operation, you must yield it.

For example:

```ts
// this just defines the operation, but has no state, and no side-effects
let spawnGetWeekDayCET: Operation<Task<Response>> = spawn(fetchWeekDay('cet'));

// start two separate tasks each which are doing the same operation.
let task1 = yield spawnGetWeekDayCET;
let task2 = yield spawnGetWeekDayCET;
```
So the spawn is definitely necessary here.


> Note: In Effection, side effects happen at yield points. and the code between yield points is generally side-effect free.

cowboyd (changes requested):

website/docs/guides/spawn.md
wKich: So the `yield spawn` produce a task, and when we want to get a value from that task we need to `yield task`? And `dayUS` is actually not a value, it's a task, isn't it?

cowboyd: in your example:

```ts
let dayUS = spawn(fetchWeekDay('est'));
```

`dayUS` is an `Operation`, not a `Task`. In order to evaluate an operation you `yield`.  Internally, this will create a task to "run" the operation, and then wait until the task is complete. `spawn` takes an operation and returns a _new_ operation that evaluates to a task running the original operation A.

wKich: Oh, now I understand. Thanks

cowboyd: If you can fix the other portions of the PR, I say let's get it merged!

cowboyd:
@wKich ping

wKich:
@cowboyd done, sorry for delay

cowboyd (approved):
🙏🏻

website/docs/guides/collections.md
jnicklas: Yes, that sounds correct. `textStream` is the channel we create, so the original correction is not quite correct, but maybe that is indicative of this sentence not being quite clear enough. @cowboyd's version reads a bit clearer IMO.

wKich: And I used it ) I can't merge it,  because I don't have rights

cowboyd (approved):

wKich:
@cowboyd I don't have the right to merge it

cowboyd:
> @cowboyd I don't have the right to merge it

I don't either. It looks like the checks are stuck waiting forever. Maybe close it and re-open.

jnicklas:
I rebased this as #489

PR #419, opened 2021-07-01
upgrade typescript to latest

cowboyd:
## Motivation

TypeScript is having problems figuring out what labels mean for function operations, so maybe an upgrade will help?

PR #434, opened 2021-07-15
Upgrade to Typescript v4

minkimcello:
## Motivation
Replaces #419

## Approach
- Upgrade `typescript`, `@types/node`, and `ts-node` to latest
- Create typescript project references
- Remove `tsdx` for regular `tsc`
- Not adding covector files because we're just enhancing the DX environment and it's not affecting the consumed packages 🤷 
- Use preview version of `@frontside/eslint-config` which upgrades all the plugins to the latest versions because they don't support typescript >4.10.0.

## TODOs
- [x] ~~New typescript is not compatible with the eslint package we have in `@frontside/eslint-config` so I'll create a PR to update those. I have resolutions set in this PR just to get it passing for now but I'll revert that and replace it with the new version of `@frontside/eslint-config`~~
  - [x] ~~merge [@frontside/eslint-config pr #33](https://github.com/thefrontside/javascript/pull/33)~~
  - [x] [@frontside/eslint-config pr #33](https://github.com/thefrontside/javascript/pull/33) generated a preview package of `@frontside/eslint-config@2.0.0-0a28874` so I've updated the dependency of that package in this monorepo to that preview version for now.
  - [x] Add changefiles for every package as a follow-on PR to capture the fact that code transpilation is happening with a different TypeScript version #448


packages/process/package.json
```diff
@@ -11,33 +11,26 @@
...
   },
   "author": "Frontside Engineering <engineering@frontside.com>",
-  "license": "MIT",
-  "files": [
-    "README.md",
-    "CHANGELOG.md",
-    "dist/**/*",
-    "src/**/*"
-  ],
+    "license": "MIT",
```
cowboyd: Is the indenting off here?
```suggestion
  "license": "MIT",
```

cowboyd (approved):

cowboyd (approved):

PR #435, opened 2021-07-20
label @effection/process Operations

jbolda:
## Motivation

This begins work on #424. There are a few elements that I have added comments above which I suspect we want to eventually label, but are Streams or such that cannot be labelled yet.

## Approach

I ran a few snippets and viewed them in the inspector. Anything that had a generic term, I applied something more specific if possible.

### DIFF
--- a/packages/process/src/exec/posix.ts
+++ b/packages/process/src/exec/posix.ts
@@ -1,4 +1,4 @@
-import { spawn, Task, Operation, createFuture } from '@effection/core';
+import { spawn, Task, Operation, createFuture, label, withLabels } from '@effection/core';
 import { createChannel } from '@effection/channel';
 import { on, once, onceEmit } from '@effection/events';
 import { spawn as spawnProcess } from 'child_process';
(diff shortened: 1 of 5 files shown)


cowboyd:
@jbolda Streams can have _names_ which are then appended to the name of that stream's operation. 

https://github.com/thefrontside/effection/blob/v2/packages/subscription/src/stream.ts#L32

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection">effection/stream.ts at v2 · thefrontside/effection</a></strong></div><div>A framework for Node and the browser that makes building concurrent systems easy to get right. - effection/stream.ts at v2 · thefrontside/effection</div></blockquote>

packages/process/src/exec/posix.ts
```diff
@@ -59,22 +59,31 @@ export const createPosixProcess: CreateOSProcess = (command, options) => {
       };
 
       yield spawn(function*(task) {
+        yield label({ name: 'exec', state: 'running' });
         yield spawn(function*() {
-          let value: Error = yield once(childProcess, 'error');
+          yield label({ name: 'listen for error' });
+          let value: Error = yield withLabels(once(childProcess, 'error'), {
+            name: 'untilFirst(error)',
```
jnicklas: I'm a bit confused by this naming scheme, it looks like `untilFirst` is a function which takes an error as an argument of `error`, but this doesn't really reflect the implementation. Would it make sense to name this something like `untilTermination` or `trapError` or just `trap` or something?

jnicklas: Maybe we've already established this notation elsewhere? I'm a bit out of the loop at the moment.

cowboyd: Yeah, we don’t really have good guidelines and conventions on how best to label yet. Perhaps we should come up with some

cowboyd (commented):
@jbolda @jnicklas looks like we're a few linting errors away from making this ready to merge?

jnicklas:
Running this program:

``` typescript
import { exec, daemon } from './src/index';
import { sleep, spawn } from '@effection/core';
import { main } from '@effection/main';
import { runInspectServer } from '@effection/inspect-server';

runInspectServer({ port: 47500 });

main(function*() {
  yield daemon("yarn parcel watch src/index.ts");
  yield exec("echo hello world");
  yield exec("yarn ts-node moo.ts");

  yield sleep(5000);

  yield spawn(exec("echo lets get together").expect());

  yield;
});
```

Before

Screenshot 2021-09-02 at 16 32 36

After

Screenshot 2021-09-02 at 16 32 53

jbolda:
@jnicklas thanks for taking this over 🙏 been tied up with simulacrum.

cowboyd (approved):
This is so much nicer!

It does raise an interesting question regarding resources though. What do you do when all of the child resources of the resource task have finished? Should the resource go away? It seems like the `exec()` resources here that are finished should be gone, right?

jnicklas:
> It does raise an interesting question regarding resources though. What do you do when all of the child resources of the resource task have finished? Should the resource go away? It seems like the `exec()` resources here that are finished should be gone, right?

I think there is no general way for us to detect whether a resource is "done". Just because there is no task right now, doesn't mean there won't ever be. Think of something like wrapping express in a resource, where each request would have a task. If there is currently no request, there might not be any tasks, but the resource is still "active" in some sense, and further tasks might be spawned in the future.

I actually think it's kind of sensible that a resource lives forever within the scope it was spawned. For `exec` we do have the operational API `exec().join()` which should be preferred in those cases where the process is short-lived. This won't leave behind any resource zombies.

Issue #429, opened 2021-07-09
Make setting  `Future`s more ergonomic


cowboyd:
A `Future` is like a promise that resolves synchronously. It's similar to a `Maybe` interface in that the only way you can use the future value is to "unwrap" it with the `consume` method:

```ts
future.consume(value => {
  if (value.state === 'completed') {
   // do thing
  } else if (value.state === 'errored') {
  // do other thing
  })
})
```

To consume a future asynchronously, you can just use the normal Promise methods of `then()`, `catch()`, `finally()`

This api is low-level and unlikely to change, because it's unlikely that folks will need to consume futures directly in their own code and so it's ok if this code is a bit more explicit and cumbersome and needing to always be unpacked with an `if` or `switch` statement. What is more likely however is that folks will want to use futures as operations which means that their user-land code will be more concerned with settling a future than consuming it. Unfortunately, our `resolve` api is a bit cumbersome in this regard in that you need to always pass it a valid `Value<T>`, where value is:

```ts
export type Value<T> =
  | { state: 'errored'; error: Error }
  | { state: 'completed'; value: T }
  | { state: 'halted' };
```

So, in order to settle a successful future, you need to pass the `state: 'completed'`. E.g.

```ts
let { resolve, future } = createFuture<number>();

resolve({ state: 'completed', value: 5 });
```
This is awkward, but safe in TypeScript, but both awkward _and_ potentially dangerous in JavaScript where it is easy to mispell 'completed' and throw a great big spanner in the works.

I propose that we rename the current `resolve` function to `produce()` (the opposite of consume). `produce()/consume()` will be the function pairs that do the explicit packing / unpacking of the future value, but then we can add `resolve()`, `reject()` which produce success and error values respectively. These methods will be immediately intuitive to anybody familiar with promises.

> note: I don't think we should include a shortcut for producting a `halt()` value, since this is something that almost never happens from user-land.

### Rollout Process

- [x] rename `resolve()`  -> `produce()` and add a deprecation warning to `resolve()` that way, any code using it is notified. At this point, `resolve()` will just be a deprecated alias to `produce()`
- [ ] change code in all dependent repositories using `resolve()` to `produce()`
  - [x] simulacrum
  - [x] bigtest
  - [ ] covector
  - [ ] any others not mentioned here
- [x] now that `resolve()` is available, introduce `resolve(), reject()`
  - [ ] change code in dependent repositories to use `resolve() / reject()`
    - [ ] simulacrum
[...]

PR #437, opened 2021-07-22
Deprecate Future#resolve in favour of Future#produce

dagda1:
## Motivation

Part 1 of #429 renames `Future#resolve` to `produce`

## Approach

Deprecate resolve and introduce a new `produce` function to the `Future` interface.


cowboyd (approved):

cowboyd:
@dagda1 oops, looks like we forgot a changeset

dagda1:
> @dagda1 oops, looks like we forgot a changeset

I have a covector changeset but the action is still set up to look for changesets.

I think we should get rid of the action that flags this and go all in on covector.

That would mean rebasing the existing PRs that are the old changesets but no big deal.

cowboyd:
Ah, agree 100%

PR #441, opened 2021-07-23
Add jack as author to PR action

minkimcello:
## Motivation
I noticed the commits that are being pushed to apply covector changes in the `package version` prs are authored by the person that triggered the workflow. According to [this](https://github.com/peter-evans/create-pull-request/blob/main/README.md#action-inputs) README, it seems we can add a `author` property to the action to specify who we want pushing those commits.

## Approach
Added:
```
- uses: peter-evans/create-pull-request@v3
  with:
    author: frontsidejack frontsidejack@users.noreply.github.com
```


jbolda (approved):
We might want to make a note to have this functionality when we add PRs opening directly to covector.

PR #452, opened 2021-07-26
Resuming pre-release with Covector

minkimcello:
https://ptb.discord.com/channels/700803887132704931/701818777746145280/869281580147965962

### DIFF
--- a/.changes/covector-bug.md
+++ b/.changes/covector-bug.md
@@ -1,0 +1,22 @@
+---
+"@effection/atom": major
+"@effection/channel": major
+"@effection/core": major
+"effection": major
+"@effection/events": major
+"@effection/fetch": major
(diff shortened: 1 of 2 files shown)


jbolda:
We also need to manually add it to the `pre.json`. This is all things that would have been set if we had started this prerelease series with covector (and eventually will be handled automatically when we add a command to do it).

```
diff --git a/.changes/pre.json b/.changes/pre.json
index f4b6138..4a25367 100644
--- a/.changes/pre.json
+++ b/.changes/pre.json
@@ -1,4 +1,4 @@
 {
   "tag": "beta",
-  "changes": []
+  "changes": [".changes/covector-bug.md"]
 }
```

jbolda (approved):

PR #455, opened 2021-07-26
Add spawn operation directly to Task.

### LINKED ISSUE #360: task.spawn as operation

cowboyd:
## Motivation

resolves #360

## Approach

This adds a `spawn()` operation to the `Task` api which, unsuprisingly, spawns a child in the context of a particular task. Note that this is an _operation_, and not a synchronous function. Calling the `spawn()` method will not actually do anything until the operation is evaluated.

Now, it is very easy and safe to spawn operations within the context of a specific task without having to use the `within()` syntax (which we might deprecate in the future).

### Before

```js
yield spawn(function*() {
  //do stuff
}).within(scope);
```

### After

```js
yield scope.spawn(function*() {
  // do stuff
});

### DIFF
--- a/.changes/task-spawn-operation.md
+++ b/.changes/task-spawn-operation.md
@@ -1,0 +1,5 @@
+---
+"effection": patch
+"@effection/core": patch
+---
+add `Task#spawn` operation to spawn new task with a specific scope
--- a/packages/core/src/operations/spawn.ts
+++ b/packages/core/src/operations/spawn.ts
@@ -2,7 +2,7 @@ import { Operation, Resource } from '../operation';
 import type { Task, TaskOptions } from '../task';
 
 interface Spawn<T> extends Resource<Task<T>> {
-  within(scope: Task): Resource<Task<T>>;
+  within(scope: Task): Operation<Task<T>>;
 }
 
 export function spawn<T>(operation?: Operation<T>, options?: TaskOptions): Spawn<T> {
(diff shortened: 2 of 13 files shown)


dagda1 (approved):
👍

jnicklas (approved):

PR #459, opened 2021-07-28
fix files array in @effection/atom

dagda1:
## Motivation

I deleted the `dist` and `src` parts of the package.json in a previous PR.

## Approach

Readd them

### DIFF
--- a/packages/atom/package.json
+++ b/packages/atom/package.json
@@ -15,7 +15,9 @@
   "license": "MIT",
   "files": [
     "README.md",
-    "CHANGELOG.md"
+    "CHANGELOG.md",
+    "dist/**/*",
+    "src/**/*"
   ],
   "scripts": {
     "lint": "eslint '{src,tests}/**/*.ts'",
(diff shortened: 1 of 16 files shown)


minkimcello:
https://github.com/thefrontside/effection/commit/eea73398477723d3ab1ed1d39a5941fedaa792bc

I'm not sure if it was intentionally added to begin with (or just copied over from other packages) but `inspect-ui`, `inspect-utils`, and `inspect-server` had the `files` property too.

dagda1:
> [eea7339](https://github.com/thefrontside/effection/commit/eea73398477723d3ab1ed1d39a5941fedaa792bc)
> 
> I'm not sure if it was intentionally added to begin with (or just copied over from other packages) but `inspect-ui`, `inspect-utils`, and `inspect-server` had the `files` property too.

They somehow missed my initial cull and are still there.  That is why they are not in the git diff.

minkimcello (approved):

PR #464, opened 2021-08-03
fix: @effection/core to ship dist/esm and dist/cjs

jbolda:
## Motivation

`@effection/core@2.0.0-beta.6` did not ship with the `dist/esm` and `dist/cjs` folders. It appears that copying the `package.json` into the `dist` directory breaks the `files` resolution.

## Approach

It seems that the `dist/package.json` creates a new root. If we add `esm/**/*` and `cjs/**/*` it appears to resolve the subdirs correctly.

### DIFF
--- a/.changes/core-esm-cjs-dist.md
+++ b/.changes/core-esm-cjs-dist.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


dagda1 (approved):
Great spot 👍

PR #475, opened 2021-08-10
Typedoc initial styles

jorgelainfiesta:
## Motivation

Styles needed to be done

## Approach

I only changed the CSS, mainly variables but also some other small details. These are very basic styles, we'd probably want to get further later on.


jnicklas:
The deploy preview looks the same to me:

https://deploy-preview-475--effection.netlify.app/effection/api

Not sure why the changes in this PR are not being applied there

<blockquote><div><strong><a href="https://deploy-preview-475--effection.netlify.app/effection/api">effection</a></strong></div><div>Documentation for effection</div></blockquote>

jorgelainfiesta:
I don't really understand how the deployment works. I think that the API is deployed to a different netlify instance that is not effection's docusaurus.

minkimcello:
@jnicklas the preview of the effection site will still redirect `./api` to the main production site

@jorgelainfiesta the preview was off for some reason but I switched it on (on netlify ui). if you push an empty commit it it should hopefully build a preview of the api site

minkimcello:
oh on second thought, i think the reason why i disabled it to begin with was because it builds PRs that are against the production branch. and the production branch is `dingdong12345`. (because we're triggering builds with a build hook instead of on pushes to a specific branch to prevent premature publishing)

Screen Shot 2021-08-10 at 6 14 59 AM

jorgelainfiesta:
How do we access the preview of this PR then? Also, I think I'm quite confused about what to do to generate the preview now

minkimcello:
i am quite confused too. there are many limitations with netlify and originally i put this api-docs-building workflow together in a way where we don't have to worry about it because how often are we going to need previews for the api docs aside from making style changes? anyway, we have three options:

1. the previews can only be triggered if it's against the production branch. if we toggle on production branch, it will build and deploy per merge which we don't want. the effection website will still point to the branch deploy (v2-effection-api.netlify.com) so it doesn't matter if we have (effection-api.netlify.com) but it would waste a lot of netlify resources.

2. one really crappy way would be to create a fake branch (it'll be designated as the "production" branch that's never pushed to) and change the base branch of this PR from `v2` to that branch to trigger the deploy. but this means you need to remember to change it back to `v2` before merging.

3. another really, really crappy way would be to post screenshots.

cowboyd:
> and the production branch is `dingdong12345`

Thank god I wasn't drinking coffee at the time!

jnicklas:
> and the production branch is dingdong12345

Finally a sensible solution in the debate on what to name the default branch!

jorgelainfiesta:
Screenshot 2021-08-10 at 18 07 52

Screenshot 2021-08-10 at 18 08 30

jnicklas (approved):

cowboyd:
@minkimcello After the merge to the production branch, it still appears that they styles were not applied. 

What if we just always build the previews against PRS to `v2`? Is that an option?

minkimcello:
@cowboyd the reason why it hasn't been updated is because it only builds/deploys when the version PR is merged. unfortunately there's no way for the workflow to know if we're merging a pr strictly for the purposes of updating the api docs.

we could enable previews on PRs against v2 but that means every feature pr that is merged will trigger a build/deploy but won't actually affect the website. and the reason why it was disabled was because of the assumption that we'll have a lot more effection feature PRs than website PRs but i can go ahead and enable it if that's what you want.

cowboyd:
>  and the reason why it was disabled was because of the assumption that we'll have a lot more effection feature PRs than website PRs but i can go ahead and enable it if that's what you want.

That's definitely true that we will continue to have more feature PRs, but I guess the question is, what do we lose by "overbuilding" If it's money, how much money?

jbolda:
Netlify very recently added better support for building multiple sites from a monorepo. If nothing in the specific site subdirectory changed, then the build no-ops. We might be able to make good use of that here.

https://docs.netlify.com/configure-builds/common-configurations/monorepos/#build-multiple-sites-from-a-monorepo

<blockquote><div><strong><a href="https://docs.netlify.com/configure-builds/common-configurations/monorepos/">Build from a subdirectory or monorepo</a></strong></div><div>Netlify builds, deploys, and hosts your front end.
  Learn how to get started, see examples, and view documentation for the modern web platform.</div></blockquote>

PR #476, opened 2021-08-10
Expand and collapse tasks in the inspector

jnicklas:
There can be a lot of information in the inspector, it is useful to be able to expand and collapse individual tasks, to minimize noise.

Screenshot 2021-08-10 at 11 37 59
Screenshot 2021-08-10 at 11 38 03


cowboyd (approved):
Looks great!

PR #480, opened 2021-08-11
Introduce task scope

jnicklas:
## Motivation

This introduces the idea of task scope, as a lower level construct which can be used instead of resources.

Currently, some base operations, such as `spawn` and `label` are based on resources, because resources are the only escape hatch allowing one to "expand the scope" as we like to say, or it can be seen as "inlining" an operation.

However this has presented a problem for us: we want resources to introduce a task which "groups" a resources tasks, which can be seen in the inspector. But adding this additional layer is incompatible with e.g. `spawn` being a resource.

## Approach

This solves the problem by introducing the concept of task `scope`. Every task which is not a root task has a scope. Normally this scope is the parent task, but it does not necessarily have to be.

This scope is accessible to "inlining" operations such as `spawn`, and also indirectly to resources, since the newly introduce resourceTask is running inside the scope.

In this PR, there is no convenience added to access the `scope` for operations. Instead they must us `task.options.scope` to obtain the scope. This is also partly meant as "syntactic vinegar" since such operations probably should not be the norm.

### Alternate Designs

We have explored many, many alternative designs to this approach, but crucially it is very difficult to find an approach which composes well, and that is the main advantage of the current system

We have seen with v0.1 of effection that using just a notion of `parent` does not compose well for building resource-like structures. This is why resources were introduced in the first place. There is some risk of repeating this mistake, but it is mitigated by the fact that we now have a solid resource system.

There are some complications with higher order operations. While this does prepare the way for making higher order operations such as `all` work better with resources, it is unclear what the exact semantics should be in the case of an error.

### TODOs and Open Questions

- [x] how does this interact with higher order operations?
- [x] do we provide access to the resource task to a consumer of the resource? and if so, how?

### DIFF
--- a/packages/core/src/controller/iterator-controller.ts
+++ b/packages/core/src/controller/iterator-controller.ts
@@ -45,7 +45,7 @@ export function createIteratorController<TOut>(task: Task<TOut>, iterator: Opera
           produce({ state: 'completed', value: next.value });
         }
       } else {
-        yieldingTo = createTask(next.value, { resourceScope: options.resourceScope || task, ignoreError: true });
+        yieldingTo = createTask(next.value, { scope: options.resourceTask || task, ignoreError: true });
         yieldingTo.consume(trap);
         yieldingTo.start();
         options.onYieldingToChange && options.onYieldingToChange(yieldingTo);
--- a/packages/core/src/controller/iterator-controller.ts
+++ b/packages/core/src/controller/iterator-controller.ts
@@ -3,7 +3,6 @@ import { OperationIterator } from '../operation';
 import { createTask, Task } from '../task';
 import { Operation } from '../operation';
 import { createFuture, Value } from '../future';
-import { createRunLoop } from '../run-loop';
 
 const claimed = Symbol.for('effection/v2/iterator-controller/claimed');
 
@@ -13,12 +12,11 @@ interface Claimable {
   [claimed]?: boolean;
 }
 
-export function createIteratorController<TOut>(task: Task<TOut>, iterator: OperationIterator<TOut> & Claimable, options: Options = {}): Controller<TOut> {
+export function createIteratorController<TOut>(task: Task<TOut>, iterator: OperationIterator<TOut> & Claimable, options: Options): Controller<TOut> {
   let didHalt = false;
   let yieldingTo: Task | undefined;
 
   let { produce, future } = createFuture<TOut>();
-  let runLoop = createRunLoop();
 
   function start() {
     if (iterator[claimed]) {
@@ -32,7 +30,7 @@ export function createIteratorController<TOut>(task: Task<TOut>, iterator: Opera
   }
 
   function resume(iter: NextFn) {
-    runLoop.run(() => {
+    options.runLoop.run(() => {
       let next;
       try {
         next = iter();
--- a/packages/core/src/controller/iterator-controller.ts
+++ b/packages/core/src/controller/iterator-controller.ts
@@ -45,7 +45,7 @@ export function createIteratorController<TOut>(task: Task<TOut>, iterator: Opera
           produce({ state: 'completed', value: next.value });
         }
       } else {
-        yieldingTo = createTask(next.value, { scope: options.resourceTask || task, ignoreError: true });
+        yieldingTo = createTask(next.value, { scope: task.options.yieldScope || task, ignoreError: true });
         yieldingTo.consume(trap);
         yieldingTo.start();
         options.onYieldingToChange && options.onYieldingToChange(yieldingTo);
--- a/packages/core/src/controller/resource-controller.ts
+++ b/packages/core/src/controller/resource-controller.ts
@@ -6,6 +6,7 @@ import { createFuture } from '../future';
 
 export function createResourceController<TOut>(task: Task<TOut>, resource: Resource<TOut>, options: Options): Controller<TOut> {
   let delegate: Controller<TOut>;
+  let resourceTask: Task;
   let { scope } = task.options;
   let { produce, future } = createFuture<TOut>();
 
@@ -13,16 +14,19 @@ export function createResourceController<TOut>(task: Task<TOut>, resource: Resou
     if(!scope) {
       throw new Error('cannot spawn resource in task which has no resource scope');
     }
-    let init;
+
+    let name = resource.name || resource.labels?.name || 'resource';
+
+    resourceTask = scope.run(undefined, { labels: { name, type: 'resource' } });
+
+    let iterator;
     try {
-      init = resource.init(scope, task);
+      iterator = resource.init(resourceTask, task);
     } catch(error) {
       produce({ state: 'errored', error });
       return;
     }
-    let name = resource.name || resource.labels?.name || 'resource';
-    let resourceTask = scope.run(undefined, { labels: { name, type: 'resource' } });
-    delegate = createIteratorController(task, init, { resourceTask, runLoop: options.runLoop });
+    delegate = createIteratorController(task, iterator, { resourceTask, runLoop: options.runLoop });
     delegate.future.consume((value) => {
       produce(value);
     });
@@ -33,5 +37,14 @@ export function createResourceController<TOut>(task: Task<TOut>, resource: Resou
     delegate.halt();
   }
 
-  return { start, halt, future, type: 'resource', operation: resource };
+  return {
+    type: 'resource',
+    start,
+    halt,
+    future,
+    get resourceTask() {
+      return resourceTask;
+    },
+    operation: resource
+  };
 }
--- a/packages/core/src/controller/resource-controller.ts
+++ b/packages/core/src/controller/resource-controller.ts
@@ -1,4 +1,3 @@
-import { createIteratorController } from './iterator-controller';
 import { Controller } from './controller';
 import { Resource } from '../operation';
 import { Task } from '../task';
--- a/packages/core/src/controller/resource-controller.ts
+++ b/packages/core/src/controller/resource-controller.ts
@@ -1,10 +1,10 @@
-import { Controller } from './controller';
+import { Controller, Options } from './controller';
 import { createIteratorController } from './iterator-controller';
 import { Resource } from '../operation';
 import { Task } from '../task';
 import { createFuture } from '../future';
 
-export function createResourceController<TOut>(task: Task<TOut>, resource: Resource<TOut>): Controller<TOut> {
+export function createResourceController<TOut>(task: Task<TOut>, resource: Resource<TOut>, options: Options): Controller<TOut> {
   let delegate: Controller<TOut>;
   let { resourceScope } = task.options;
   let { produce, future } = createFuture<TOut>();
@@ -20,7 +20,7 @@ export function createResourceController<TOut>(task: Task<TOut>, resource: Resou
       produce({ state: 'errored', error });
       return;
     }
-    delegate = createIteratorController(task, init, { resourceScope });
+    delegate = createIteratorController(task, init, { resourceScope, runLoop: options.runLoop });
     delegate.future.consume((value) => {
       produce(value);
     });
--- a/packages/core/src/controller/resource-controller.ts
+++ b/packages/core/src/controller/resource-controller.ts
@@ -6,21 +6,23 @@ import { createFuture } from '../future';
 
 export function createResourceController<TOut>(task: Task<TOut>, resource: Resource<TOut>, options: Options): Controller<TOut> {
   let delegate: Controller<TOut>;
-  let { resourceScope } = task.options;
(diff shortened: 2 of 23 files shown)


cowboyd:
This should allow us to avoid the double-spawn problem when using spawn itself. If we have this, then I wonder if we could use it for a, or an idea I had that if we could specify resources as _two_ operations, both of which run inside a scope that is created for the resource. The "init" returns the value of the resource (as it does today) but then the "body" accepts the value and then runs potentially forever. The type sig would be roughly:

```ts
interface Resource<T> {
  init: Operation<T>;
  operation: (value: T) => Operation<void>;
}
```

This would be how you could write the server code that's clean to compose, but also easy to get right about defining your core group.


```ts
export function createServer(): Resource<Server> {
  return {
    *init() {
      let server = http.createServer();
      server.listen();
      if (!server.listening) {
        let error = yield race([
          once(server, 'error'),
          once(server, 'listening')
        ]);
        if (error) {
          throw error;
        }
      }

      yield spawn(function* listenForServerErrors() {
        let error = yield once(server.http, 'error');
        throw error;
      });

      return {
        http,
        address: server.address() as AddressInfo
      };
    },
    operation: server => on(server, 'request').forEach(function*(request) {
      /// body of server
    }
  }
}
```

It occurs to me now, that if we had some way to compose operations within a single task, then we could do this without any special options. Let's say there was an `enqueue()` method that added an operation to a task. This would be almost equivalent of the `then()` method in Promise, except that it would allow you to chain multiple operations together inside the same task programatically.

```js
task.run(resource.init || Future.resolve()).enqueue(value => resource.operation(value));
```

This would let you write spawn as a simple resource again:

```ts
// spawn is just a resource with no `init` operation
export function spawn<T>(operation: Operation<T>): Operation<Task<T>> {
  return { operation };
}
```

jnicklas:
@cowboyd there are a few reasons I'm not a fan of splitting up `init` like this:

1. It is often the case that we want to have access to shared data between what happens in the `init` and what happens in the long running server. For example, imagine we are consuming a resource in `init` and want to use its value in the background. There is no real sensible way of "transferring" a value from `init`, especially a resource, to `operation`.
2. It restricts composition. Imagine in the example that you posted above that we would want to create an abstraction around http server and use it in your resource. How would this work? There isn't a good way to extend the lifetime of the http server that was created within `init` to the larger scope of `operation`. And like we mentioned last time, the idea of "moving" a resource or task somewhere else in the tree is unappealing, not least because of implementation concerns.
3. It reintroduces the really annoying problem we had in v1 where the resource task terminates prematurely.

> Let's say there was an enqueue() method that added an operation to a task

I explored the idea of running operations "inline" inside of a task when working on v2. We had some idea of a task having a list of operations that it would iterate, and having the option to append operations to a task. I think this is roughly what you are suggesting here, right? Unfortunately this just fundamentally doesn't really mesh with the idea of having a task resolve to a single value. Otherwise `enqueue` is just the same as `run`, which is fine, but that still requires us to get a hold of the outer scope to run the operation. Otherwise we'll just be restricted to the scope of the yield point, which doesn't help us.

If you just see `enqueue` as a higher level operation, sort of like `then` on promise, then that might be a worthwhile addition, I've certainly wanted this from time to time (though quite rarely), but I don't think that in itself is powerful enough to solve this problem here.

cowboyd:
> It is often the case that we want to have access to shared data between what happens in the init and what happens in the long running server. For example, imagine we are consuming a resource in init and want to use its value in the background. There is no real sensible way of "transferring" a value from init, especially a resource, to operation.

The argument to operation is the value returned by `init`. Any value generate in init can be passed as part of that value.

> It restricts composition. Imagine in the example that you posted above that we would want to create an abstraction around http server and use it in your resource. How would this work? There isn't a good way to extend the lifetime of the http server that was created within init to the larger scope of operation.

This idea would depend on both operations running within the same task, so  I would imagine that composition works the same as it does today:

```js
yield createServer()
```

And of course, the server operation would be more correctly expressed as:

```ts
operation: server => function*() {
try {
  //server body
} finally {
  server.close();
}
```


It is a really good point about breaking the 1:1 nature of `Task`:`Operation` and I don't see that being a change free of unforeseen consequences. Still, it makes me feel like perhaps scope and task really are two different things.

cowboyd:
> do we provide access to the resource task to a consumer of the resource? and if so, how?

@jnicklas I think it's fair to say that any interface to the resource task should be through the resource value itself. That seems to optimize for encapsulation, but still allows for poking holes in the encapsulation where needed. I mean, trivially, the resource could define its own `task` attribute to expose it fully.

packages/core/src/operations/spawn.ts
```diff
@@ -1,18 +1,24 @@
-import { Operation, Resource } from '../operation';
+import { Operation, OperationFunction } from '../operation';
+import { createFuture } from '../future';
 import type { Task, TaskOptions } from '../task';
 
-interface Spawn<T> extends Resource<Task<T>> {
+interface Spawn<T> extends OperationFunction<Task<T>> {
```
cowboyd, line 5: Is there a reason not to just make this an `Operation<Task<T>>` Just to keep the interface as thin as possible and so that external code is only ever thinking of spawn as an opaque operation?

cowboyd (commented):
This feels like a really low impact way of solving the problem, and if I squint, it looks like it's a bayb step towards having an explicit scope that is apart from task, which we can potentially explore in a `v3`

Upon reflection, I don't like having a separate operation for the resource body either, and I don't think we need to consider that given given that now all spawns happen in the resource scope and not in the parent.

packages/core/src/controller/resource-controller.ts
```diff
@@ -6,21 +6,23 @@ import { createFuture } from '../future';
...
-      init = resource.init(resourceScope, task);
+      init = resource.init(scope, task);
     } catch(error) {
       produce({ state: 'errored', error });
       return;
     }
-    delegate = createIteratorController(task, init, { resourceScope });
+    let name = resource.name || resource.labels?.name || 'resource';
+    let resourceTask = scope.run(undefined, { labels: { name, type: 'resource' } });
+    delegate = createIteratorController(task, init, { resourceTask });
```
cowboyd: I know this is the way it was, but why must the `init()` function be a generator? Can we not just use `createController()` here?

packages/core/src/future.ts
```diff
@@ -27,10 +27,9 @@ export interface NewFuture<T> {
   resolve(value: Value<T>): void;
 }
 
-export function createFuture<T>(): NewFuture<T> {
+export function createFuture<T>(result?: Value<T>): NewFuture<T> {
   let runLoop = createRunLoop();
   let consumers: Consumer<T>[] = [];
-  let result: Value<T>;
```
cowboyd: What about following the promise convension and having a `Future.resolve()` method.

packages/core/src/operations/ensure.ts
```diff
@@ -1,22 +1,23 @@
...
+        yield;
+      } finally {
+        let result = fn();
+        if(result) {
+          yield result;
+        }
+      }
+    }, { labels: { name: 'ensureHandler' } });
+    return createFuture({ state: 'completed', value: undefined }).future;
+  }, { name: 'ensure' });
```
cowboyd: We can avoid all the labelling just by using functions with names, right?

packages/core/src/operations/spawn.ts
jnicklas: I'm thinking that we're going to simplify this and turn it into a regular operation once we remove `within`. I had some issue with the typing here (maybe this was in an intermediate state though) that prompted me to change this to `OperationFunction`.

packages/core/src/controller/resource-controller.ts
jnicklas: I think we probably could, though it does introduce somewhat of a circular dependency. I think the intention originally was to keep it simple, and only allow generators with an eye towards possibly generalizing this later on. That being said, I don't think there is a particularly strong use case for using something other than a generator.

packages/core/src/future.ts
jnicklas: That would make sense too, I don't have a strong opinion on this either way. I was a bit unsure of how to name such a thing, so I just ended up picking this solution since it seemed to be less of a hassle.

packages/core/src/operations/ensure.ts
jnicklas: Yes, I think so, good point!

jnicklas:
> The argument to operation is the value returned by init. Any value generate in init can be passed as part of that value.

I see you've changed your mind on this, but for posterity's sake: this works fine with regular values, but becomes significantly trickier when the value is also a resource.

packages/core/src/future.ts
cowboyd: Yeah, it's just that `Future` and `createFuture` are part of our public API, and so as we converge on 2.0 stable, we need to be extra careful. I think that `Future.resolve(x)` is very straightforward and seems 99.9% likely to be part of the public API going forward, but `createFuture({type: 'completed', value: x }).future` seems like it will be rarely used except internally, takes up the real estate of the a potential argument to `createFuture()` (no idea what that might be)

packages/core/src/operations/spawn.ts
cowboyd: 👍🏻 on removing `within()` and I guess the question is, is it a breaking change to narrow the type further on?  I think the answer is no. So if it is something that has to wait until then, I'm all for it.

packages/core/src/controller/resource-controller.ts
cowboyd: I agree that it will be rare to write an init function with something other than with a generator , but I think the strongest reason to make it more general is to reduce systemic complexity. That way, we can say "yes, `init()` is magical, but only in this very specific way"

jnicklas: That's a good point. I'd rather tackle this in a separate PR though.

packages/core/src/future.ts
jnicklas: We could also add this as a low level API and add helper functions such as `createResolvedFuture` or something? Not sure how I feel about `Future.resolve` as an API, it matches the promise API, but not any other Effection APIs.

packages/core/src/operations/spawn.ts
jnicklas: I've tried to use just plain `Operation` here, but it indeed does not work. Since we're extending an interface here and some types in the Operation union cannot be extended, we have to be more specific.

packages/core/src/future.ts
jnicklas: I've removed this part of the PR and create the resolved futures manually and added an issue for this, so as not to hold up this PR with this bike shed: https://github.com/thefrontside/effection/issues/495

jnicklas:
I've tinkered with this a bit, and I think it's now in a good state to be considered.

I've added a way of accessing the resource task by going through the init task, that is the init task has a handle to the resource task that it creates in the background. This isn't super convenient, but I think accessing this task is rare and should probably mostly be done for debugging purposes.

packages/core/src/controller/iterator-controller.ts
```diff
@@ -47,7 +47,7 @@ export function createIteratorController<TOut>(task: Task<TOut>, iterator: Opera
           produce({ state: 'completed', value: next.value });
         }
       } else {
-        yieldingTo = createTask(next.value, { resourceScope: options.resourceScope || task, ignoreError: true });
+        yieldingTo = createTask(next.value, { scope: options.resourceTask || task, ignoreError: true });
```
jnicklas: One interesting thing is that this is actually the *only* spot in `iterator-controller` where `task` is used, so we could also just pass in the `resourceTask` as `task`, instead of adding this option. Not sure how I feel about that though.

packages/core/src/controller/resource-controller.ts
```diff
@@ -6,21 +6,27 @@ import { createFuture } from '../future';
...
-  let { resourceScope } = task.options;
+  let resourceTask: Task;
+  let { scope } = task.options;
   let { produce, future } = createFuture<TOut>();
 
   function start() {
-    if(!resourceScope) {
+    if(!scope) {
       throw new Error('cannot spawn resource in task which has no resource scope');
     }
```
jnicklas, line 15: There is another option here of spawning the resource within `task` instead of `scope`. This is pretty much only relevant when using a resource as a root task, which pretty much makes no sense, since the resource will terminate as soon as `init` is done, so I'm not sure that it really matters.

jnicklas:
One possible problem with this implementation is that the resourceTask and the initTask are completely independent of each other. That means that when the initTask fails for some reason, the resourceTask will survive. It seems intuitive that a crash in the initTask would also take down the resourceTask, I think. @cowboyd do you have any thoughts on this?

cowboyd (approved):

PR #482, opened 2021-08-13
Copy down README into effection package explicitly

cowboyd:
## Motivation
It appears that NPM pack script does not resolve the symlinks of explicitly included files [for reasons of ego][ego]

## Approach
This copies it down from the root on the prepack step. A little bit of extra work on each internal build, but considering that most building happens via typescript references, that's not as big a problem

[ego]: https://github.com/npm/npm/issues/3310#issuecomment-15904722


minkimcello (approved):
i tried symlink-ing locally (for the previous pr) and the markdown preview on vscode was just showing the path as text. i figured it would work as intended once published. 😅

PR #486, opened 2021-08-16
Use Object.create to create error object with trace

### LINKED ISSUE #485: Error decoration fails in Firefox
When implementing https://github.com/thefrontside/effection/pull/347, we decided on not wrapping the error, and instead creating a sort of new error object, with the same prototype and transferring all properties from the original error. The implementation of this looks like this:

https://github.com/thefrontside/effection/blob/8d827f26fbec159d5a56c6f9897131566d650194/packages/core/src/error.ts#L7-L21

Unfortunately this does not work as expected on Firefox:

Screenshot 2021-08-16 at 10 35 25

Here we can see that the `stack` property is not transferred to the new error object. Presumably there is some C-implementation under the hood which does some special magic, and Firefox simply does not expect such weird shenanigans.

It seems like we need to go back to the drawing board on how to store effection traces.

jnicklas:
Closes #485 

Unfortunately copying the properties to the error objects as we did before does not work in all JS engines, notably Firefox, see #485

This option gives us a less clean prototype chain, but the properties do seem to at least be correctly available on all JS engines.

Another option for implementing this would be using a proxy object, but I don't think there is any benefit to doing so here.

### DIFF
--- a/.changes/error-object-create.md
+++ b/.changes/error-object-create.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": "minor"
+---
+
+Use Object.create to wrap error objects rather than copying properties
--- a/packages/core/src/error.ts
+++ b/packages/core/src/error.ts
@@ -12,10 +12,10 @@ export function addTrace(error: Error & Partial<HasEffectionTrace>, task: Task):
     state: task.state,
   };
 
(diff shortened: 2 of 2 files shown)


taras (approved):
👍

PR #491, opened 2021-08-23
Add resolve/reject/halt functions when creating future

jnicklas:
This makes future more ergonomic to use, since the `produce` function is a little awkward, especially when used from plain-JS.

### DIFF
--- a/packages/core/src/future.ts
+++ b/packages/core/src/future.ts
@@ -23,8 +23,9 @@ export interface Future<T> extends Promise<T>, FutureLike<T> {
 export interface NewFuture<T> {
   future: Future<T>;
   produce(value: Value<T>): void;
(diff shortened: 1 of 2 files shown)


packages/core/src/future.ts
```diff
@@ -23,8 +23,9 @@ export interface Future<T> extends Promise<T>, FutureLike<T> {
 export interface NewFuture<T> {
   future: Future<T>;
   produce(value: Value<T>): void;
-  /** @deprecated Use produce(value) instead */
```
dagda1, line 26: It might be an idea to keep the `/** @deprecated Use produce(value) instead */`.  It tells the language server to put a strike through into language server aware IDEs

Screenshot 2021-08-24 at 14 30 34

cowboyd (approved):

PR #497, opened 2021-08-27
Enable side effects in inspect-ui package

jnicklas:
Otherwise when packages using `yarn prepack`, everything in `index.tsx` will be tree-shaken away, and the inspector will just be a blank page.


jnicklas:
I also added `optimize: false` since there is no need to optimize size for the inspector and the resulting output is easier to debug.

cowboyd (approved):

PR #498, opened 2021-08-27
Name event streams

jnicklas:
Allows event streams created through on/onEmit to be named


cowboyd (approved):
👍🏻 It also occurred to me that at some point we may want to also support labelling for streams.

PR #506, opened 2021-08-31
Add toString output for tasks

jnicklas:
This simplifies debugging when using the inspector would be overkill. It just shows a nicely formatted ASCII rendering of the task structure.

Screenshot 2021-08-31 at 16 25 08

### DIFF
--- a/.changes/task-to-string.md
+++ b/.changes/task-to-string.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": "minor"
+---
+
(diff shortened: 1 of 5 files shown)


cowboyd (approved):
Looks amazing 😍

I'm assuming that chalk will work outside nodejs?

jnicklas:
This was accidentally pushed as part of #480, unfortunately, chalk does *not* work in the browser, see #510. Let's find some way of colorizing this later.

PR #510, opened 2021-09-02
Remove chalk dependency

jnicklas:
This commit was actually not meant to be pushed, and was accidentally pushed as part of #480. Using chalk breaks
in the browser.


cowboyd (approved):
Looks good, but I think we need a changeset

jnicklas:
The change that introduced this hasn't been released, so there's no need for a changeset, I think.

PR #511, opened 2021-09-06
Add labels to fetch package

### LINKED ISSUE #425: add labels to @effection/fetch

jnicklas:
Closes #425

### DIFF
--- a/packages/fetch/src/fetch.ts
+++ b/packages/fetch/src/fetch.ts
@@ -1,4 +1,4 @@
-import { spawn, Operation, Resource } from '@effection/core';
+import { ensure, withLabels, Operation, Resource } from '@effection/core';
 import { fetch as nativeFetch } from 'cross-fetch';
 import { AbortController } from 'abort-controller';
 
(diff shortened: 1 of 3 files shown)


packages/fetch/src/fetch.ts
```diff
@@ -29,26 +23,33 @@ export function fetch(info: RequestInfo, requestInit: RequestInit = {}): Fetch {
   }
 
   return {
+    name: `fetch \`${info}\``,
+    labels: {
+      expand: false
+    },
```
cowboyd: What about representing some of the `requestInit` inside the labels? It seems to me that parameters like `method` are extremely handy to have.

jnicklas: Good idea!

jnicklas:
I added the request method in the labels and changed the naming scheme a bit, thinking we might adopt this naming scheme in other places (such as @effection/process) as well?

cowboyd (approved):
I like that naming scheme

PR #515, opened 2021-09-07
Add process guide

jnicklas:
A new guide for the @effection/process package

### DIFF
--- a/website/docs/guides/processes.md
+++ b/website/docs/guides/processes.md
@@ -114,21 +114,35 @@ want a process to live for the entire duration of the program, or in the case
 of effection for the entire duration of an operation. When using `exec`,
 effection ensures that the process cannot live *longer* than the operation, but
 it does not ensure that the process cannot live *shorter* than the operation.
+
+An example of this might be spawning a server which must keep running so that
+we can send requests to it. If the server exists for any reason, even if it
+exits successfully with an exit code of `0`, that is an unexpected condition,
+and we need to throw an error.
+
 This is where `daemon` comes in. When using `daemon`, if the process finishes
 before the operation is complete, even if it finishes with an exit code of `0`,
 an error is thrown.
 
+Here is an example of using `daemon` to spawn a server process in the
+background:
+
 ``` typescript
-import { main, spawn } from 'effection';
+import { main, spawn, fetch } from 'effection';
 import { daemon } from '@effection/process';
 
 main(function*() {
-  let myProcess = yield daemon('npm start');
+  let myProcess = yield daemon('my-server --port 3000');
 
-  yield spawn(myProcess.stdout.forEach((text) => console.log(text)));
+  yield myProcess.stdout.filter((chunk) => chunk.includes("ready")).expect();
 
-  yield;
+  let result = yield fetch('http://localhost:3000').json();
+
+  console.log('got result:', result);
 });
 ```
+
+Since we're sending a request to the server using `fetch`, the server must be
+kept running.
 
 [resource]: /docs/guides/resources
--- a/website/docs/guides/processes.md
+++ b/website/docs/guides/processes.md
@@ -3,4 +3,132 @@ id: processes
 title: Spawning processes
 ---
 
-*Coming soon!*
+Effection simplifies the process of managing any kind of resource, and we've
+taken great care to provide a good experience when you want to spawn external
+processes.
+
+Processes are managed via the separate `@effection/process` package, which
+currently only works in node.
+
+## Exec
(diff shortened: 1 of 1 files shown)


website/docs/guides/processes.md
```diff
@@ -3,4 +3,132 @@ id: processes
...
+
+`exec` returns a [resource][] which means that if you want to create an object
+which uses an external process internally, you will have to create it as a
+resource as well.
+
+`exec` gives you a lot of control, including the ability to work with the
+input and output streams of an external process, and to block and wait for
+the process to complete using `myProcess.join()`.
+
+Often we expect the process to complete orderly with an exit code of `0`
```
cowboyd:
```suggestion
Often we expect the process to complete in an orderly fashion with an exit code of `0`
```

website/docs/guides/processes.md
```diff
@@ -3,4 +3,132 @@ id: processes
...
+## Daemon
+
+A common problem with processes is what to do if they exit too early. Often we
+want a process to live for the entire duration of the program, or in the case
+of effection for the entire duration of an operation. When using `exec`,
+effection ensures that the process cannot live *longer* than the operation, but
+it does not ensure that the process cannot live *shorter* than the operation.
+This is where `daemon` comes in. When using `daemon`, if the process finishes
+before the operation is complete, even if it finishes with an exit code of `0`,
+an error is thrown.
```
cowboyd: I think an example here would help illustrate what is a pretty subtle point. Maybe something like:

If you have an operation that depends on a file server running in order to serve static content. If that file server resource _ever_ goes away, even if it shuts down gracefully, then that means your operation can no longer do its job and so it must fail.... or something similar

cowboyd (commented):
This is a nice simple treatment for another absolutely critical component. 

Aside from the inline comments, I do feel like we need to specifically spend just a little more time talking about how to handle `stdin` and `stdout` as streams with links to the stream documentation since it is almost always a requirement, and it's almost always a sticking point that we have had to stop and explain in the past.

website/docs/guides/processes.md
jnicklas: Great point, I've updated this section a bit to try and make it clearer.

website/docs/guides/processes.md
```diff
@@ -3,4 +3,146 @@ id: processes
...
+  yield myProcess.stdout.filter((chunk) => chunk.includes("ready")).expect();
+
+  let result = yield fetch('http://localhost:3000').json();
+
+  console.log('got result:', result);
+});
+```
+
+Since we're sending a request to the server using `fetch`, the server must be
+kept running.
```
cowboyd:
```suggestion
Since we're sending a request to the server using `fetch`, the server must be
kept running. If it ever stops, then the operation will not be able to do its job, and so it will fail immediately.
```

website/docs/guides/processes.md
```diff
@@ -3,4 +3,146 @@ id: processes
...
+resource as well.
+
+`exec` gives you a lot of control, including the ability to work with the
+input and output streams of an external process, and to block and wait for
+the process to complete using `myProcess.join()`.
+
+Often we expect the process to complete in an orderly fashion with an exit code of `0`
+and we don't want to have to add any error handling in case the process
+exits with a non-zero error code. `expect()` is a convenient operation
+which blocks just like `join()` but throws on any non-zero error code.
```
cowboyd:
```suggestion
which blocks just like `join()` but throws an error on any non-zero exit code.
```

website/docs/guides/processes.md
```diff
@@ -3,4 +3,146 @@ id: processes
...
+main(function*() {
+  let myProcess = yield exec('npm start');
+
+  yield spawn(myProcess.stdout.forEach((text) => console.log(text)));
+
+  let result = yield myProcess.join();
+  console.log(`terminated with exit code ${result.code}`);
+});
+```
+
```
cowboyd:
```suggestion
> Note: every process has `stdout` and `stderr`  properties which can be consumed as [effection streams](./collections)
```

jbolda:
Can we add a link in this guide to the `@effection/process` package / API?

cowboyd (approved):

cowboyd:
> Can we add a link in this guide to the `@effection/process` package / API?

This does seem like a good idea. Although, our API docs, while publishing are not well filled out. Still, incomplete information is better than none.

jnicklas:
Merging this now as is. Let's fill out the section on streams once we have made the necessary improvements to them.

PR #524, opened 2021-09-08
Remove node package

jnicklas:
This removes the deprecated `@effection/node` package from the codebase. Note that existing versions of this package are still available via npm, we just won't publish any NEW versions of it.


cowboyd (commented):
Do we need to remove it from covector?

cowboyd (approved):
> Do we need to remove it from covector?

🤦🏻‍♂️

Issue #502, opened 2021-08-27
Resources created through higher order operations terminate too early.


jnicklas:
Suppose we have some operations which creates two resources, both of which take some time to `init`:

``` typescript
function frob*() {
  let foo = yield resourceFoo();
  let bar = yield resourceBar();
  // do something
}
```

If we want to parallelize the intialization of these resources, the obvious way is to use `all`:

``` typescript
function frob*() {
  let [foo, bar] = yield all([resourceFoo(), resourceBar()]);
  // do something
}
```

However, this currently does not work. `all` only works on regular operations, when used with a resource the resource gets terminated as soon as `all` is done. The same is true for other "higher-order operations", that is operations which take another operation as an argument, such as `withTimeout`. Some special care needs to be taken in these operations to handle resources properly.

When #480 is implemented, these operations will be able to set the scope of any tasks they spawn to their own scope, which effectively "inlines" any resources. So the implementation is simple. It's worth keeping in mind that *all* higher-order operations will need to decide on how to handle resources, even those defined in user code.

## Error handling

Some special care needs to also be taken about error handling for resources. The exact semantics of this are non-obvious. If the resource fails "in the background", that is for example a task it spawns fails, then this is handled as normal. The interesting case is what happens when `init` fails. For something like `withTimeout`, we'd probably want to terminate the resource task here as well, but the same is not necessarily true for `any`. There is probably no blanket solution which works the same for all higher-order operations.

PR #527, opened 2021-09-09
Enable resource support for higher order operations

jnicklas:
See #502 

This makes the `all`, `race` and `withTimeout` operations work properly with resources

Depends on #533

### DIFF
--- a/.changes/higher-order-operations.md
+++ b/.changes/higher-order-operations.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": "minor"
+---
+
+Enable support for resources in higher order operations `all`, `race` and `withTimeout`.
--- a/packages/core/src/operations/all.ts
+++ b/packages/core/src/operations/all.ts
@@ -2,22 +2,30 @@
 import type { Operation } from '../operation';
 import type { Task } from '../task';
 import { withLabels } from '../labels';
+import { spawn } from './spawn';
 
 type All<T extends Operation<any>[]> = {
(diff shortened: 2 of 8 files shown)


jnicklas:
@cowboyd I'd like to get your opinion on this implementation, I'm not super enthusiastic about manually halting the resource tasks in case of failure, but I had a tough time coming up with another implementation which allows us to let resources "escape" the `all`, without also letting regular operations escape without this manual halting.

cowboyd:
What about making `all` itself a resource? This would make the implementation really simple, no?:

```ts
type All<T extends Operation<any>[]> = {
  [P in keyof T]: T[P] extends Operation<infer TArg> ? TArg : never;
}

export function all<T extends Operation<any>[]>(operations: T): Resource<All<T>> {
  return {
    labels: { name: `all()`, count: operations.length },
    *init() {
      let tasks: Task[] = [];
      for (let operation of operations) {
        tasks.push(yield spawn(operation));
      }
      return yield Promise.all(tasks);
    }
  };
}
```

Cons to this approach:

1. all resources are spawned under the all resource task (although is this a bad thing?)
2. This implementation won't complete synchronously when able, so we would need to implement a `Future.all()` equivalent if we wanted that.

jnicklas:
@cowboyd I did consider that option, the downside is that regular operations spawned within the resource live "too long" that is they are not constrained to the "all", they'll be spawned in the parent. We could, of course, manually halt them, but then we're back to manually halting stuff, and I feel that only resources living too long is better than everything living to long.

cowboyd (approved):

PR #532, opened 2021-09-10
Add errors guide

### LINKED ISSUE #493: How do you recover from errors in resources?
We run into this situation in simulacrum where we are starting multiple servers, and allowing the port to be configured, but if multiple processes select the same port, then the server resource fails, and there isn't a really good way to recover from it, so the entire server ends up crashing.

How would you even recover from this? I can think of a couple of options. One way would be to have some operation that funnelled (and allowed you to catch) all errors coming to a task from itself:

```js
import { run, self } from 'effection';

run(function*() {
  yield createResourceOne();
  yield createResourceTwo();
  
  try {  
    yield self;
  } catch (error) {
    console.log('error occured');
  }
});

```

Another might be to establish a "trap" for specific tasks:

``` javascript
import { run, trap } from 'effectcion'

run(function*() {
  let one = yield createResourceOne();
  let two = yield createResourceTwo();

  try {
    yield trap(one, two);
  } catch (error) {
    console.log('error occured');
  }
});
```

That would allow handling specific errors from one resource but not the other by just not saying that you want to trap it.

jnicklas:
Add a guide for error handling within effection.

Closes #493


website/docs/guides/errors.md
```diff
@@ -0,0 +1,190 @@
...
+You might be surprised that we do *not* enter the catch handler here! Instead,
+our entire main task just fails. This is by design! We are only allowed to
+catch errors thrown by whatever we yield to directly, not by any spawned
+children or resources running in the background. This makes error handling more
+predictable, since our catch block will not receive errors from any background
+task, we're better able to specify which errors we actually want to deal with.
+
+## Error boundary
+
+If we do want to catch an error from a spawned task (or from a [Resource][]) then
```
cowboyd, line 133: I never knew you could link like this 🤯

website/docs/guides/errors.md
```diff
@@ -0,0 +1,190 @@
...
+    }
+  } catch(err) {
+    console.log("it blew up:", err.message);
+  }
+});
+```
+
+## Main error
+
+We have used the `main` entry point to run our code to run our code, when we build
```
cowboyd: Slight typo adjustment and re-wording suggestion.

```suggestion
We have already seen using the `main` entry point to run our code when we build
```

Also, since we're referencing having seen this feels like we should link to the guide that introduces `main`

website/docs/guides/errors.md
```diff
@@ -0,0 +1,190 @@
...
+    let content = yield readFile(fileName);
+    console.log(content.reverse().toString());
+  } catch(err) {
+    throw new MainError({ message: `no such file ${fileName}`, exitCode: 200 });
+  }
+});
+```
+
+When using `MainError` the stack trace will not be printed and the exit code
+specified in the error will be used.
```
cowboyd: I think an example of the difference will really cement the concept. What about showing this example code running both with and without a `MainError`

website/docs/guides/errors.md
```diff
@@ -0,0 +1,190 @@
...
+we need to introduce an intermediate task which allows us to bring the error into
+the foreground. We call this pattern an "error boundary":
+
+``` typescript
+import { main } from 'effection';
+import { tickingBomb } from './ticking-bomb';
+
+main(function*() {
+  try {
+    yield function*() { // error boundary
```
cowboyd, line 143: Take this one with a grain of salt because I'm not sure how I feel about it myself. The example as-is is very clear, but it is also a good habit to get into to name your error boundary so that it shows up in the inspector.

```suggestion
    yield function* errorBoundary() {
```

cowboyd (approved):
Excellently written, and absolutely critical information to know when writing Effection code

website/docs/guides/errors.md
jnicklas: I see your point, but I almost feel that this would be confusing, both in the example, and showing up in the inspector. Maybe a more neutral name like `inner` or something like that would make sense, but if I saw this in the inspector I would be confused why everything is under an error boundary.

PR #537, opened 2021-09-21
Catch errors thrown during resource init

### LINKED ISSUE #535: Cannot catch errors raised from inside Resource.init() operation
If an exception happens in the `init()` operation of a resource, then
that exception cannot be caught from within the operation that is
calling for the initialization. In otherwords, since `init()` is being
yielded to, the intuition is that you can catch errors happening
there. However, the following script fails with a `catch me if you can`

``` typescript
import { main } from 'effection';

main(function*() {
  try {
    yield {
      *init() {
        throw new Error('catch me if you can');
      }
    }
  } catch (error) {
    console.log('caught error');
  }
})
```

Unfortunately, the workaround is very messy because putting an error
boundary around the resource menas that it is now spawned in a
short-lived operation.

jnicklas:
Closes #535

This adjusts the propagation of errors for resources so that errors thrown during `init` can be cought. Previously, an error thrown during `init` would propagate to the resource task, causing the resource task to fail. This would have the undesirable side-effect that even if the `init` error was caught (for example to retry), then the outer task would still fail, due to the failure in the resource task.

Instead, a failure in `init` now causes the resource task to become *halted* rather than *errored*, sort of as though it were a sibling, while still maintaining the existing hierarchy of the `init` task being a child.

This means that just as before, a failure in `init` will cause the resource to be shut down, but this failure will not propagate out to the outer task. Since the failure will always be thrown from the *resource constructor*, the failure will still propagate, but we have more control, since we can catch the error from the resource constructor if it is yielded to.

### DIFF
--- a/.changes/catch-init-error.md
+++ b/.changes/catch-init-error.md
@@ -1,0 +1,5 @@
+---
+"@effection/core": "minor"
+---
+
+Adjust the propagation of errors for resources to make it possible to catch errors from `init`
--- a/packages/core/src/controller/resource-controller.ts
+++ b/packages/core/src/controller/resource-controller.ts
@@ -21,13 +21,19 @@ export function createResourceController<TOut>(task: Task<TOut>, resource: Resou
 
     initTask = resourceTask.run((task) => resource.init(resourceTask, task), {
       yieldScope: resourceTask,
-      labels: { name: 'init' }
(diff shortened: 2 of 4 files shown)


cowboyd (approved):
🎊

PR #540, opened 2021-09-22
Improve process streams

jnicklas:
This makes a number of pretty significant and breaking changes to process streams, in the hopes of making them more useful, more consistent and easier to understand.

While we have found a pretty intuitive API for working with the processes themselves, their stdout and stderr streams have been a bit neglected. We are already foreseeing some things that we want to do in bigtest that are difficult to do with the current API.

First of all we create a solid foundation by making the std streams over Buffers. This is what node exposes by default, and since we would pay a performance penalty in cases where we unncessarily convert to strings, we should just expose this.

However, working with raw buffers is tedious, so we expose two smart mapping functions on iostreams `text()` and `lines()`. `text()` just converts the buffers into strings. `lines` is a bit fancier and creates a new stream which emits once for each printed line (waiting until the line is complete).

Additionally we improve the `buffer` function on `Stream` by making it return a resource, and also taking an optional limit, so we can create a bounded buffer using a RingBuffer behind the scenes.

In bigtest we have previously wanted to cache the previous X number of lines. With this change, something like the following becomes possible, which was previously pretty challenging:

``` typescript
let app = yield exec(options.appCommand);
let lines = yield app.stdout.lines.buffer(1000);

yield lines.filter((l) => l.includes('ready')).expect();
```

### DIFF
--- a/.changes/process-std.md
+++ b/.changes/process-std.md
@@ -1,0 +1,5 @@
+---
+"@effection/process": "minor"
+---
+
+The `stdout` and `stderr` properties now return `OutputStream` instead of regular `Stream`. Addiotionally, the `buffered` option has been removed.
--- a/packages/stream/src/buffer.ts
+++ b/packages/stream/src/buffer.ts
@@ -1,0 +1,32 @@
+export interface Buffer<T> extends Iterable<T> {
+  push(value: T): void;
+}
+
+export function createBuffer<T>(size = Infinity): Buffer<T> {
+  if(size === Infinity) {
+    return new Array<T>();
+  } else {
+    return createRingBuffer<T>(size);
+  }
+}
+
+export function createRingBuffer<T>(size: number): Buffer<T> {
+  let head = 0;
+  let buffer: T[] = [];
+
+  return {
+    push(value: T): void {
+      buffer[head] = value;
+      head += 1;
+      if(head >= size) {
+        head = 0;
+      }
+    },
+
+    *[Symbol.iterator]() {
+      for(let current = 0; current < buffer.length; current++) {
+        yield buffer[(current + head) % buffer.length];
+      }
+    }
+  };
+}
--- a/.changes/buffer-resource.md
+++ b/.changes/buffer-resource.md
@@ -1,0 +1,5 @@
+---
+"@effection/subscription": "minor"
+---
+
+The `buffer` method on `Stream` returns a resource and can receive an optional limit using a ring buffer for efficient bounded caching
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -119,7 +119,6 @@
       "path": "./packages/process",
       "manager": "javascript",
       "dependencies": [
-        "@effection/channel",
         "@effection/core",
         "@effection/events",
         "@effection/stream"
--- a/.changes/remove-string-buffer.md
+++ b/.changes/remove-string-buffer.md
@@ -1,0 +1,5 @@
+---
+"@effection/subscription": "minor"
+---
+
+Remove the `stringBuffer` method on `Stream`
--- a/packages/main/test/node.test.ts
+++ b/packages/main/test/node.test.ts
@@ -1,17 +1,20 @@
 import expect from 'expect';
 import { describe, it, beforeEach } from '@effection/mocha';
+import { Stream } from '@effection/stream';
 
 import { exec, Process, ProcessResult } from '@effection/process';
 import { terminate, interrupt } from './helpers';
 
 describe('main', () => {
   let child: Process;
+  let stdout: Stream<string>;
 
   describe('with successful process', () => {
     beforeEach(function*() {
-      child = yield exec(`ts-node ./test/fixtures/text-writer.ts`, { buffered: true });
+      child = yield exec(`ts-node ./test/fixtures/text-writer.ts`);
+      stdout = yield child.stdout.lines().buffer();
 
-      yield child.stdout.filter((s) => s.includes("started")).expect();
+      yield stdout.grep("started").expect();
     });
 
     describe('interrupting the process', () => {
@@ -21,7 +24,7 @@ describe('main', () => {
       });
 
       it('shuts down gracefully', function*() {
-        yield child.stdout.filter((s) => s.includes("stopped")).expect();
+        yield stdout.grep("stopped").expect();
       });
     });
 
@@ -32,7 +35,7 @@ describe('main', () => {
       });
 
       it('shuts down gracefully', function*() {
-        yield child.stdout.filter((s) => s.includes("stopped")).expect();
+        yield stdout.grep("stopped").expect();
       });
     });
   });
--- a/packages/process/package.json
+++ b/packages/process/package.json
@@ -40,7 +40,6 @@
     "typescript": "^4.3.5"
   },
   "dependencies": {
-    "@effection/channel": "2.0.0-beta.16",
     "@effection/core": "2.0.0-beta.14",
     "@effection/events": "2.0.0-beta.16",
     "@effection/stream": "2.0.0-beta.1",
--- a/packages/process/src/exec.ts
+++ b/packages/process/src/exec.ts
@@ -1,7 +1,7 @@
 /// <reference types="../types/shellwords" />
 import { split } from 'shellwords';
 
-import { Task, Operation, Resource, withLabels } from '@effection/core';
+import { Task, Operation, Resource, spawn, withLabels } from '@effection/core';
 import { ExecOptions, Process, ProcessResult, CreateOSProcess } from './exec/api';
 import { createPosixProcess } from './exec/posix';
 import { createWin32Process, isWin32 } from './exec/win32';
@@ -42,22 +42,30 @@ export function exec(command: string, options: ExecOptions = {}): Exec {
     },
     join() {
       return withLabels(function*() {
-        let process = yield createProcess(cmd, { ...opts, buffered: true });
+        let process: Process = yield createProcess(cmd, opts);
+
+        let stdout = "";
+        let stderr = "";
+
+        yield spawn(process.stdout.forEach((chunk) => { stdout += chunk }));
+        yield spawn(process.stderr.forEach((chunk) => { stderr += chunk }));
 
         let status = yield process.join();
-        let stdout = yield process.stdout.expect();
-        let stderr = yield process.stderr.expect();
 
         return { ...status, stdout, stderr };
       }, { name: `exec(${JSON.stringify(command)}).join()`, expand: false });
     },
     expect() {
       return withLabels(function*() {
-        let process = yield createProcess(cmd, { ...opts, buffered: true });
+        let process: Process = yield createProcess(cmd, opts);
+
+        let stdout = "";
+        let stderr = "";
+
+        yield spawn(process.stdout.forEach((chunk) => { stdout += chunk }));
+        yield spawn(process.stderr.forEach((chunk) => { stderr += chunk }));
 
         let status = yield process.expect();
-        let stdout = yield process.stdout.expect();
-        let stderr = yield process.stderr.expect();
 
         return { ...status, stdout, stderr };
       }, { name: `exec(${JSON.stringify(command)}).expect()`, expand: false });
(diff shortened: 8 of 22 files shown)


jbolda:
I have run into a use case a handful of times now where it would be great to see the output from an `exec()` with the `stdout` and `stderr` laced together as you might see in a terminal. You can accomplish that "manually", but it would be convenient to add something to join()/expect() to allow one to easily do this.

This is sort of functionally equivalent to `stdio: 'inherit'` which is quite popular in child_process usage, but we don't really want to let the user muck with `stdio` as it becomes a rabbit hole and `'inherit'` blitzes any ability to inspect the output.

packages/subscription/src/buffer.ts
```diff
@@ -0,0 +1,41 @@
+export interface BufferLike<T> extends Iterable<T> {
+  push(value: T): void;
+}
+
+export class RingBuffer<T> implements BufferLike<T> {
+  private head = 0;
+  private buffer: T[] = [];
```
cowboyd: If we want to store a tail as well, then we could allocate the buffer up front so that the JavaScript engine can efficiently allocate underlying memory. E.g. `buffer:T[] = Array(size)`;

packages/subscription/src/buffer.ts
```diff
@@ -0,0 +1,41 @@
...
+    let head = this.head;
+    let current = 0;
+
+    return {
+      next: (): IteratorResult<T> => {
+        if(current >= this.buffer.length) {
+          return { done: true, value: undefined };
+        }
+
+        let value = this.buffer[(current + head) % this.buffer.length];
```
cowboyd: Three cheers for integer division! It's like the Haley's Comet of development. It comes around so rarely, you have to appreciate it when it does.

packages/subscription/src/buffer.ts
```diff
@@ -0,0 +1,41 @@
+export interface BufferLike<T> extends Iterable<T> {
+  push(value: T): void;
+}
+
+export class RingBuffer<T> implements BufferLike<T> {
```
cowboyd: For better or for worse, we now have the convention that any functions on a data structure will work unbound as in the case of de-structuring assignment, and I think we should adhere to that expectation. In this case, we already have a separate interface `BufferLike` separate from the implementation, so it seems like we wouldn't be losing anything by changing this to a `createRingBuffer()` function.

cowboyd (commented):
> I have run into a use case a handful of times now where it would be great to see the output from an `exec()` with the `stdout` and `stderr` laced together as you might see in a terminal. You can accomplish that "manually", but it would be convenient to add something to join()/expect() to allow one to easily do this.

I agree. In fact, this is most often what I have wanted when using effection as a manager of imperative commands. You only need the fully realized output in your program if you're going to do something with it, but if you're just running scripts for their side effects, as in an `npm install` or an `xcode build` what you want is to stream the output, even if it's a one-off `exec()`.

What parameters could we add to `join/expect`?

```ts
// this is pretty nice n' easy.
yield exec('npm install').expect({ stdio: 'inherit' });

// could overload to allow for inspection
yield exec('npm install').expect({ stdio: data => console.log(data.toString()) });

// could do individually
yield exec('npm install').expect({ 
  stdout: data => console.warn(data.toString())
  stderr: data => console.warn(data.toString())
});

// or a mix of the two
yield exec('npm install').expect({ 
  stdout: 'inherit',
  stderr: data => console.warn(data.toString())
});
```

Not sure what the best API is, but it's definitely needed.

Otherwise looks great! All other comments are in-line.

cowboyd:
> What parameters could we add to join/expect?

Another option would be to extend the `Exec` and `Daemon` builder apis:

```ts
yield exec(`npm install`).stdio('inherit').expect();
```

packages/subscription/src/stream.ts
```diff
@@ -21,12 +22,7 @@ export interface Stream<T, TReturn = undefined> extends OperationIterable<T, TRe
   collect(): Operation<Iterator<T, TReturn>>;
   toArray(): Operation<T[]>;
   subscribe(scope: Task): Subscription<T, TReturn>;
-  buffer(scope: Task): Stream<T, TReturn>;
-  stringBuffer(scope: Task): StringBufferStream<TReturn>;
-}
-
-export interface StringBufferStream<TReturn = undefined> extends Stream<string, TReturn> {
-  value: string;
+  buffer(limit?: number): Resource<Stream<T, TReturn>>;
```
cowboyd: I'm not a fan of making the default a potential footgun. I'd rather make the limit a mandatory parameter, and if you want to opt in to a completely unbounded buffer, then you can still do so, but you have to call it out.

```ts
let lines = yield app.stdout.lines.buffer(Infinity);
```

packages/subscription/src/buffer.ts
jnicklas: I think there are some downsides to this:

1. We need to keep track of how much we've filled the buffer with (not too bad)
2. We will need to allocate the entire buffer up front, even if we don't end up actually using it

I'm pretty sure that all JS implementations will do something more efficient than reallocating the underlying array every time you push a value. I think the normal way to do this is to do exponential growth, that is double the underlying size each time the capacity is exceeded. But it'll still be less efficient than allocating the whole thing upfront.

jnicklas:
An alternate design here is to add the `text()` and `lines()` methods to *all* streams. But I'm not sure how I feel about this. It would remove some complexity, but doing something like `someNumberStream.lines()` is obviously pretty weird.

packages/subscription/src/buffer.ts
jnicklas: The purpose of the `BufferLike` interface is to cover both a normal array and our ringbuffer with the same interface.

My thinking was that this ring buffer is sort of library-ish and should really be extracted into its own thing at some point. I just didn't like any of the existing implementations that I could find. So it should eventually be extraced, and then having it work more like other JS datastructures (`Set` , `Map`, etc), felt more natural to me.

packages/subscription/src/stream.ts
jnicklas: I would personally not come up with the idea of passing in `Infitity` here. We can call out that possibility in the documentation, but it's at least not very intuitive. There are many streams which don't ever emit very many values and for those streams, using an infinite buffer is just fine. I don't see this as that serious of a foot gun, so I'd prefer the more intuitive API.

jnicklas:
There is some danger to claiming that our implementation works the same as using `inherit`, because we're emulating its behaviour, but we can't actually guarantee it works exactly the same. I think flushing and the like could actually present issues where we behave slightly differently.

Like @jbolda noted, if we actually pass down `inherit`, we lose the ability to read data from the stdout and stderr streams, which is inconvenient.

It's pretty easy now to emulate the `inherit` behaviour yourself, using:

``` typescript
yield spawn(myProcess.stdout.forEach((c) => process.stdout.write(c)));
```

But if we add something like this, then IMO we shouldn't call this `inherit`. And I think it's challenging to find an API here which is intuitive and useful.

I think a good solution might be to actually support `inherit` for where it is useful. In that case the `stdout` and `stderr` streams would be empty. If someone wants *both* the ability to read the streams *and* have them be printed to the console, then implementing that yourself is pretty easy.

cowboyd:
> But if we add something like this, then IMO we shouldn't call this inherit. And I think it's challenging to find an API here which is intuitive and useful.

Yeah, I was hesitant about calling it `inherit` because that isn't what it is. However, I've very rarely not had to have this line

```ts
yield spawn(myProcess.stdout.forEach((c) => process.stdout.write(c)));
```

While it's easy, it's also tedious, especially when it's the 90 in the 90/10

I think it is an annoyance in our API, but I do prefer to kick the can down the road rather than committing to an alternative that turns out to not be good.

> An alternate design here is to add the text() and lines() methods to all streams. But I'm not sure how I feel about this. It would remove some complexity, but doing something like someNumberStream.lines() is obviously pretty weird.

We could have a `split()` method that split on a separator which would convert any stream of `T` into a stream of `T[]` and we be applicable to all streams.

packages/subscription/src/buffer.ts
cowboyd: Excellent point. There's no sense in trying to outwit 40 years of GC research.

packages/subscription/src/buffer.ts
cowboyd: I agree with having a `BufferLike` interface with multiple implementations of which `RingBuffer` is one, and functionally, it's exactly the same whether you call `new RingBuffer(size)` or `createRingBuffer(size)`. The main thing is having the code abide by the conventions of the codebase in which it lives. I.e. when in Rome....

It's not a biggie, but my goal in suggesting it is to make things as consistent as possible for future contributors.

packages/subscription/src/stream.ts
cowboyd: What about making the limit default to `Infinity` instead of being nullable? I think that would be a good compromise that allows you to easily create infinite buffers, but the typedocs, and the auto-complete will keep the potential pitfalls visible in your peripheral vision.

cowboyd:
> We could have a split() method that split on a separator which would convert any stream of T into a stream of T[] and we be applicable to all streams.

Note that I still think there should be an `IOStream` with `text()` and `lines()` even if we add a `split()` method. It's just too handy, and there's no sense it trying to squeeze it into the super-interface. Maybe `TextStream` is a better name though since IO implies that it's read/write and I can imagine this being very useful for any stream unrelated to `stderr` and `stdout` ?

packages/stream/src/buffer.ts
```diff
@@ -0,0 +1,41 @@
...
+          return { done: true, value: undefined };
+        }
+
+        let value = this.buffer[(current + head) % this.buffer.length];
+
+        current += 1;
+        return { done: false, value };
+      }
+    };
+  }
```
cowboyd: Totally optional, but could be perhaps more succinctly expressed with a generator:

cowboyd:
Much nicer API 👍🏻 

I'm assuming the changes we talked about managing the process streams via `stdio` options will be in a follow-on?

packages/subscription/src/buffer.ts
jnicklas: I've updated the buffer code to be more in line with our other APIs, and it actually made it a bit cleaner.

jnicklas:
@cowboyd I think this is ready for another look. I will post a follow up PR for the changes we worked on yesterday to improve stdio handling, since it really is a separate feature.

packages/subscription/src/buffer.ts
jnicklas: I had a super gross implementation here first, which did some weird wrapping around to make this work, until I realized that this implementation is possible and it made me really happy 😅

.changes/io-stream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effection/subscription": "minor"
+---
+
+Add `IoStream` interface and `createIoStream` function. IoStreams are like regular streams, but can convert output into text or linewise output.
```
cowboyd:
```suggestion
Add `OutputStream` interface and `createOutputStream` function. OutputStreams are like regular streams, but can convert output into text or linewise output.
```

.changes/process-std.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effection/process": "minor"
+---
+
+The `stdout` and `stderr` properties now return IoStreams instead of regular streams. Addiotionally, the `buffered` option has been removed.
```
cowboyd:
```suggestion
The `stdout` and `stderr` properties now return `OutputStream`s instead of regular streams. Additionally, the `buffered` option has been removed.
```

cowboyd (approved):

packages/stream/src/buffer.ts
```diff
@@ -0,0 +1,42 @@
...
+          return { done: true, value: undefined };
+        }
+
+        let value = buffer[(current + head) % buffer.length];
+
+        current += 1;
+        return { done: false, value };
+      }
+    };
+  }
```
cowboyd: 100% optional, but might be a tad clearer as a generator

packages/stream/src/buffer.ts
jnicklas: 🤦 of course! It's *much* easier to express with a generator!

PR #541, opened 2021-09-22
Add grep method to Stream

jnicklas:
This adds a grep method to stream, the point of this is to make it easier to scan for a given string, especially in the upcoming iostreams.

``` typescript
// before
yield myProcess.stdout.lines().filter((l) => l.includes("ready")).expect();
// after
yield myProcess.stdout.lines().grep("ready").expect();
```


cowboyd (approved):
😍 ❤️ ❤️‍🔥

PR #544, opened 2021-09-23
Add homepage for API docs and include more packages

jnicklas:
Rather than using the README for the typedoc documentation, we add a specific home page which explains the structure of the documentation and links to the relevant packages.

We explain the purpose of Effection in the guides and other sections of the homepage, so repeating it here is unnecessary and unhelpful.

Before:

Screenshot 2021-09-23 at 14 12 30

After:

Screenshot 2021-09-23 at 14 11 21

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,0 +1,24 @@
+# Effection
+
+Here you will find a complete API reference for all effection subpackages. The
+[effection][] package reexports several other packages,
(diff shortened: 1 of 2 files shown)


cowboyd:
This looks good to me, but I'll leave final call to @jorgelainfiesta

PR #545, opened 2021-09-24
Document core package

jnicklas:
Adds typedoc documentation for the @effection/core package


cowboyd (approved):
This is absolutely fantastic, and I didn't see anything obvious that needed my input :)

I think the best thing at this point is to get it deployed and then have another read because it's orders of magnitude better that what we have now (which is nothing)

PR #546, opened 2021-09-27
Upgrade typedoc

jnicklas:
Upgrade typedoc and remove the custom theme. Custom themes no longer work the same in typedoc 0.22 and later and we don't want to be stuck on an old version. We have the option of adding custom styles in the `typedoc.css` file, for a lightweight styling approach.


jnicklas:
A screenshot:

Screenshot 2021-09-27 at 12 38 09

IMO this looks just fine.

minkimcello (approved):

PR #547, opened 2021-09-27
Split `Stream` into its own package

jnicklas:
Splits off `Stream` from the subscription package into its own `@effection/stream` package.

This just makes more sense from an organizational perspective.

The fact that `Stream` is in the subscription package is kind of an accident, and it feels a bit odd that such a fundamental structure doesn't have its own package.


cowboyd (approved):
Makes sense, and we should probably communicate somewher either in the guides, or in the API docs, that the public API that is commited to is `effection` and that any of the internal packages may change in breaking ways.

PR #549, opened 2021-09-28
Add package readmes

cowboyd:
## Motivation

Anytime someone navigates to a package of ours via NPM, they are presented with a "this package does not have a readme" which not only looks un-profesh, it doesn't help them out. We don't know why they got to that page... probably via an `npm ls` or a `yarn why`, but however they did get there, it shouldn't be the end of their journey. Instead, we want whatever leaf node of our network that people encounter to be an onramp to our community.

## Approach

This removes almost everything from each individual package README, and instead includes links to the 
1. licence of the package
2. our website
3. our discord
4. a link to the effection website
5. a link to the relevant guide for the packgae if one exists.


jnicklas (approved):
Beuatiful 👯

PR #550, opened 2021-09-29
Remove operation resolutions

### LINKED ISSUE #548: Deprecate resolutions
We don't really need resolutions anymore now that we have futures, we should deprecate them.

jnicklas:
With the integration of Future, we have a better primitive for the kinds of problems that operation resolutions were meant to solve. Rather than having multiple separate ways of doing the same thing, let's double down on futures and remove resolution functions entirely.

Closes #548


cowboyd (approved):
😎😎
image

PR #553, opened 2021-09-29
Remove internal dev dependencies

jnicklas:
Covector can currently not distinguish between dev dependencies and regular dependencies. We settled on not specifying dev depencies as covector dependencies, but this leads to a situation where they are not getting bumped at all, causing us to depend on old versions. Since yarn is able to resolve the packages anyway, as long as they are in the same workspace, we can just get rid of the devDependency entries for them entirely.

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -143,7 +143,8 @@
       "path": "./packages/stream",
       "manager": "javascript",
       "dependencies": [
-        "@effection/core"
+        "@effection/core",
+        "@effection/subscription"
       ]
     },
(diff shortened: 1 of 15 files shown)


jnicklas:
I don't think this is an ideal solution, but it's better than the status quo, imo.

cowboyd (approved):
Seems fair, although it should be noted that this is a quirk of how yarn workspaces function, and won't work with `npm 7` which we eventually want to move to. Still, better to have correct release logic in exchange for kicking this can down the road.

PR #561, opened 2021-09-30
Ignore externals in typedocs

cowboyd:
## Motivation

There is no need to include symbols from external libraries, especially ones that are not re-exported. Maybe there is a reason that this is the default, but for me they only seem to clutter things up.

## Approach

Leave 'em out!

## Screenshots

**before**

image

**after**

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ packages/*/docs
 .node-version
 .parcel-cache
 dist*/
+docs
 *.tgz
 pkg/
 .nyc_output/
--- a/typedoc.js
+++ b/typedoc.js
@@ -15,6 +15,7 @@ module.exports = {
   "customCss": "typedoc.css",
   "name": "Effection API Reference",
   "entryPointStrategy": "packages",
+  "excludeExternals": true,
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,6 +4,7 @@ packages/*/docs
 .node-version
 .parcel-cache
 dist*/
+docs
```
jnicklas, line 7: Where did the commit go which fixed this? I know I fixed this at some point, but this is not the correct (or complete) fix. The folder in the `./website/docs` gets ignored this way, which is incorrect.

jnicklas: See #563

cowboyd:
Actually, was  indicative of the fact that `@effection/fetch` was exporting `AbortController` for no good reason, and so this flag only serves to cover up symptoms. It's much better to just [fix the exports](https://github.com/thefrontside/effection/pull/566/files#diff-bb3fc1dad2ae5a09c4f6c065a987d0d6719e469a95aaad147885938e1bbdcdde) rather than suppress the warning signs.

PR #566, opened 2021-10-01
📝 Add typedocs for @effection/fetch

cowboyd:
## Motivation

Document all the things 

## Approach

I always like plenty of examples in my docs, so I tried to include some as hints about usage, but without going so overboard that it didn't let you get a concise definition of what the method did. Not sure what the best balance is here. 

I also optimized the module exports while doing this to account for two things:

1. The `Fetch` interface was not being exported from `fetch.ts` which makes it difficult to properly type a reference to `fetch()` using typescript. It also means that it wasn't showing up in the typedoc.
2. `AbortController` was being exported for no good reason that I can determine. Any abort signal that you pass down into fetch is ignored. As far as I can tell the only effect of exporting `AbortController` is that it showed up in the typedoc.

### DIFF
--- a/packages/fetch/src/fetch.ts
+++ b/packages/fetch/src/fetch.ts
@@ -2,14 +2,91 @@ import { ensure, withLabels, Operation, Resource } from '@effection/core';
 import { fetch as nativeFetch } from 'cross-fetch';
 import { AbortController } from 'abort-controller';
 
+/**
+ * A resource that can be used to create an HTTP request. This is the return
+ * value of [[fetch]]. It also contains helper methods for performing common
+ * requests.
+ */
 export interface Fetch extends Resource<Response> {
+  /**
+   * An operation that executes the request and produces an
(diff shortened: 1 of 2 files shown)


packages/fetch/src/fetch.ts
```diff
@@ -2,14 +2,91 @@ import { ensure, withLabels, Operation, Resource } from '@effection/core';
...
+ * allowing you to handle the entire request cycle in a single line. For
+ * example, the following will execute a request, read it fully, and parse it
+ * as JSON. An exception is raised if the response is not in the 200 range.
+ *
+ * ```ts
+ * let json = yield fetch('https://example.com/movies.json').json();
+ *```
+ *
+ * Like most operations in Effection, [[fetch]] is stateless, and does not
+ * actually _do_ anything unless it yielded to:
```
cowboyd:
```suggestion
 * actually _do_ anything unless it is yielded to:
```

jnicklas (approved):

PR #567, opened 2021-10-01
📝Remove fetch as an entry point to main typedocs

cowboyd:
## Motivation

`fetch()` and its associated APIs are now part of the main `effection` module and as such, `@effection/fetch` is more of a private API and should not appear as part of the typedocs since there's no guarantee that it won't be moved somewhere else at any point.

## Approach

remove it from the sidebar.

### DIFF
--- a/typedoc.json
+++ b/typedoc.json
@@ -9,7 +9,6 @@
   "entryPoints": [
     "packages/atom",
     "packages/effection",
-    "packages/fetch",
     "packages/mocha",
     "packages/process",
     "packages/react",


jnicklas:
In #544 I went the other way and exposed *all* of the packages. I find the docs of the subpackages to be much easier to read, and it's easier to see which parts belong together.

cowboyd:
> In #544 I went the other way and exposed _all_ of the packages. I find the docs of the subpackages to be much easier to read, and it's easier to see which parts belong together.

I can see that. I guess what I’m trying to discourage is folks thinking about the npm packages themselves as a public API, but I agree grouping them into categories of usage is super helpful. Is there any way to control the labels on the sidebar?

jnicklas:
> Is there any way to control the labels on the sidebar?

Maybe. TypeDoc has a category feature, but IMO it does not work very well, and last I checked it does not apply the categories in the sidebar at all.

cowboyd:
I suppose we can leave it then. I just don't want anyone to be tempted to `import { Stream } from '@effection/stream'` that way, when we move things around internally like we did when we moved streams out of subscription, we're free to do so without yanking the carpet out from underneath someone.

PR #573, opened 2021-10-07
Release v2 of Effection 🥳🥳🥳

jnicklas:
## Motivation

We want to actually ship something

## Approach

ship it squirrel

### Alternate Designs

Don't ship it

We have to live with it as is for a while

### TODOs and Open Questions
- [ ] Should we ship it?


packages/atom/package.json
```diff
@@ -41,9 +41,9 @@
     "extends": "../../package.json"
   },
   "dependencies": {
-    "@effection/channel": "2.0.0-beta.21",
-    "@effection/core": "2.0.0-beta.17",
-    "@effection/stream": "2.0.0-beta.6",
+    "@effection/channel": "^2.0.0",
+    "@effection/core": "^2.0.0",
+    "@effection/stream": "^2.0.0",
```
cowboyd: This might be a good time to change the "contrib" packages to use "effection" directly.

cowboyd (commented):
Jeremiah Johnson

packages/atom/package.json
jnicklas: Yes, we might want to do that. I'm still not sure which option is better 🤷‍♂️

cowboyd (approved):

PR #575, opened 2021-10-08
Blog post introducing Effection

jnicklas:
This is my first draft. Let's explore how to write this post together.

[Rendered](https://deploy-preview-575--effection.netlify.app/effection/docs/guides/blog/)


taras:
@jnicklas it's wonderful. I love the tone and directness while being thought-provoking. It's also very short which makes it a nice fast read. I had a few thoughts while reading it.

Do we want to mention when one might encounter a problem with async/await? 

> For example, async/await works well while you're being very optimistic about what will happen but immediately falls apart when the constraints change. For example, you might find that your application works well when it runs at a human speed but starts failing when running much faster by automated testing. It's often too late to make changes to the application and you end up creating workarounds or avoid doing integration testing. Effection allows you to ensure that your application will work reliably in all conditions, not just the once that you can perceive during development. 

^^ I'm not sugging that we should use this example specifically, but I wonder it's helpful to mention that Effection helps you handle asynchrony correctly without surprises. Another example might be better.

2. Do we want to mention that we applied Effection in many different contexts within BigTest?

> BigTest is a brand new integration testing platform that we're creating from scratch that prioritizes developer experience. A good developer experience has to be fast and responsive. Creating a responsive system that is high concurrent requires a lot of precision. Everything has the instant the previous operation completes without polling and cancellation has to be effortless otherwise precious developer time is wasted waiting for irrelevant work to complete. When a developer wants the application to stop, they might be able to click CTRL-C and everything should stop without leaving any processes hangingaround. Reaching this level of precision without Effection would be impossible.

^^ again, not suggesting that we use these specific words, but I wonder if we want something in this spirit for people to understand why we created effection.

3. Similar to suggestion 2, do we want to mention the asynchrony use-cases that we cover? Things like websockets, events, promises, and shell spawn with cancallation?

cowboyd:
Yes, this is really awesome, and I almost hesitate to add content to it for fear of diluting it.

One thing that did occur to me is that it might connect more quickly if instead of using `spawn()` in the examples, we use `Promise.all()` since `Promise.all()` is familiar while `spawn()` is new (but also has an analogue in Effection). Furthermore, using `Promise.all()` and `Promise.race()` are taught as the "correct" way to do concurrency in JavaScript because they at least all give error safety to the enclosing operation. In other words, if you are using `await Promise.race()` and `await Promise.all()` then it is impossible for the enclosing task to not fail if any child task fails.

By showing the flaws inherent in with these familiar "correct" operations, we can highlight the need for alternatives that also feel familiar.

I don't know if these would be helpful, but maybe we can spice up the blog by adding diagrams to show the difference:

**async / await**

IMG_2019

**Effection**

IMG_2020

These certainly only scratch the surface of what you can do with effection, and it doesn't even touch on explicit cancellation as a third outcome, but does imply why the ability to programmatically cancel without programmer intervention is so important to writing correct code.

jorgelainfiesta:
Great write up Jonas! I'd condense the intro a bit, and add a conclusion that it's kind of call to action. I can take care of that as part of the editing process.

Regarding @taras  suggestions, I'd leave them for another post like "How we use effection in Bigtest". 

About @cowboyd, while that'd make the post tighter, I don't know if most people are familiar with that syntax. I'd just shot for the current examples and add a short paragraph saying "even if you use Promse.race you'd still run into XYZ because ABC." The diagrams would be a great additions though. 

If we can decide on a conclusion for the syntax, I can take the text through the editing process and set it to be ready to be published in our blog.

jorgelainfiesta:
Thanks Jonas! the diagrams are great! I'll start the editing process with the current copy!

jnicklas:
This is now published

PR #581, opened 2021-10-14
Apply new styles for docusaurus

jorgelainfiesta:
## Motivation

Make it easier for people to navigate through frontside's website

## Approach

Applied the same theme from interactors by copy+pasting the relevant files. 

## Screenshot

Screenshot 2021-10-14 at 17 42 26


cowboyd:
This looks fantastic. I tried all the links from my phone and they worked great. Questions

1. Are we going to do a separate PR for the type docs? At the very least they need abotonar to link back to the main docs.
2. I’ve seen on some sites the “external link” icon on the top bar next to links like GitHub and discord. Would that be appropriate here?

jorgelainfiesta:
1. I have yet to create the styles for Typedoc, it's unexplored territory for me. Last time I tried to create styles it was very difficult, but Jonas updated it to a new version that relies on JSX I think. Maybe it'll be easier. I'd like to prioritize having the three sites with compatible navigations. Then I'll move to the API docs, which opens in a new tab at the moment for that reason.

2. I removed the external link icons from the navigation because almost all the links were 'external', plus "Github" and "Discord" kind of imply external on their own so it felt redundant. Footer still has the external icons because I didn't do much style work there, but it looks a bit saturated with icons imo.

cowboyd (approved):
Well I love it! The only downside is that now it looks incomplete when you actually use it without the same navbar being on the BigTest and Effection sites 😆

jnicklas:
I like it a lot, the only feedback I have is that there are two similar but not identical arrows close to each other, which looks odd to me:
Screenshot 2021-10-15 at 12 48 15

jorgelainfiesta:
Arrows are fixed, and I have a PR for each of the three sites to have the new styles. We can merge them all together so the navigation is smoother.

PR #584, opened 2021-10-21
Improved inspector protocol

jnicklas:
The current inspector protocol is very naive. Essentially on each and every state change we serialize the *entire* Effection tree and sent it across the wire. While this works (somewhat) for very small sample applications, it utterly fails for complex applications with thousands of tasks. This creates a much more efficient communication protocol where only the changes that occur are sent across the wire. On the other end, we then reassemble those changes into a state tree, which we can
pass to react to render.


jnicklas:
This is only failing on covector now, I think because the dispatch package is not published. I'm having some problems getting it published though.

cowboyd (approved):
I'm kinda excited about the possibilities of how dispatch can be used to synchronize state trees as we've seen now in three separate cases.

Issue #589, opened 2021-11-11
Add `AbortSignal` resource


cowboyd:
[AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) is a standard JavaScript mechanism for signalling cancellation. In fact, we use it inside `@effection/fetch`. However, for integration with _any_ 3rd party library that uses abort signals as native `fetch()` does, it would be really nice to have the ability to quickly create an abort signal that is tied to the current task's scope, and then uses it as a connection between that task and the 3rd party:

```js
function* request() {
  let signal = yield createAbortSignal();
  return yield fetch('/some/url', { signal });
}
```

steps:
- [ ] refactor @effection/fetch to use an abort signal resource internally.
- [ ] move abort signal resource to either its own package or `@effection/core` and export from `effection`
- [ ] add typedocs

taras:
@cowboyd  can you write the docs for this before implementation so it gets done right away?

cowboyd:
There's the matter of the typedocs and the guides. Here's the typedoc:

```ts
/**
 * Create an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) bound 
 * to the current scope. Whenever that scope is completed, errored, or halted, the abort signal will 
 * be triggered.
 *
 * Cancellation and teardown is handled handled automatically in Effection, but in systems where
 * it is not, the lifespan of a transaction needs to be explicitly modeled. A very common way to do this is with 
 * the `AbortSignal`. By creating an abort signal bound to an Effection task, we can pass that signal down to 
 * 3rd party consumers that can use it to register shutdown callbacks that will be invoked whenever the task is done
 * (either by completion, failure, or cancellation). An example is the native [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch#syntax) API which takes a `signal` parameter in order to trigger when the HTTP request should
 * be cancelled.  This is how you would bind the lifetime of the HTTP request to the lifetime of the current task.
 * 
 * ```js
 * function* request() {
 *   let signal = yield createAbortSignal();
 *   return yield fetch('/some/url', { signal });
 * }
 * ```
 */
```

@jnicklas where do you think this should slot into the guides?

jnicklas:
@cowboyd I think the most natural place would be in the lifecycle guide, just after `ensure` https://frontside.com/effection/docs/guides/lifecycle#ensure

[...]

PR #595, opened 2021-11-18
Extract abort-controller from fetch to core 

minkimcello:
## Motivation

To address #589

## Approach

I put the new resource in `@effection/core` instead of creating a new package

## TODOs

- [x] Add typedocs

### DIFF
--- a/.changes/abort-signal.md
+++ b/.changes/abort-signal.md
@@ -1,0 +1,6 @@
+---
+"@effection/fetch": "minor"
+"@effection/core": "minor"
+---
+
+Extract `AbortSignal` from `@effection/fetch` to `@effection/core` as a resource
--- a/.changes/abort-signal.md
+++ b/.changes/abort-signal.md
@@ -1,5 +1,5 @@
 ---
-"@effection/fetch": "minor"
+"@effection/fetch": "patch"
 "@effection/core": "minor"
 ---
 
--- a/packages/core/src/abort-signal.ts
+++ b/packages/core/src/abort-signal.ts
@@ -1,8 +1,8 @@
 import { spawn } from './operations/spawn';
 import { Resource } from "./operation";
-import { AbortController } from 'abort-controller';
+import { AbortController, AbortSignal } from 'abort-controller';
 
-function createAbortSignal(): Resource<any> {
+function createAbortSignal(): Resource<AbortSignal> {
   return {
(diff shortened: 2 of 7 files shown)


.changes/abort-signal.md
```diff
@@ -0,0 +1,6 @@
+---
+"@effection/fetch": "minor"
```
cowboyd: This is just a patch to fetch because it’s api does not change

cowboyd (commented):
Looking good. We should add labels to both the resource and the spawned teardown block.

minkimcello:
@cowboyd like [this](https://github.com/thefrontside/effection/pull/595/commits/acfd471885c4d883deac498af4dbdc43456203e1)?

packages/core/src/abort-signal.ts
```diff
@@ -0,0 +1,28 @@
+import { spawn } from './operations/spawn';
+import { Resource } from "./operation";
+import { AbortController, AbortSignal } from 'abort-controller';
+
+function createAbortSignal(): Resource<AbortSignal> {
+  return {
+    labels: {
+      name: 'createAbortSignal()'
+    },
```
cowboyd: This will work, although name is special and you can use it directly. Also I would name it 'AbortSignal' since that is what the resource is (createAbortSignal) is the op that makes it

packages/core/src/abort-signal.ts
```diff
@@ -0,0 +1,28 @@
...
+      let controller = new AbortController();
+      yield spawn(function* () {
+        try {
+          yield;
+        } finally {
+          controller.abort();
+        }
+      }, {
+        labels: {
+          name: 'controller.abort()'
```
cowboyd: This should probably be something like "ensure signal aborted"

packages/core/src/abort-signal.ts
cowboyd: in other words, you can just say:

```ts
{
  name: 'AbortSignal',
  *init() {...
```

cowboyd:
@minkimcello looks great! Let's drop in the typedoc, and then I'll update the guides.

cowboyd (approved):

cowboyd:
@minkimcello guides added 👍🏻

jnicklas (approved):

PR #599, opened 2022-01-06
Add a guide for TypeScript

cowboyd:
## Motivation

There are some pitfalls with using Effection and TypeScript together, and with the expectations of type-safety that come with TypeScript, it can be one of the biggest stumbling blocks to adoption. While we can't fix them directly (only the TypeScript team can do that), we can at least explain what they are, _why_ they can't be fixed by anybody but the TypeScript team,  and ways that you can mitigate them.

## Approach

I tried to just enumerate the challenges that we've encountered, and propose some of the solutions.

Things I'm not entirely sure of:

- Is this the best location for this guide?
- In which order should I put the sections: The one on operations first? or generators?

### DIFF
--- a/website/docs/guides/typescript.md
+++ b/website/docs/guides/typescript.md
@@ -174,11 +174,7 @@ upvoting one of the [proposed solutions][2]. It doesn't have to be an essay,
 just a simple, true statement like the following to let them know you're out
 there:
 
-``` text
-I use JavaScript generators a lot to write more powerful code than would
-be possible otherwise. It would be amazing if TypeScript were able to typecheck
-programs like mine in a 100% hassle-free way.
-```
+> I use JavaScript generators a lot to write more powerful code than would be possible otherwise. It would be amazing if TypeScript were able to typecheck programs like mine in a 100% hassle-free way.
 
 You can make a difference!
 [operations]: ./tasks#operations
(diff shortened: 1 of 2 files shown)


taras:
> Effection itself is written in TypeScript, and so comes bundled with its own type definitions. As a result, there's nothing special required to make them interoperate. However, there are still some idiosyncracies with using Effection and TypeScript together. This section contains information about what these idiosyncrasies are, and what you can do to make your development experience as painless as possible.

@cowboyd what do you think about this instead

> Effection is written in TypeScript and comes bundled with its own type definitions. Effection doesn't require any special setup to use in a TypeScript project, but there are some TypeScript specific idiosyncrasies to keep in mind. This section describes these idiosyncrasies and what you can do to make development experience of using Effection in your TypeScript project as painless as possible.

cowboyd:
@taras Feels great!

dagda1 (approved):
I definitely like the `unwrap` helper 👍

cowboyd:
@taras @dagda1 What would you think about adding a suggestion that they respectfully upvote the issue for generator typing on GitHub? The TS folks have indicated that it's the best way to register interest.

dagda1:
> @taras @dagda1 What would you think about adding a suggestion that they respectfully upvote the issue for generator typing on GitHub? The TS folks have indicated that it's the best way to register interest.

I don't think it would do any harm.  If generators were more mainstream then they would do something.

cowboyd:
I'm going to go ahead and merge. @taras LMK if you have any feedback on the CTA at the end and I'll submit another PR.

PR #600, opened 2022-01-10
add OperationObject and `toOperation` method

wKich:
## Motivation

In https://github.com/thefrontside/interactors/pull/166 we used `effection` for interactors, but running interactions as operations are very inconvenient.

## Approach

Added `OperationObject` and support creating controller from it.


packages/core/src/operation.ts
```diff
@@ -3,6 +3,8 @@ import type { Task } from './task';
 import type { Labels } from './labels';
 import type { FutureLike } from './future';
 
+export const toOperation = Symbol.for('toOperation');
```
cowboyd: I think that we should use the converntions for symbol names even I the name of the variable isn't `Symbol.operation`

```suggestion
export const toOperation = Symbol.for('Symbol.operation');
```

packages/core/src/operation.ts
```diff
@@ -40,6 +42,10 @@ export interface OperationFunction<TOut> extends Labelled {
   (task: Task<TOut>): Operation<TOut>;
 }
 
+export interface OperationObject<TOut> extends Labelled {
```
cowboyd: As mentioned in chat, I think that this should not be a `Labelled`.

```suggestion
export interface OperationObject<TOut> {
```

cowboyd (dismissed):

packages/core/src/operation.ts
wKich: In that case we have to add checks to almost all places where we are trying to get labels from operation. Because `name` and `labels` fields are optional, I guess is better to make it extends from `Labelled` to reduce change impact

packages/core/src/controller/object-controller.ts
```diff
@@ -0,0 +1,46 @@
...
+import { OperationObject } from '../operation';
+
+export function createObjectController<TOut>(task: Task<TOut>, _obj: OperationObject<TOut>, createController: () => Controller<TOut>): Controller<TOut> {
+  let delegate: Controller<TOut>;
+  let { produce, future } = createFuture<TOut>();
+
+  function start() {
+    try {
+      delegate = createController();
+      task.setLabels({ ...delegate.operation?.labels });
```
wKich: @cowboyd, what do you think about it? `setLabels` already merges `task.labels` and even more `task.labels == obj.labels` so I thought there is no reason to pass those.

packages/core/src/operation.ts
cowboyd: I disagree. Even though it is painful for us, the user never works with the labels directly, they use `withLabels()` and `setLabels()` so I think that we need to cope with the pain internally to simplify the external interface because having a `name` property is very common for an object, and we should avoid that conflict even if that name is also a string.

cowboyd:
I had an idea 💡 

Custom symbols are awkward because they diverge from the well-known symbols, so it's hard to know if you're using them correctly. Specifically, every other symbol is clearly marked in the code as `[Symbol.x]` like `[Symbol.iterator]` or `[Symbol.toStringTag]` What if we made it so that we could use our Effection symbols that way? Monkey-patching the global `Symbol` object is a bad idea, but we could have the best of both worlds if we used a `Symbol` pony fill:


```ts
//core/src/symbol.ts
const SymbolOperation = Symbol.for('Symbol.operation');

export interface SymbolExtentions {
  operation: typeof SymbolOperation;
}

export type SymbolConstructor = typeof Symbol & SymbolExtentions;

const SymbolExtended: SymbolConstructor = Object.create(Symbol, {
  operation: {
    writable: false,
    enumerable: false,
    configurable: false,
    value: SymbolOperation
  }
});

export { SymbolExtended as Symbol };
```

This would allow us to use it just like everywhere else, but without fear of conflict, and existing symbols would continue to work because we'd use the global constructor as the prototype:

```ts
import { Symbol, Operation } from 'Effection';

const operator: Operation<string> = {
  *[Symbol.operation]() {
      return yield Promise.resolve("hello world");
  },
  *[Symbol.iterator]() {
    return yield* [1,2,3];
  }
}
```

In either case, we need to make sure that type checking works globally and so will probably have to include some version of this hack:

https://github.com/thefrontside/effection/blob/%40effection/subscription-v0.9.0/packages/subscription/src/symbol-subscribable.ts 

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection">effection/symbol-subscribable.ts at @effection/subscription-v0.9.0 · thefrontside/effection</a></strong></div><div>A framework for Node and the browser that makes building concurrent systems easy to get right. - effection/symbol-subscribable.ts at @effection/subscription-v0.9.0 · thefrontside/effection</div></blockquote>

packages/core/src/controller/object-controller.ts
```diff
@@ -0,0 +1,54 @@
...
+export function createObjectController<TOut>(task: Task<TOut>, obj: OperationObject<TOut>, createController: () => Controller<TOut>): Controller<TOut> {
+  let delegate: Controller<TOut>;
+  let { produce, future } = createFuture<TOut>();
+
+  function start() {
+    try {
+      delegate = createController();
+      let labels: Labels = {};
+      if (isObjectOperation<TOut>(delegate.operation)) labels = extractLabels(delegate.operation);
+      if (isNotObjectOperation<TOut>(delegate.operation)) labels = delegate.operation?.labels ?? {};
```
cowboyd: Let's just use `else`. Also, I rarely weigh in on style, but blockless if was, after all, the cause of [heartbleed](https://heartbleed.com).

packages/core/src/labels.ts
```diff
@@ -16,7 +18,13 @@ export type Labels = Record<string, string | number | boolean>;
  * entirely.
  */
 export function withLabels<T>(operation: Operation<T>, labels: Labels): Operation<T> {
-  if(operation) {
+  if (isObjectOperation<T>(operation)) {
+    let original = operation[Symbol.operation];
+    Object.assign(operation, {
+      ...labels,
+      [Symbol.operation]: original
+    });
```
cowboyd: Why mutate if we don't have to?

```suggestion
    return {
      ...labels,
      [Symbol.operation]: original
    };
```

packages/core/src/labels.ts
```diff
@@ -16,7 +18,13 @@ export type Labels = Record<string, string | number | boolean>;
...
+  } else if (operation) {
     operation.labels = { ...(operation.labels || {}), ...labels };
```
cowboyd, line 28: Same here, we no longer need to mutate. Also the use of an `OperationObject` as a delegate allows for graceful handling of `undefined` operation.

```ts
else {
  return {
    ...(operation && operation.labels),
    ...labels,
    [Symbol.operation]: operation
  }
}
```

packages/core/src/predicates.ts
```diff
@@ -19,3 +20,11 @@ export function isFuture<T>(value: any): value is FutureLike<T> {
...
   return typeof(value) === 'object' && typeof(value.init) === 'function';
 }
+
+export function isObjectOperation<TOut>(x: unknown): x is OperationObject<TOut> {
+  return typeof x === 'object' && x != null && Symbol.operation in x;
+}
+
+export function isNotObjectOperation<TOut>(x: unknown): x is Exclude<Operation<TOut>, OperationObject<TOut>> {
+  return !isObjectOperation(x);
+}
```
cowboyd: I'm not sure this really carries its weight if it is used once and can just be a simple `!`

packages/core/src/symbol.ts
```diff
@@ -0,0 +1,18 @@
+const SymbolOperation = Symbol.for("Symbol.operation");
+
+export interface SymbolExtentions {
+  operation: typeof SymbolOperation;
```
cowboyd, line 7: We need to make sure that these are globally compatible. In other words, if someone defines an operation with different versions of Effection (imports `Symbol` from different ES modules) that it will still be type checkable.

In other words if we copy and paste this file into `symbol1.ts` and `symbol2.ts` we should be able to:

```ts
import { Symbol as Symbol1 } from './symbol1';
import { Symbol as Symbol2 } from './symbol2';

// they are assignable from each other and so operations will type-check
const operation1: typeof Symbol2.operation = Symbol1.operation;
const operation2: typeof Symbol1.operation = Symbol2.operation;

assert(operation1 === operation2);
```

packages/core/src/controller/function-controller.ts
```diff
@@ -10,9 +12,11 @@ export function createFunctionController<TOut>(task: Task<TOut>, fn: OperationFu
   function start() {
     try {
       delegate = createController();
+      let labels: Labels = {};
+      if (isObjectOperation<TOut>(delegate.operation)) labels = extractLabels(delegate.operation);
```
cowboyd: Can we put this extract labels business into the controller?

packages/core/src/controller/object-controller.ts
wKich: I can't use `else` here. Typescript doesn't understand it

packages/core/src/labels.ts
wKich: Because that's how `withLabels` works according the documentation `Apply the given labels to an operation.` So it doesn't create new operation. Or I don't understand something? Originally `withLabels` replaces some labels in operation, so I tried to keep that logic. And if we change it, we will not replace labels, instead we create another operation, keep old labels and start depending on `controller` implementation.

packages/core/src/controller/object-controller.ts
wKich: Hm... interesting, I tried to use `else` before and got typescript complaining

packages/core/src/labels.ts
wKich: So real replacement will happened in `controller`

wKich: I mean this
image

wKich: Should we change tests?

packages/core/src/symbol.ts
wKich: I guess it will not work with types. But we can make it work in runtime

wKich: So you mean to patch `Symbol` global object, right?

cowboyd: No, not patch the global symbol, just make sure that they have the same type. See https://github.com/microsoft/TypeScript/issues/35909

packages/core/src/index.ts
```diff
@@ -4,13 +4,14 @@ import { Effection } from './effection';
 
 export { State, StateTransition } from './state-machine';
 export { createTask, Task, TaskOptions, TaskInfo, TaskTree } from './task';
-export { Operation, Resource } from './operation';
+export { Operation, Resource, OperationObject } from './operation';
```
wKich: @cowboyd Why we don't export other operation types? Are there any rules what stuff should be exported from lib and what shouldn't?

packages/core/src/labels.ts
```diff
@@ -15,19 +18,30 @@ export type Labels = Record<string, string | number | boolean>;
  * removed. See {@link setLabels} if you want to replace the existing labels
  * entirely.
  */
-export function withLabels<T>(operation: Operation<T>, labels: Labels): Operation<T> {
-  if(operation) {
-    operation.labels = { ...(operation.labels || {}), ...labels };
+export function withLabels<T>(operation: Operation<T>, labels: Labels): OperationObject<T> {
```
wKich: @cowboyd  `withLabels` start returning `OperationObject` instead of just `Operation` is it ok?

packages/core/src/symbol.ts
wKich: I change it a little bit. Is it correct now?

packages/core/test/labels.test.ts
```diff
@@ -2,26 +2,31 @@ import './setup';
...
+          originalOperation,
           { foo: "foo", bar: "bar" }
         ),
         { foo: "quox", blah: "blah" }
       );
 
-      await expect(someOperation?.labels).toEqual({ foo: "quox", bar: "bar", blah: "blah" });
+      expect(originalOperation).toHaveProperty("labels.spam", "spam");
+      expect(originalOperation).not.toHaveProperty("labels.bar", "bar");
+      expect(someOperation).toEqual({ foo: "quox", bar: "bar", blah: "blah", spam: "spam", [Symbol.operation]: originalOperation });
```
wKich: @cowboyd I added some assertion those describe that `withLabels` doesn't override labels

packages/fetch/test/fetch.test.ts
```diff
@@ -116,8 +116,8 @@ describe("fetch in node", () => {
     });
 
     it("applies labels", function*() {
-      expect(fetch('http://www.example.com').json()?.labels?.name).toEqual('fetch(\'http://www.example.com\').json()');
-      expect(fetch('http://www.example.com', { method: 'POST' }).json()?.labels?.method).toEqual('POST');
+      expect(fetch('http://www.example.com').json()?.name).toEqual('fetch(\'http://www.example.com\').json()');
+      expect(fetch('http://www.example.com', { method: 'POST' }).json()?.method).toEqual('POST');
```
wKich: @cowboyd because `withLabels` API was changed that test fails, so I decided to export `OperationObject` from `effection` and use it for `Fetch` interface declaration

packages/process/test/exec.test.ts
```diff
@@ -24,7 +24,7 @@ describe('exec', () => {
     });
 
     it('applies labels', function*() {
-      expect(exec("foo").join()?.labels?.name).toEqual('exec("foo").join()');
+      expect(exec("foo").join()?.name).toEqual('exec("foo").join()');
```
wKich: @cowboyd the same as for `fetch` tests is here

packages/process/test/exec.test.ts
```diff
@@ -72,8 +72,8 @@ describe('exec', () => {
           let error: unknown;
           let proc = yield exec("argle", { arguments: ['bargle'] });
           try {
-            yield proc.expect()
-          } catch (e) { error = e; }
+            yield proc.expect();
+          } catch (e) { error = e }
```
wKich: I got eslint errors here, so I decided to fix them

packages/core/src/symbol.ts
cowboyd: That looks right to me.

packages/core/src/operation.ts
```diff
@@ -53,10 +50,10 @@ export interface OperationObject<TOut> extends Record<string, unknown> {
...
-  OperationObject<TOut> |
-  Resource<TOut> |
-  undefined;
+  | OperationPromise<TOut>
+  | OperationGenerator<TOut>
+  | OperationFuture<TOut>
+  | OperationFunction<TOut>
+  | OperationObject<TOut>
+  | Resource<TOut>
+  | undefined;
```
cowboyd: formatting changes should happen in separate PRs

packages/core/src/controller/iterator-controller.ts
```diff
@@ -12,7 +12,7 @@ interface Claimable {
   [claimed]?: boolean;
 }
 
-export function createIteratorController<TOut>(task: Task<TOut>, iterator: OperationIterator<TOut> & Claimable, options: Options): Controller<TOut> {
+export function createIteratorController<TOut>(task: Task<TOut>, iterator: OperationGenerator<TOut> & Claimable, options: Options): Controller<TOut> {
```
cowboyd: This change makes sense, but is not in the scope of this PR.

packages/core/src/labels.ts
cowboyd: You are right. Even though this might be better, it is a breaking API change and cannot be added to 2.0

`withLabels()` must have the exact same behavior

packages/core/src/controller/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { Labels } from "../labels";
```
cowboyd: Let's just call this file `extract-labels.ts` 

"utils" is kind like drawer I have in my house where I keep, screw drivers, chewing gum, glue, and the keys to lock I lost 3 years ago, but am convinced I will find.

packages/core/src/index.ts
cowboyd: I think we want to discourage using any of the sub-operation types (with the possible exception of Resource). However, in the same way that `Iterable` and `Iterator` are different, we may want to export the `OperationObject` interface as `Operator`. For the time being, let's _not_ export it.

packages/core/src/labels.ts
cowboyd: As discussed, let's keep the mutative behavior. In `v3` we will have `withLabels` just return an OperationObject. This will also allow us to get rid of the `Labelled` interface entirely. But for now, we have to keep it as is.

Of course, since OperationObject is a new api and does not have existing behavior, we can return a new version of it. Although there is little point in passing an object to `withLabels` since you could just create the delegation yourself. In fact, we'll probably deprecate `withLabels` entirely in `v3` and remove it in `v4`.

packages/core/src/labels.ts
```diff
@@ -15,19 +18,30 @@ export type Labels = Record<string, string | number | boolean>;
...
-export function setLabels<T>(operation: Operation<T>, labels: Labels): Operation<T> {
-  if(operation) {
-    operation.labels = labels;
-  }
-  return operation;
+export function setLabels<T>(operation: Operation<T>, labels: Labels): OperationObject<T> {
+  return {
+    ...labels,
+    [Symbol.operation]: isObjectOperation<T>(operation) ? operation[Symbol.operation] : operation
+  };
```
cowboyd: As discussed, we should keep the old behavior.

packages/core/src/operation.ts
```diff
@@ -40,6 +41,10 @@ export interface OperationFunction<TOut> extends Labelled {
   (task: Task<TOut>): Operation<TOut>;
 }
 
+export interface OperationObject<TOut> extends Record<string, unknown> {
```
cowboyd: Do we need to include the other valid key types? I don't know. At least it seems ok in the playground https://www.typescriptlang.org/play?ts=4.5.4#code/MYewdgzgLgBCAOBTATgQygS3ALhgVzAwEc9EYIBPAWwCMQAbGAXhgGVq76A6AMxGQAUAcgQp0-IQEoA3AChZUCkhgB5JGkzgAPABUAfMxgDJzAwHFEYMVH5a11rGC2owFPQBod7l27myMYFAoPKjAZPYatvowiAAeQWAAJhAwAEqIoMiJWtDIAQDm7vhgANZgIADuYAYA3rIwDTAA2qIajgC6uBHojrp6cgC+8qCQsLFd6uLIWgBuIBiJBix1jTAAFoj09CC4QhX89IlCRY31jQBULZOaYO3GMDUD7mcNl+y0DFwYQZHIdyaPWQDIA

```suggestion
export interface OperationObject<TOut> extends Record<string|symbol|number, unknown> {
```

packages/main/src/browser.ts
```diff
@@ -3,13 +3,16 @@ import { isMainError } from './error';
 
 export * from './error';
 
+const DEFAULT_NAME = 'entry point';
+
```
cowboyd: I don't think having a constant here adds any clarity

packages/main/src/node.ts
```diff
@@ -4,6 +4,8 @@ import { isMainError } from './error';
 
 export * from './error';
 
+const DEFAULT_NAME = 'entry point';
+
```
cowboyd: I don't think having a constant here adds any clarity

packages/process/test/exec.test.ts
cowboyd: We should do all eslint fixes in a single PR that is just eslint fixes. Also, why isn't our CI catching these?

packages/core/src/controller/object-controller.ts
```diff
@@ -0,0 +1,51 @@
+import { Task } from '../task';
+import type { Controller } from './controller';
+import { createFuture } from '../future';
+import { OperationObject } from '../operation';
+import { isObjectOperation } from '../predicates';
+import { extractLabels } from './utils';
+
+export function createObjectController<TOut>(task: Task<TOut>, obj: OperationObject<TOut>, createController: () => Controller<TOut>): Controller<TOut> {
```
cowboyd: `ObjectController` and `FunctionController` are basically the same? Can we make a `DelegateController`?

packages/main/src/browser.ts
wKich: I extract it to the constant, because value is used twice.

packages/core/src/labels.ts
```diff
@@ -26,8 +34,29 @@ export function withLabels<T>(operation: Operation<T>, labels: Labels): Operatio
...
+    } as Operation<T>;
+  } else if (operation) {
     operation.labels = labels;
   }
   return operation;
 }
+
+export function extractLabels<T>(operation: Operation<T>): Labels | undefined {
+  if (!isObjectOperation<T>(operation)) {
+    return operation?.labels;
```
cowboyd: Do we need to also treat the name as a special case here?

packages/core/src/operation.ts
```diff
@@ -40,6 +41,10 @@ export interface OperationFunction<TOut> extends Labelled {
   (task: Task<TOut>): Operation<TOut>;
 }
 
+export interface OperationObject<TOut> extends Record<string | symbol | number, unknown> {
```
cowboyd, line 44: @wKich I was more raising the question than saying what the answer was. Both `Record<string, unknown>` and `Record<string|symbol|number,unknown>` seem to work. Question is which is more forward compatible? Do you have an opinion?

cowboyd (commented):
We're getting close.

packages/core/src/labels.ts
cowboyd: In in other words if this is not an `ObjectOperation` then both `name` and `labels` are fields that contain labels.

packages/core/src/operation.ts
wKich: I think `Record<string|symbol|number,unknown>` is better, but also you will not have numbers as a keys, using symbols is pretty rare case. But at least `string|symbol|number` is the most wider type that supported by `Record`

image

packages/fetch/src/fetch.ts
```diff
@@ -1,4 +1,4 @@
-import { AbortSignal, createAbortSignal, withLabels, Operation, Resource } from '@effection/core';
+import { AbortSignal, createAbortSignal, withLabels, OperationObject, Resource } from '@effection/core';
```
cowboyd: `OperationObject` is a private type, so none of our public interfaces should return it explicitly. Let's revert this back to `Operation`

packages/process/src/exec.ts
```diff
@@ -10,8 +10,8 @@ export * from './exec/api';
 export * from './exec/error';
 
 export interface Exec extends Resource<Process> {
-  join(): Operation<ProcessResult>;
-  expect(): Operation<ProcessResult>;
+  join(): OperationObject<ProcessResult>;
+  expect(): OperationObject<ProcessResult>;
```
cowboyd:
```suggestion
  join(): Operation<ProcessResult>;
  expect(): Operation<ProcessResult>;
```

cowboyd (dismissed):
We've got some places where we're having a public API returning `OperationObject` and not `Operation`. Wherever possible, and I think it is everywhere, we should use the opaque `Operation` return type. There are some cases where we currently are returning `Resource` which it is too late to change, but for all new APIs, let's use `Operation<T>`

packages/fetch/src/fetch.ts
wKich: @cowboyd Looks like your comment targets to one of old commits. The actual version of PR doesn't export `OperationObject` from `core` nor uses it in `fetch` and `exec`.

packages/core/src/operation.ts
cowboyd: 👍🏻

packages/inspect-utils/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@effection/atom": "2.0.2",
-    "@effection/dispatch": "2.0.1",
+    "@effection/dispatch": "2.0.2",
```
cowboyd: Is there a reason to bump this?

packages/core/src/labels.ts
cowboyd: @wKich ☝🏻

packages/inspect-utils/package.json
wKich: For this PR, no. But it fixes a problem that `inspect-utils` uses old version, not the same from the monorepo. I can move that change in separate PR

frontsidejack:
The following preview packages have been published by Covector:
- `@effection/channel@2.0.2-add-operation-object.1643201536852`
- `@effection/duplex-channel@2.0.3-add-operation-object.1643201536852`
- `@effection/dispatch@2.0.3-add-operation-object.1643201536852`
- `@effection/atom@2.0.3-add-operation-object.1643201536852`
- `@effection/core@2.1.0-add-operation-object.1643201536852`
- `effection@2.0.3-add-operation-object.1643201536852`
- `@effection/events@2.0.2-add-operation-object.1643201536852`
- `@effection/fetch@2.0.3-add-operation-object.1643201536852`
- `@effection/inspect@2.1.3-add-operation-object.1643201536852`
- `@effection/inspect-server@2.1.3-add-operation-object.1643201536852`
- `@effection/inspect-utils@2.1.3-add-operation-object.1643201536852`
- `@effection/main@2.0.2-add-operation-object.1643201536852`
- `@effection/mocha@2.0.3-add-operation-object.1643201536852`
- `@effection/process@2.0.3-add-operation-object.1643201536852`
- `@effection/react@2.1.2-add-operation-object.1643201536852`
- `@effection/subscription@2.0.2-add-operation-object.1643201536852`
- `@effection/stream@2.0.2-add-operation-object.1643201536852`
- `@effection/websocket-client@2.0.3-add-operation-object.1643201536852`
- `@effection/websocket-server@2.0.3-add-operation-object.1643201536852`
- `@effection/inspect-ui@2.1.3-add-operation-object.1643201536852`

<p align="right">c623a84</p>

packages/fetch/src/fetch.ts
```diff
@@ -1,4 +1,4 @@
-import { AbortSignal, createAbortSignal, withLabels, Resource, Operation } from '@effection/core';
+import { AbortSignal, createAbortSignal, withLabels, Operation, Resource } from '@effection/core';
```
cowboyd: What does this change accomplish?

packages/process/src/exec.ts
```diff
@@ -1,7 +1,7 @@
 /// <reference types="../types/shellwords" />
 import { split } from 'shellwords';
 
-import { Task, Resource, spawn, withLabels, Operation } from 'effection';
+import { Task, Operation, Resource, spawn, withLabels } from 'effection';
```
cowboyd: What does this change accomplish?

packages/fetch/src/fetch.ts
wKich: Take a look the diff of whole PR. I've just reverted my old changes

packages/process/src/exec.ts
wKich: The same as for `fetch.ts`

.changes/add-operation-object.md
```diff
@@ -0,0 +1,11 @@
+---
+"@effection/fetch": "patch"
```
cowboyd: Only the test case changed, right?

```suggestion
```

wKich: Yep. Do I need remove `patch` for other packages, where I changed only devDeps, too?

cowboyd: I think for the typescript, it is ok to leave the patch in since it could _theoretically_ cause the package to be compiled differently thereby changing what JS gets shipped to NPM

cowboyd (approved):
Looks good!

PR #604, opened 2022-01-18
rename effection's `OperationIterator` to `OperationGenerator` to avoid naming issues with another `OperationIterator`

wKich:
## Motivation

There are two different `OperationIterator` types. The one is in core package and another is in subscription package. https://github.com/thefrontside/effection/blob/41101923824849238a462969e5be3945f2fcc139/packages/subscription/src/operation-iterator.ts#L6

## Approach

That one from core package is actually extends from `Generator` so I decided to rename it to `OperationGenerator`

### DIFF
--- a/packages/core/src/controller/iterator-controller.ts
+++ b/packages/core/src/controller/iterator-controller.ts
@@ -1,5 +1,5 @@
 import { Controller, Options } from './controller';
-import { OperationIterator } from '../operation';
+import { OperationGenerator } from '../operation';
 import { createTask, Task } from '../task';
 import { Operation } from '../operation';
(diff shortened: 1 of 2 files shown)


cowboyd (commented):
While this is technically sound, I think I'd like to hold off on removing this API and perhaps creating an alias to it and then deprecating the old one.

cowboyd (approved):
In theory, this is a 100% semver-safe change, so I'm going to approve. I can't see how it could possibly break, but I've been surprised before, so I guess we'll find out :)

cowboyd:
@wKich I think we forgot the changeset

PR #608, opened 2022-01-24
Upgrade `cross-fetch` and remove `node-fetch`

### LINKED ISSUE #607: Update @effection/fetch to depend on `cross-fetch ^3.1.5`
There's a SEC issue with the version of `cross-fetch` we're using https://github.com/lquixada/cross-fetch/pull/124

We need to upgrade. At the same time, there is no need for us to depend directly on `node-fetch` since that is specified in the `cross-fetch` dependency.

minkimcello:
## Motivation

To resolve #607

## Approach

- Upgrade `cross-fetch` to `3.1.5`
- Remove `node-fetch`
- Update `@effection/process` tests
- Added change file for a patch bump of `@effection/fetch` only as the changes for `@effection/process` in dev env only

### DIFF
--- a/packages/process/test/daemon.test.ts
+++ b/packages/process/test/daemon.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, beforeEach, afterEach } from '@effection/mocha';
+import { describe, it, beforeEach, afterEach, captureError } from '@effection/mocha';
 import expect from 'expect';
 
 import { run, Task, createFuture, fetch, spawn } from 'effection';
@@ -42,13 +42,7 @@ describe('daemon', () => {
       task.halt();
     });
     it('kills the process', function*() {
-      let error;
-      try {
-        yield fetch(`http://localhost:29000`, { method: "POST", body: "hello" })
-      } catch (err) {
-        error = err;
-      }
-      expect(error).toHaveProperty('name', 'FetchError');
+      expect(yield captureError(fetch(`http://localhost:29000`, { method: "POST", body: "hello" }))).toHaveProperty('name', 'FetchError');
     });
   });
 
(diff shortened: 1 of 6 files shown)


packages/process/test/daemon.test.ts
```diff
@@ -42,9 +41,14 @@ describe('daemon', () => {
...
-
     it('kills the process', function*() {
-      yield expect(fetch(`http://localhost:29000`, { method: "POST", body: "hello" })).rejects.toHaveProperty('name', 'FetchError');
+      let error;
+      try {
+        yield fetch(`http://localhost:29000`, { method: "POST", body: "hello" })
+      } catch (err) {
+        error = err;
+      }
+      expect(error).toHaveProperty('name', 'FetchError');
```
cowboyd: Any reason this wasn't working the same as before?

minkimcello: It returns the `Fetch` interface and not the response without the `yield`

cowboyd: Ah, I see we're using the `Effection` fetch here. If we're going to treat the error as a value, then let's just capture it as a return value instead of re-assigning a variable.

```ts
let error = yield function*() {
  try {
    yield fetch(`http://localhost:29000`, { method: "POST", body: "hello" });
  } catch (error) {
    return error;
  }
}
expect(error).toMatchObject({ name: 'FetchError' });
```

This is more composable (decoupled from context) since it can be abstracted to handle any operation, not just a `fetch`. 

```ts
function* capture<T>(operation: Operation<T>): Operation<Error | T> {
  try {
    return yield operation;
  } catch (error) {
    return error;
  }
```

Which we can then use as 

```ts
expect(yield capture(fetch(`http://localhost:29000`, { method: "POST", body: "hello" }))).toMatchObject({ name: 'FetchError' });
```

In fact, I think there might be already a `captureError` operation in `@effection/mocha`

minkimcello: 👍 https://github.com/thefrontside/effection/pull/608/commits/cd7698e2dc3d8683fe55ce9f618777a9daa7f9b2

cowboyd (approved):
Looks great!

PR #609, opened 2022-01-25
Use `@effection/dispatch` for inspector utils from monorepo

wKich:
## Motivation

Version of `@effection/dispatch` isn't the same as used in monorepo. That's why if we do some changes in `dispatch` package we will not be able to test that changes in inspector-utils

## Approach

Update `@effection/dispatch` to actual version


cowboyd (commented):
Looks good, although we need a changeset.

packages/inspect-utils/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@effection/atom": "2.0.3",
-    "@effection/dispatch": "2.0.1",
+    "@effection/dispatch": "2.0.3",
```
cowboyd, line 32: This is another example of where we need `convector` to support `^` dependencies.

@jbolda Can you point us to roughly where we would need to make the change?

cowboyd:
I'm going to close this for now as I think it won't be needed for v3

PR #612, opened 2022-01-26
Standardize build tools for all packages

cowboyd:
## Motivation

While trying to build a `@effection/jest` integration, I was getting unrelated yarn errors because of an old version of parcel being incompatible with a certain yarn version. We need to upgrade parcel in inspect-ui, but first we need to make sure that we're using the name node and yarn everywhere.

## Approach
It turns out that there were several packages that were using their own volta configs. This points all of them at the same config from the base of the repo

Note that there is no changeset, since this only effects which version of yarn is used to install a subset of packages, and which version of node is used to run repository management.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "volta": {
     "node": "14.17.1",
-    "yarn": "1.22.4"
+    "yarn": "1.22.17"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


wKich (approved):
One version for all package is always good :+1:

PR #613, opened 2022-01-26
Add @effection/jest package

cowboyd:
## Motivation

[Jest](http://jestjs.io) is a popular framework used on many projects, including backstage, and so in order to use Effection to manage complex asynchrony inside a test case, we need to have an official effection integration.

## Approach

This uses the same techniques as in `@effection/mocha`, and most of the same test cases. However, I stopped short of trying to extract a common implementation in order that we can see what works and what doesn't between the two frameworks. I'm hoping that there can be a little bit of back and forth about the best way to do things, before we settle on the "one true way" (If we ever do)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
       "packages/fetch",
       "packages/effection",
       "packages/mocha",
+      "packages/jest",
       "packages/dispatch",
       "packages/duplex-channel",
       "packages/atom",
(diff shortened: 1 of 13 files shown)


taras (commented):
@cowboyd can you write docs for this on frontside.com now while it's fresh in your mind?

cowboyd:
@taras The documentation needs to be written for mocha too, so I was planning on doing that in a separate PR. There are actually a couple of documentation PRs that need to happen

* rethink our testing documentation
* We need to document our `Symbol.operation`

taras (approved):
👍

PR #617, opened 2022-01-28
Create totally independent scopes for before{All/Each}()

cowboyd:
## Motivation
It turns out that you don't actually want to have the `beforeEach` scope to descend from the `beforeAll` because if we have a failure in that scope, the parent needs to outlive it so that it is around when the second test runs. As it is now, if you have two testcases:

+ before all
   -> it does one thing
   -> it does another thing

If the first test fails, it will kill the `beforeAll` scope because it is a child and so the error propagates to the parent.

## Approach
This makes breaks that problematic error propagation by making the suite scope and the test scope *independent siblings* so that a test can fail without crashing the suite.

This also adds some checks to make sure both before and after a test runs that both scopes are in good order and if they are failing, the error is raised.

### DIFF
--- a/.changes/independent-test-scopes.md
+++ b/.changes/independent-test-scopes.md
@@ -1,0 +1,4 @@
+---
+"@effection/jest": patch
+---
+test suite and each test get independent sibling scopes.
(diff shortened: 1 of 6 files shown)


taras (approved):
👍

PR #619, opened 2022-02-03
Don't run each `it.eventually()` trial directly in the test's "each" scope

cowboyd:
## Motivation

`it.eventually()` runs the test operation over and over again until it either passes, or times out. However, we were running each trial directly in the test's "each" scope, so if the trial failed, then it crashed the test's scope which then caused the test to crash, even though we were catching the error inside the eventually loop.

## Approach
This just calls the trial as a normal operation whithin the scope of the current task, that way it can either catch or propagate using the normal rules of nested tasks.

### DIFF
--- a/.changes/dont-run-eventually-in-each-scope.md
+++ b/.changes/dont-run-eventually-in-each-scope.md
@@ -1,0 +1,4 @@
+---
+"@effection/jest": patch
+---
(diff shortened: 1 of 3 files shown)


taras (approved):
👍

wKich (approved):
Looks good

PR #621, opened 2022-02-07
Run resolved value from promise as operation

wKich:
## Motivation

I this PR https://github.com/thefrontside/interactors/pull/166 effection was added to interactors to able execute interactions as operations. But there is one problem. BigTest adds its own interaction wrapper [here](https://github.com/thefrontside/bigtest/blob/v0/packages/agent/app/run-lane.ts#L117-L123). That wrapper is async function. The original action function executes `converge`, which returns a generator. The problem is when we `run` that BigTest wrapper returns a promise that resolves in `converge` generator. But effection handle that `generator` as a end result value and `converge` function is never been executed.

## Approach

Handle resolved value as operation.

### Alternate Designs

- Rewrite interaction wrappers to accept only generators

There is something similar implementation for `FutureOperation` that maybe also should be fixed in the same manner.

### TODOs and Open Questions

I noticed that any `falsy` value treated for suspending task. So if `FunctionOperation` returns falsy value it means task will be suspended. But there are possible scenarios where resolved value from `PromiseOperation` could be `0` or `null` or `false` even empty string and it's could be a valid value. For `FunctionOperation` that behavior is different.

### DIFF
--- a/packages/core/src/controller/controller.ts
+++ b/packages/core/src/controller/controller.ts
@@ -10,6 +10,7 @@ import { createFutureController } from './future-controller';
 import { createResourceController } from './resource-controller';
 import { Future } from '../future';
 import { Symbol } from '../symbol';
+import { UnknownOperationTypeError } from '../error';
 
 export interface Controller<TOut> {
   type: string;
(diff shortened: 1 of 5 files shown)


cowboyd:
For better or for worse, this is a breaking change that could only accompany a major semver release. So I think we need to explore other workaround for the time being.

wKich:
Let's leave things as they are.

PR #628, opened 2022-03-29
process env var inline

jbolda:
## Motivation

Some CLI commands come with the expectation that you can pass env vars to the command via `cli --env $ENV`. This has the inherent expectation that you are using a bash-like shell to run the command. On windows, the fallback is `cmd.exe` which cannot handle this. This PR adds the ability to specify a string which opens up the ability to specify `bash` on Windows, in which case this functionality works.

## Approach

Allow strings as input and add a bunch of tests to better represent the expectations behind the various options.

### Alternate Designs

We discussed possibly having "presets" where you can opt into "bash-like" syntax with additional caveats over the default, but this opens up a bit more manual use which is also something that may be helpful.


packages/process/test/exec.test.ts
```diff
@@ -158,26 +181,166 @@ describe('exec', () => {
...
+
+        expect(stdout).toEqual('boop\n');
+        expect(code).toBe(0);
+      });
+    });
+
+    describe('env as option - shell: true', () => {
+      let shell = true;
+
+      it('echo env', function* () {
```
cowboyd: tests should read as complete sentences starting with the word `it` e.g. "it can echo a passed in environment variable"

packages/process/test/exec.test.ts
```diff
@@ -158,26 +181,166 @@ describe('exec', () => {
...
+    it('automatically parses the command arguments using shellwords', function* () {
       let proc = exec('echo "first" | echo "second"', {
-        shell: false
+        shell: false,
       });
       let { stdout }: ProcessResult = yield proc.expect();
 
-      expect(stdout).toEqual("first | echo second\n");
-    })
+      expect(stdout).toEqual('first | echo second\n');
```
cowboyd: It looks like there are a bunch of un-related formatting changes in this PR. If we want to add automated code-formatters, that should be done as part of a separate PR.

cowboyd (commented):
Looks good to me. The shell-as-string works in linux as well... or at least I assume it does because it is supported by the underlying node ChildProcess

cowboyd (approved):
Looks good. Do we need to make any changes to documentation?

jbolda:
> Looks good. Do we need to make any changes to documentation?

I think this is probably fine just as part of the API docs. I think that all just happens automatically now, eh?

cowboyd:
> > Looks good. Do we need to make any changes to documentation?
> 
> I think this is probably fine just as part of the API docs. I think that all just happens automatically now, eh?

Agree, and yes, it does run the docs upon release, but we've changed the API from `boolean` -> `string | boolean`, so we should probably update the docs according.

cowboyd:
🎉

PR #635, opened 2022-05-03
✨Build inspect-ui with stable parcel

cowboyd:
## Motivation

The ESM entry points don't exist 😱 so when trying to include the inspect ui into another UI (in this case Backstage) it just gets a bunch of `ENOENT` errors , so obviously this needs to be fixed. However, parcel is in the way by using the "browser" field of package.json. When you import from `@effection/inspect-ui` in a React app compiled with webpack, you don't want to require from `index.html`, you want to get at the actual ESM bundled code. The first step then, is to move `app/index.html` away from "browser" field and just build it to a `dist-app` directory that can contain the entire app. Then, we can build the JavaScript code to `dist-esm` and actually make imports from there work.

## Approach

Since this UI was written, the parcel bundler has been improved and stabalized a lot 🎉 This just upgrades to the stable version, removes patches to the typescript transform, and uses the "app" target.
build it.

### DIFF
--- a/.changes/upgrade-parcel.md
+++ b/.changes/upgrade-parcel.md
@@ -1,0 +1,4 @@
+---
+"@effection/inspect-ui": patch
+---
+Upgrade parcel build to public release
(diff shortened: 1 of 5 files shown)


jnicklas (approved):
LGTM!

PR #637, opened 2022-05-03
Add ESM Build of `@effection/inspect-ui` React Components

cowboyd:
## Motivation
I'm embeddding the Effection Inspector into the Backstage UI, and so I need the ability to access the inspector-ui components directly. In other words:

```ts
import { App as InspectorApp } from '@effection/inspect-ui';
```

However, the "browser" and "module" entry points don't actually point to anything, because the inspector ui was only ever shipped as an independent application up to this point.

## Approach
This compiles all the components into `dist-esm` and copies all of the CSS and SVG assets there too. That way, they can be consumed by 3rd party application.


taras (approved):
👍

PR #638, opened 2022-05-03
Make React a flexible peer dependency

cowboyd:
## Motivation
React is very finicky about having more than one copy of the react code itself in a single app

https://reactjs.org/warnings/invalid-hook-call-warning.html

Having a hard dependency on React 17 means then that a local copy of React 17 will be installed in this package and that it will be incompatible with React 16 apps.

## Approach
This changes the dependency on react to a peer dependency, and furthermore allows it to work on react 16 versions that have hooks, and version 17.


taras (approved):
👍

PR #640, opened 2022-05-06
Upgraded to inspector-ui to react-router@6.3.0

taras:
## Motivation

I'm in the process of upgrading inspector-ui to make it embeddable into Backstage. One of the problems that we observed in our integration is that navigating to index page caused the browser to go to root. For example, going from `/inspector/tasks/1`, navigated to `/` instead of `/inspector`. This was caused by the fact that routing was not scoped within the route. We can fix that with relative navigation and nested routes introduced in React Router 6.0.

## Approach

1. Upgraded to react-router@6.3.0. 
2. Defined one root route with 2 child routes as a route object
3. Replaced `<Switch>` with `useRoutes`
4. Refactored to use `element` which allow to eliminate unnecessary wrapper components
5. Refactored to use `Navigate` instead of `Redirect`
6. Created TaskTreePage & TaskPage 

[React Router 6.3 compatibility](https://github.com/backstage/backstage/pull/10450) hasn't been merged yet. It maybe problematic, but I think it should be ok. 

## Learning

[Upgrading from React Router 5](https://reactrouter.com/docs/en/v6/upgrading/v5)

### DIFF
--- a/packages/inspect-ui/app/app.tsx
+++ b/packages/inspect-ui/app/app.tsx
@@ -4,7 +4,7 @@ import { Slice } from '@effection/atom';
 import { InspectState } from '@effection/inspect-utils';
 import { SettingsMenu, SettingsContext, DEFAULT_SETTINGS } from './settings';
 import { TaskPage } from './task-page';
-import { TaskTreeRoot } from './task-tree-root';
+import { TaskTreePage } from './task-tree-page';
 
 export type InspectStateSlice = Slice<InspectState>;
 
@@ -18,7 +18,7 @@ export function App({ slice }: AppProps): JSX.Element {
       children: [
         {
           index: true,
-          element: <TaskTreeRoot slice={slice} />,
+          element: <TaskTreePage slice={slice} />,
         },
         {
           path: "tasks/:id",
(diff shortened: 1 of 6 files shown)


cowboyd:
What happens when we embed into an app that does not use react router at all?

taras:
That has same behaviour as before. It just works

cowboyd:
> That has same behaviour as before. It just works

Right, but what if the embedding app has its own routing scheme? I'd imagine that this would mess with it.

cowboyd:
@taras Sorry for asking the questions piecemeal. Let me try and lay out everything that I'm thinking about. What happens if:

1. embedded into a `react-router@v5` app?
2. embedded into an app without react router at all?
  * will this hijack the url?
  * Is there a compatibility layer you can wrap around it to adapt to your own routing scheme?

We don't have to do this now, but what would it look like to separate the routing and the components?

cowboyd:
I think we may need to increase the timeout on windows, as I've started to see this fail intermittently.

cowboyd (approved):

taras:
> embedded into a react-router@v5 app?

I'm not sure if we need to worry about this use case. `react-router@6` has so many nice things that there is no reason to stay on v5 unless you're stuck. 

> embedded into an app without react router at all?

Which router is used is controlled by the router that is rendered in `main.tsx`. We'd need to do some organizing to make it easier to embed the entire thing.

PR #641, opened 2022-05-09
Add useResource hook.

cowboyd:
## Motivation

When embedding the inspector into another react app, we need to load a slice of data that is continuously updated from the server. This data is loaded from either a `WebSocket` or an `EventSource` and so it takes some time to become available as the server and client negotiate could fail at any point. Once this has been negotiated, and only then, is the stream set up and the subscription can be used to drive the inspector. It turns out that it is helpful to model this very Effectionate process directly in the state

## Approach
This adds a hook callled `useResource()` that can be used to consume a resource directly from within a React Component. Resources can either be `pending`, `resolved`, or `rejected` and as a result, the handle returned by a resource does not let you access the resource directly, but via a `Maybe`-like interface called `ResourceHandle`.

```tsx
  const handle = useResource(operation, [dep1, dep2]);

  if (handle.type === 'pending') {
    return <p>Loading</p>;
  } else if (handle.type === 'rejected') {
    return <p>{handle.error.toString()}</p>;
  } else {
    return <MyComponent value={handle.value}/>
  }
```

It's interestig to note that `useResource` can be used with _any_ `Operation` type, and _not_ just `Resource`. This is because normal operations also provide a value asynchronously. The only difference is that they are completely done and do not have any on-going computation associated with them. The following is totaly valid:

```tsx
  const handle = useResource(Promise.resolve('hello'), [dep1, dep2]);

  if (handle.type === 'pending') {
    return <p>Loading</p>;
  } else if (handle.type === 'rejected') {
    return <p>{handle.error.toString()}</p>;
  } else {
    return <MyComponent value={handle.value}/>
  }
```

### DIFF
--- a/.changes/react-use-resource.md
+++ b/.changes/react-use-resource.md
@@ -1,0 +1,4 @@
+---
+"@effection/react": minor
+---
+introduce the `useResource()` hook for working directly with resources.
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -2,4 +2,5 @@ export { useStream } from './use-stream';
 export { useSubscription } from './use-subscription';
 export { useOperation } from './use-operation';
 export { useSlice } from './use-slice';
+export { ResourceHandle, useResource } from './use-resource';
 export { EffectionContext } from './context';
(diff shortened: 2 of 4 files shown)


taras (approved):
👍 we should add Suspense integration after this

PR #645, opened 2022-05-10
Fix inspect UI package

taras:
## Motivation

Earlier today, @cowboyd tried to include Inspect UI in Backstage, but couldn't because it was missing the typings file. I wanted to fix this by making sure that typings file was generated and included with the bundle.

## Approach

1. I changed the `@effection/inspect-ui` package to primarily serve as a provider of the `dist-esm` and `dist-app` directories by removing `index.js` and `index.d.ts` from the package
2. `@effection/inspect-server` now attempts to find `@effection/inspect-ui/dist-app/index.html` and complains if it doesn't find it.
3. I changed `main` entry point to point to `app/index.ts` This is unlikely to be used any bundler because it should be using `module` entry point to bet the browser bundle
4. Changed `types` to point to `dist-esm/index.d.ts` which is now generated from `app/index.ts`

## Verified

I used preview package `@effection/inspect-ui@2.1.5-tm-fix-inspect-ui-package.1652216146800` to verify that it embeds properly into Backstage.


packages/inspect-server/src/server.ts
```diff
@@ -17,6 +19,15 @@ export interface InspectServer {
   port: number;
 }
 
+function appDir() {
+  let index = require.resolve("@effection/inspect-ui/dist-app/index.html");
+  assert(existsSync(index), `the @effection/inspect-ui app has gone missing!
+Expected ${index} to exist, but it did not.
```
cowboyd, line 25: I'm a little bit worried about breaking the encapsulation between packages here because it means that the server package cannot be re-organized freely without breaking packages that consume it. This is a minor annoyance for the inspect server since it is part of the same mono-repo and the tests should start failing if we change the location of the built app. However, this will also break dependent packages that want to serve the inspector app without necessarily using the inspect server. Is there a reason why this function cannot live in the `@effection/inspect-ui` package?

packages/inspect-ui/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@effection/inspect-ui",
   "version": "2.1.5",
   "description": "Web interface for inspecting effection applications",
-  "main": "index.js",
-  "types": "index.d.ts",
+  "main": "app/index.ts",
+  "types": "dist-esm/index.d.ts",
```
cowboyd, line 6: Why not create a `dist-cjs` directory just like all our other packages?

cowboyd (approved):
Technically, this is a semver breaking change, but given that making it work would require us to somehow piling hacks onto the `inspect-ui` entry points, I think it's ok to just break the interface.

What this really tells us is that the parcel `app` itself should actually live in the `inspect-server` package entirely, and that the `inspect-ui` package should just be the components.

I don't see third party consumers of the app as being very likely, and in that event we can always export the `appDir()` function from the `inspect-server`.

PR #646, opened 2022-05-12
Embrace "structured concurrency" in tagline

cowboyd:
## Motivation
Originally, we had shied away from using the term "structured concurrency" directly since it was not that well known, a bit academic sounding and we wanted to appeal to a practical audience. However, now that it is becoming more mainstream with the advent of structured concurrent systems in Java and Swift, it is much more recognizable and people will be explicitly shopping for structured concurrency. 

We should state explicitly that yes, this is structured concurrency (and effects) so that folks searching for  those things will find us and know that it's what we're looking for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 # Effection
 
-A JavaScript framework that makes building concurrent systems easy to get right.
+Structured concurrency and effects for JavaScript.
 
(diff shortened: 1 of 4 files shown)


taras (approved):
YESSSHH

PR #648, opened 2022-05-18
Add Erlang style supervisor

jnicklas:
Depends on #649 

## Motivation

One of the visions for Effection was to recreate part of the brilliant OTP system for Erlang, by enabling the easy creation of supervision trees. Supervision trees enable fault tolerance through the "just let it crash" philosophy. While Effection code is very good at surfacing errors, actually dealing with those errors has been tricky. With supervision trees, parts of the system or the whole system can be kept running and restarted easily.

## Approach

Why reinvent the wheel? OTPs supervisors are good and battle tested. This basically copies most of the functionality of OTP supervisors.

For more information, see the [README](https://github.com/thefrontside/effection/blob/supervisor/packages/supervisor/README.md)

### Alternate Designs

There are millions of other ways of doing the same thing.

The code is complicated, and there are probably a ton of nasty edge cases.

### TODOs and Open Questions

- [ ] This interacts somewhat poorly with resources. Is there something we can do about that?
- [x] Missing `simple_one_for_one` strategy. IMO this should probably be a completely separate thing, since the external interface for `simple_one_for_one` should be quite different. It feels like Erlang sort of shoehorned this into one module, but it really shouldn't be.
- [x] The `period` and `intensity` options are not yet implemented. I think this is a very necessary part of the supervision system, but it's tricky to implement. Also, I think having some kind of backoff would be nice.
- [x] Type safety for the `args` option could be improved. I think the only nice way of doing this is by using the multiple-overloads trick.

## Learning

https://www.erlang.org/doc/man/supervisor.html

### DIFF
--- a/packages/supervisor/README.md
+++ b/packages/supervisor/README.md
@@ -1,0 +1,79 @@
+# @effection/supervisor
+
+A task supervisor similar to Erlang's OTP supervisor module. Supervisors can
+monitor other tasks and restart them as needed.
+
+The functionality and options of the Effection supervisor are broadly similar
+to Erlang's.
+
+## Known limitations:
+
+Some functionality from Erlang's supervisors is as of yet still missing:
+
+- `simple_one_for_one` supervision strategy
+- the `intensity` and `period` options, which prevent restart loops
+
+Contributions for these are very welcome!
+
+## Usage
+
+This packages exposes two methods `runSupervisor` and `createSupervisor`.
+
+With `runSupervisor`, the supervisor runs in the foreground as a task, which
+makes it easy to start and keep it running, for example with `main`:
+
+```typescript
+import { main } from 'effection';
+import { runSupervisor } from '@effection/supervisor';
+import { fobrulator, phaseInverter } from './operations';
+
+main(runSupervisor([
+  { name: 'fobrulator', run: fobrulator, args: ['fob'] },
+  { name: 'phaseInverter', run: phaseInverter, args: [true] },
+]);
+```
+
+With `createSupervisor`, the supervisor runs in the background as a resource,
+which makes it possible to dynamically control the supervisor:
+
+```typescript
+import { main } from 'effection';
+import { runSupervisor } from '@effection/supervisor';
+import { fobrulator, phaseInverter } from './operations';
+
+main(function*() {
+  let supervisor = yield createSupervisor();
+
+  supervisor.addChild({ name: 'fobrulator', run: fobrulator, args: ['fob'] });
+  supervisor.addChild({ name: 'phaseInverter', run: phaseInverter, args: [true] });
+
+  yield;
+]);
+```
+
+## Options
+
+The supervisor can currently take the following options:
+
+- `strategy`, one of:
+  - `oneForOne`: If one child task terminates and is to be restarted, only that child process is affected. This is the default restart strategy.
+  - `oneForAll`: If one child task terminates and is to be restarted, all other child processes are terminated and then all child processes are restarted.
+  - `restForOne`: If one child task terminates and is to be restarted, the 'rest' of the child tasks (that is, the child tasks after the terminated child task in the start order) are terminated. Then the terminated child task and all child tasks after it are restarted.
+- `shutdown`, one of:
+  - `never`: Automic shutdown is disabled. This is the default setting.
+  - `anySignificant`: The supervisor will shut itself down when **any** significant child terminates, that is, when a transient significant child terminates normally or when a temporary significant child terminates normally or abnormally.
+  - `allSignificant`: The supervisor will shut itself down when **all** significant children have terminated, that is, when the **last active** significant child terminates.
+
+## Child specs
+
+The child specification can contain the following options:
+
+- `name` *required*: A name which identifies the child.
+- `run` *required*: A function which returns an operation.
+- `args`: a list of arguments to pass to the `run` function.
+- `type`: one of:
+  - `permanent`: The child is always restart when it terminates, regardless of the reason. Note that children explicitly halted through `haltChild` are not restarted.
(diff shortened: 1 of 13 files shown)


packages/supervisor/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@frontside/tsconfig": "^1.2.0",
+    "@types/mocha": "^7.0.1",
+    "@types/node": "^13.13.4",
+    "expect": "^25.4.0",
+    "mocha": "^8.3.1",
+    "ts-node": "^10.4.0",
+    "typescript": "^4.3.5"
+  },
+  "dependencies": {
+    "effection": "2.0.4"
```
cowboyd: Can we make this an `^` dependency?

packages/supervisor/README.md
```diff
@@ -0,0 +1,79 @@
...
+```typescript
+import { main } from 'effection';
+import { runSupervisor } from '@effection/supervisor';
+import { fobrulator, phaseInverter } from './operations';
+
+main(function*() {
+  let supervisor = yield createSupervisor();
+
+  supervisor.addChild({ name: 'fobrulator', run: fobrulator, args: ['fob'] });
+  supervisor.addChild({ name: 'phaseInverter', run: phaseInverter, args: [true] });
```
cowboyd, line 48: What is the value in passing the args in an array rather than just doing it in a callback?

```
supervisor.addChild(({ name: 'phaseInverter', run: () => phaseInverter(true) }))
```

cowboyd (dismissed):
This looks amazing! Sorry I've taken so long to get to reviewing it, but I didn't have much time last week.

I am curious about the dynamic nature of the `addChild`, `haltChild` and `handAndRemoveChild`. What is the modern use case for being able dynamically configure the supervisor like this as opposed to say, passing in the child specifications to the constructor?

Actually, in stating the question, I think I may have answered it. It would be awesome in Backstage to be able to enable and disable plugins via a UI, and this would be perfect for that sort of thing.

A thing like `addChild` would even allow you to add a new plugin without even restarting the server! 

🧘🏻‍♂️🧘🏻‍♂️🧘🏻‍♂️

packages/supervisor/README.md
jnicklas: That's a fair point! I basically just translated this from the Erlang version, but with anonymous closures, it is far less useful, as you say. Also, this makes it much easier to get the types right.

jnicklas:
> This looks amazing! Sorry I've taken so long to get to reviewing it, but I didn't have much time last week.
> 
> I am curious about the dynamic nature of the `addChild`, `haltChild` and `handAndRemoveChild`. What is the modern use case for being able dynamically configure the supervisor like this as opposed to say, passing in the child specifications to the constructor?
> 
> Actually, in stating the question, I think I may have answered it. It would be awesome in Backstage to be able to enable and disable plugins via a UI, and this would be perfect for that sort of thing.
> 
> A thing like `addChild` would even allow you to add a new plugin without even restarting the server!

I think, by far the most common use case is adding children in the constructor. I think for most usecases involving dynamically adding and removing children, you really want the `simple_one_for_one` type supervisor. But IMO in TypeScript, this supervisor should have a totally different interface. In Erlang some methods just change signature when this strategy is passed in, which makes (some) sense in a dynamic language, but becomes pretty cumbersome in a statically typed language. I do think there are *some* usecases for dynamically adding children, like you pointed out.

packages/supervisor/README.md
jnicklas: Okay, looking at this again, I do see one point to this, which is that a (hypothetical) simple_one_for_one supervisor, would probably want to specify the operation to run and its arguments separately, at least that is how the Erlang version works, so it could look something like this:

```typescript
let supervisor = createDynamicSupervisor({ run: someOperation });
supervisor.addChild({ args: ['foo', 'bar', 'baz'] });
```

Or soemthing. I'm not sure what the API should look like exactly. This would mean that having `args` on the regular supervisor would add more symmetry. But it's probably a moot point, especially since we haven't even added this type of supervisor yet. What do you think, should we remove `args` for now?

jnicklas:
@cowboyd I've added some minor improvements, including rendering error messages nicely, and fixing the dependency to use a caret. Held off on removing `args` awaiting a decision on what you think is best there.

packages/supervisor/README.md
cowboyd: Could we have a strongly typed API for this hypothetical dynamic supervisor derived from the type of `someOperation`? For example, let's say the type of the `run` property was:

```ts
interface Run<TArgs extends unknown[]> {
  (... args: TArgs): Operation<void>;
}
```

Then we could make the create function return something like:

```ts
interface SimpleOneForOne<TArgs extends unknown[]> {
  addChild(...args: TArgs): void;
}
```

e.g.
```ts
let supervisor = createDynamicSupervisor({ run: sleep });

 // @ts-expect-error
supervisor.addChild('one hundred');

// ok!
supervisor.addChild(100)
```

If that is the case, then I think we can do without `args` in all cases.

jnicklas:
@cowboyd coming up to a point where I'd actually like to try and use this. I've made a few changes and improvements. I can obviously use this from my own project, but any chance this could get merged?

cowboyd:
Hey @jnicklas! that's great to hear! I'm still OOO this week, but will have a look first thing next year!

cowboyd (approved):
I can't wait to try it myself, but it also sounds like you have a use-case for it as well. Very curious to hear what the experience is.

cowboyd:
@jnicklas I've approved, but it looks like there is a ESLint check that is failing...

cowboyd:
I'd still love to see a supervisor package, but with the advent of v3 this code would have to be re-worked significantly. I'm going to go ahead and close, but nothing would give me greater joy than to have you working on Effection code again!

PR #649, opened 2022-05-27
Expose formatError so packages other than  can make nice errors

jnicklas:
## Motivation

Other packages might want to format errors the same way that `main` does. See for example #648 

## Approach

Just make this public for both node and browser.

### DIFF
--- a/.changes/expose-format-error.md
+++ b/.changes/expose-format-error.md
@@ -1,0 +1,4 @@
+---
+'@effection/main': minor
+---
+Expose `formatError` so other packages can format errors the same way as `main`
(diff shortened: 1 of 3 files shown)


cowboyd (approved):
I was actually just missing this the other day.

jnicklas:
Not sure why the windows tests are failing? Seems like they're just timing out? I don't seem to have permissions to re-run.

cowboyd:
> Not sure why the windows tests are failing? Seems like they're just timing out? I don't seem to have permissions to re-run.

I think we need to up the timeout. I manually re-ran them and they were fine. You're good to go here.

PR #656, opened 2022-07-14
`process.expect()` never returns stderr/stdout on exception

taras:
## Motivation

We encountered that `process.expect()` is not allowing to capture `stderr/stdout` because when it throws, it never actually returns a value. I'm not really sure if there is a solution to this other than maybe attaching stderr/stdout to error message.

## Approach

I just added failing test.

### DIFF
--- a/packages/process/test/exec.test.ts
+++ b/packages/process/test/exec.test.ts
@@ -42,6 +42,13 @@ describe('exec', () => {
       expect(error.name).toEqual('ExecError');
     });
 
+    it('allows to read output on exception', function* () {
+      let { stdout, stderr } = yield exec("node './test/fixtures/hello-world-failed.js'").expect();
+
+      expect(stdout).toEqual('hello world\n');
+      expect(stderr).toEqual('boom\n');
+    });
+
     it('applies labels', function*() {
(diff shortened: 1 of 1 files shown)


cowboyd:
@taras We should absolutely print some portion of stdout and stderr to the error message. The workaround is to use `exec().join()` which will always return a process result unless the process failed to start in the first place.

packages/process/test/exec.test.ts
```diff
@@ -42,6 +42,13 @@ describe('exec', () => {
       expect(error.name).toEqual('ExecError');
     });
 
+    it('allows to read output on exception', function* () {
+      let { stdout, stderr } = yield exec("node './test/fixtures/hello-world-failed.js'").expect();
```
cowboyd, line 46: This should, in fact, never return because it raises an error. It's the same as `yield Promise.reject()` 

`expect()` means that you expect it to succeed and will fail if it doesn't. If you change this to `exec().join()` it will wait for the process to finish, whatever the result and return the process result.

I think the answer here is to include a portion of `stdout` and `stderr` into the exception so that you can quickly diagnose it.

cowboyd:
This is the "expected" behavior 😆 

However, the error message is suboptimal and should be fixed, see #657

PR #660, opened 2022-08-16
implement @effection/vitest package

jbolda:
## Motivation

Implementing `@effection/vitest` package that we can use effection-based generators within vitest.

## Approach

The implementation pretty closely mirrors what we are doing with the `@effection/jest` package.

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -171,6 +171,13 @@
         "@effection/websocket-client",
         "effection"
       ]
+    },
+    "@effection/vitest": {
+      "path": "./packages/vitest",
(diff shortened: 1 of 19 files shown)


cowboyd (commented):
We should rename `mocha.test.ts` to `vitest.test.ts`

jbolda:
Oh, it is named `vitest.test.ts`. I made a small change in the mocha test suite as those tests kept timing out on windows. I suspect we were right on the edge of the timeout, so it was a bit flaky.

cowboyd (approved):
Nice work!

PR #663, opened 2022-09-13
Use Node 18 in CI

cowboyd:
## Motivation

Because of https://github.com/vitest-dev/vitest/issues/1191 we were seeing intermittent failures in CI.

## Approach
To resolve this issue, just bump the tests suite to use Node 18. This has no effect on the release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 executors:
   linux:
     docker:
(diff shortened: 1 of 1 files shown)


jbolda (approved):

jbolda:
Actually, should we update our volta config in the `package.json` to match?

PR #665, opened 2022-09-20
upgrade docs site to docusaurus v2 stable

jbolda:
## Motivation

We were on a beta build of docusaurus. Updating to the stable version.

## Approach

Update and then deal with the `clientHeight` error via poking and searching for issues.


cowboyd (approved):
Let it rip, and we can fix issues as they arise

PR #671, opened 2022-09-26
give the InspectState task a type

dagda1:
test

### DIFF
--- a/packages/inspect-ui/app/task-tree-page.tsx
+++ b/packages/inspect-ui/app/task-tree-page.tsx
@@ -1,22 +1,22 @@
 import React, { useCallback, useMemo } from "react";
 import { useSlice } from "@effection/react";
 import { Navigate } from "react-router-dom";
-import { InspectStateSlice } from "./app";
 import { TaskTree } from "./components/task-tree";
 import { useSearchParams, useNavigate } from "react-router-dom";
 import { parse, stringify } from "query-string";
-import { InspectState } from "@effection/inspect-utils";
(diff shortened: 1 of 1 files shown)


packages/inspect-ui/app/task-tree-page.tsx
```diff
@@ -1,22 +1,22 @@
 import React, { useCallback, useMemo } from "react";
 import { useSlice } from "@effection/react";
 import { Navigate } from "react-router-dom";
-import { InspectStateSlice } from "./app";
 import { TaskTree } from "./components/task-tree";
 import { useSearchParams, useNavigate } from "react-router-dom";
 import { parse, stringify } from "query-string";
-import { InspectState } from "@effection/inspect-utils";
+import type { InspectState } from "@effection/inspect-utils";
+import { Slice } from '@effection/atom';
```
cowboyd, line 8: This is also just a type. Should we import it as such?
```suggestion
import type { Slice } from '@effection/atom';
```

dagda1: @cowboyd we should be using [this](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md) everywhere and if you run `eslint --fix` it will do this for us automatically.  `import type` is such a safer option for bundle size etc.

dagda1: This was just a test PR to try and re-create the bug that @jbolda has now fixed.

Closing

PR #678, opened 2022-11-08
remove all cjs from @effection/vitest

dagda1:
## Motivation

When trying to run `@effection/vitest` on an external project it is defaulting to `commonjs` resolution.

## Approach

Add an `exports` to `@ffection/vitest` and remove `"main"` and friends


taras (approved):
👍

packages/vitest/README.md
```diff
@@ -1,4 +1,23 @@
 # @effection/vitest
+
+## config
+
+`@effection/vitest` needs to be inlined to avoid node resolution problems and vitest version discrepencies:
```
jbolda: Can we say more about this and/or link to vite/vitest docs about it? It seemed to be working before, but is not working now? I think it would be helpful to have the "why" behind this bit of config.

dagda1: can you show me it working in an external repo from something which is not this branch?

packages/vitest/package.json
```diff
@@ -2,9 +2,12 @@
   "name": "@effection/vitest",
   "version": "2.0.1",
   "description": "Effection Integration for the Vitest framework",
-  "main": "dist-cjs/index.js",
-  "module": "dist-esm/index.js",
   "types": "dist-esm/index.d.ts",
+  "type": "module",
```
jbolda, line 6: I think this should be safe as vitest is ESM first, yea? Any caveats here we may need to think about?

dagda1: vitest is esm only from what I can gather.  I don't think it make sense to have a commonjs build

packages/vitest/README.md
jbolda: This was the original spike that I did outside of the repo.
https://github.com/jbolda/covector/tree/vitest-spike

jbolda (approved):

PR #685, opened 2022-12-21
Upgraded dependencies

taras:
## Motivation

Bumping all of the dependencies to latest using `yarn upgrade-interactive`

## Approach

1. Fixed all typescript errors that resulted from upgrading typescript
2. Moved all shared dev dependencies to root


cowboyd (approved):
Thank you 🙏

I didn't see anything other than dev dependencies updated, so should not require a changes set, right?

PR #692, opened 2023-03-31
♻️Import instructional Effection and migrate dev workflows to Deno

cowboyd:
## Motivation
For some time, `v3` work has proceeded using
https://github.com/cowboyd/instructional-effection as the base of operations. It is now time to settle on that implementation for `v3` and begin migrating that codebase into this one.

## Approach
This gets rid of all the NPM/yarn based workflows and imports the tests from instructional-effection so that we can run them on this repository. Following this we can one-by-one go through the `packages/*`  (starting with core) make sure they are still relevant, and if so, make sure that all the pertinent code and test cases are accounted for in the new tree.


taras (approved):
👍

neurosnap (approved):

cowboyd:
Merged! 🎉 

Now all our development can happen directly on `v3`. Not that this is still not the default repository branch, nor the stable release. Before we can make those changes, we need to have the docs updated, and have a migration path forward from `v2` to `v3`

Issue #693, opened 2023-04-06
Revamp `Subscription`, `Stream` APIs to make combinators easy


cowboyd:
Right now a stream is a context-free operation that when run, yields a subscription that is coupled to the current context. And the subscription will yield the next item, and importantly, when that context passes away, the subscription will be automatically be released and do whatever cleanup is necessary.

It looks like this:

```ts
export interface Operation<T> {
  [Symbol.iterator](): Iterator<Instruction, T, any>;
}

export type Stream<T, TReturn> = Operation<Subscription<T, TReturn>>;

export type Subscription<T, TReturn> = Operation<IteratorResult<T, TReturn>>;
```

But this is annoying because streams and subscriptions are very similar in the way they behave. The only difference is that one is "concrete" and the other is abstract. However, we would expect the same kind of result for both `yield* first(stream)` and `yield* first(subscription)` and so we'd like for `first` not to care if it is operating on a `Stream` or a `Subscription`. The same applies to `filter`, `map`, `forEach`, etc....

I think the answer might lie in more closely aligning `Stream` and `Subscription` with `Iterable` and `Iterator` and `AsyncIterable`, and `AsyncIterator` which is what they really are: `Stream === OperationIterable`, `Subscription ===  Operation Iterator`. 

1. `AsyncIterator`s are `AsyncIterable` by returning themselves for the `[Symbol.asyncIterator]()` method.
2. `Generator`s are `Iterable` by returning themselves for the `[Symbol.iterator]()` method.

I propose that we make `Subscription` implement `Stream`, by returning itself for the `[Symbol.iterator]` method. That way, all combinators can be implemented to work on `Stream`, and a live subscription is just a special type of stream to which you are already subscribed. The new types would look like:

```ts
export interface Operation<T> {
  [Symbol.iterator](): Iterator<Instruction, T, any>;
}

export type Stream<T, TReturn> = Operation<Subscription<T, TReturn>>;

export interface Subscription<T, TReturn> extends Stream<T,TReturn> {
  next(): Operation<IteratorResult<T, TReturn>>
}
```

Now, if we define `forEach` to work on a `Stream`:

```ts
declare function* forEach<T,R>(stream: Stream<T,R>, each?: (t: T) => Operation<void>): Operation<R>;
```

Then, it will work on both a stream, and a live subscription. The same applies for all our combinators you would care to write.

taras:
This sounds like a nice simplification.

cowboyd:
[...]

PR #695, opened 2023-04-18
migrate events

dagda1:
## Motivation

Migrate `on` and `once` to v3 and add better typing around the event type.

## Approach

Extract all the `onXXX` events from the event target and allow a fallback to string if none is found.

### DIFF
--- a/lib/deps.ts
+++ b/lib/deps.ts
@@ -1,3 +1,2 @@
 export { assert } from "https://deno.land/std@0.158.0/testing/asserts.ts";
 export * from "https://deno.land/x/continuation@0.1.5/mod.ts";
-export { expectType } from "https://esm.sh/ts-expect@1.3.0";
--- a/lib/deps.ts
+++ b/lib/deps.ts
@@ -1,2 +1,3 @@
 export { assert } from "https://deno.land/std@0.158.0/testing/asserts.ts";
 export * from "https://deno.land/x/continuation@0.1.5/mod.ts";
+export { expectType } from "https://esm.sh/ts-expect@1.3.0";
--- a/lib/events.ts
+++ b/lib/events.ts
@@ -1,8 +1,10 @@
 // deno-lint-ignore-file no-explicit-any ban-types
 import { createChannel } from "./channel.ts";
-import { action, resource, suspend } from "./instructions.ts";
+import { resource } from "./instructions.ts";
+import { first } from "./mod.ts";
 import { useScope } from "./run/scope.ts";
 import type { Operation, Stream } from "./types.ts";
+import { EventEmitter } from "node:events";
 
 type FN = (...any: any[]) => any;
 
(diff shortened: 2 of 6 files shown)


lib/deps.ts
```diff
@@ -1,2 +1,3 @@
 export { assert } from "https://deno.land/std@0.158.0/testing/asserts.ts";
 export * from "https://deno.land/x/continuation@0.1.5/mod.ts";
+export { expectType } from "https://esm.sh/ts-expect@1.3.0";
```
cowboyd: This should probably not be part of the main deps, but maybe put it in `test/suite.ts`? That way, when we bundle it for npm for example, it won't be included.

cowboyd (commented):
What do you think about a curried API that can be used with a `pipe` helper a-la https://gist.github.com/cowboyd/384b456e7093462ac0b1c7b33d6213d7

In other words `on('message')(websocket)` instead of `on(websocket, `message`)`

lib/deps.ts
dagda1: good point, I will change it.

I actually just thought, that I did not bring the tests from the `v2` in.  

I'll need to add those.

Is the external API the same and only the internals different?

dagda1:
> In other words `on('message')(websocket)` instead of `on(websocket, `message`)`

I'm a big fan of pipe so I like it.

Do you think it makes sense to support both?

`on('message')(websocket)` and `on('message')(websocket)`.  I'm thinking that most people know the latter.

cowboyd:
> Do you think it makes sense to support both?

We could try and have overloads? How difficult does that make it for you to work out the types?

dagda1:
> https://gist.github.com/cowboyd/384b456e7093462ac0b1c7b33d6213d7

Any `pipe` I have used is a generic function that does all the type magic for you, e.g [fp-ts](https://github.com/gcanti/fp-ts/blob/master/src/function.ts#L425) so you would not need to overload anything.

You just need an effection aware `pipe` that does all the nice type crazyness.

<blockquote><div> Gist</div><div><strong><a href="https://gist.github.com/cowboyd/384b456e7093462ac0b1c7b33d6213d7">Piping, buffering, filtering, mapping, and general hypothetical combinations API.</a></strong></div><div>Piping, buffering, filtering, mapping, and general hypothetical combinations API. - buffer.ts</div></blockquote>

cowboyd:
`pipe` and `flow` from `fp-ts` is pretty much exactly what we want.

I'm imagining that you would use `pipe` even for very simple stream iterations:

```ts
import { on, pipe } from "effection";
    
let websocket = new WebSocket('wss://localhost:9000');
  
yield* pipe(websocket, on("message"), forEach(function*(message) {
  console.log('received message from websocket', message)
}));
```

Something about that feels kinda right. Everything is just a combinator, even `forEach`. So simple iterations and complex iterations all have the same structure:

```ts
import { buffer, on, pipe, filter } from "effection";
  
yield* pipe(
  new WebSocket('wss://localhost:9000'),
  on('message'),
  buffer({ size: 100 }),
  map(message => JSON.parse(message.data)),
  filter(data => data.type === "result"),
  forEach(function*(result) {
    console.log('received result from websocket', result)  
  })
);
```

Perhaps we should get #693 sorted before we get this put in.

dagda1:
> Perhaps we should get #693 sorted before we get this put in.

I think Pipe and flow should be a separate PR.

dagda1:
I think it would be seriously cool to have them here

lib/events.ts
```diff
@@ -0,0 +1,122 @@
...
+    try {
+      yield* suspend();
+    } finally {
+      target.removeEventListener(
+        name,
+        resolve as EventListenerOrEventListenerObject,
+      );
+    }
+  });
+}
```
cowboyd: Can we implement this using [`first()`](https://github.com/thefrontside/effection/blob/v3/lib/first.ts#L3) on the stream?

cowboyd (commented):
A couple of items:

1. These should be compatible with `pipe()`, so that you can say `pipe(websocket, on("message"))`. That means individually, it would be `on("message")(websocket)`.
2. I don't think we should support `EventEmitter` anymore since now almost every event emitter is also a [`NodeEventTarget`](https://nodejs.org/api/events.html#class-nodeeventtarget). It caused troubles when you had event callbacks that passed multiple arguments and led to us having to make a `onEmit` and `onceEmit` variations. If folks want to use the event emitter api we can provide a separate subscription mechanism.
3. I think we can express `once()` in terms of `first()` composed with `on`. i.e. the "first" item in the event stream:

```ts
function once(target, eventName) {
  return first(on(eventName)(target));
}
```

Should `once()` be a combinator? I'm not sure.

dagda1:
> 1. These should be compatible with `pipe()`, so that you can say `pipe(websocket, on("message"))`. That means individually, it would be `on("message")(websocket)`.

This is where it gets interesting, in order for any function to work with pipe, it needs to be unary and it needs to be data last

So we would need to rewrite `on` and `once` like this:

```ts
function on2<K extends EventList<unknown> | (string & {})>(name: K) {
   return function<T extends EventTarget>(target: T): Stream<EventTypeFromEventTarget<T, K>, never> { {
```

We lose all the typing because we rely on the `target` to get the list of `onXXXX` event names for the nice autocomplete.

dagda1:
> 
> ```ts
> function once(target, eventName) {
>   return first(on(eventName)(target));
> }
> ```

nice!

dagda1:
> Should once() be a combinator? I'm not sure.

Can I get a definition of `combinator` in this context?  Do you mean `once` would be a higher order function like `map` and `filter`?

If we want anything  to work with pipe then  I think anything needs to be a combinator or curried.

cowboyd:
@dagda1 combinator is as you say, a unary function that allows this to be used with `pipe`. I feel like `on()` should be a combinator, but should `once()`?

cowboyd:
> We lose all the typing because we rely on the target to get the list of onXXXX event names for the nice autocomplete.

😭 😭 😭 

Do we not get typing because the `function` returned is parameterized? 

```ts
return function<T extends EventTarget>(target: T): Stream<EventTypeFromEventTarget<T, K>, never> {
```

cowboyd:
In other words, it should be able to infer if we say:

```ts
pipe(websocket, on("mess // => runtime knows that "message" is the only valid event name
```

right?

dagda1:
> In other words, it should be able to infer if we say:
> 
> right?

I'm not sure typescript can infer data last functions.  It will work perfectly fine with `pipe` and everything will get inferred but if you were to call the function without `pipe` then the inference will not work

I have a passing and failing test for both

ccc

I'll do a bit more digging.  I hope I'm wrong about this.

dagda1:
@cowboyd I've overloaded `on` and `once` so at least this type checks

```ts
expectType<Operation<CloseEvent>>(pipe(socket, once("close")));

expectType<Operation<MouseEvent>>(once("click", domElement));
```

Everything is beautiful with `pipe` but maybe less beautiful without.

dagda1:
Screenshot 2023-05-09 at 09 12 43

dagda1:
@cowboyd I've removed the `onFP` and `onceFP` and all the craziness associated with them

cowboyd (approved):
🥳 I love the strong typing so that you get completion on the event names! and combined with the `yield*` syntax, where you get strongly type event values, it really does make a difference.

```ts
let event = yield* once(websocket, 'close');
```

Issue #699, opened 2023-04-26
Composable stream combinators


cowboyd:
The `v2` method for combining streams was both powerful, but also limited due to being based on method chaining. The limits of chaining are well known, and so we need some way of combining them without. We can follow prior art out there by using simple function composition. This will allow folks to bring their own combinators without prototype fiddling.

Unlike a purely functional API like RxJS, we don't have any interest in re-creating flow control primitives as stream processing operations. Instead, our goal is to do linear transforms of single streams which we use as event sources for [normally structure `for-loops`](https://github.com/thefrontside/effection/issues/697)

This will allow us do do stuff like:

```ts
import { protocol } from "./protocol.ts";

let events = pipe(
  websocket,   
  on("message"),
  map(function*(event) {    
    return protocol.parse(event.data);
  }),
);

for (let next of yield* iterate(events) {
  let event = yield* next;
  /*.... handle */
}
```

What we'll need are:

* `pipe()` - basically a straight ~~ripoff~~ port of the `fp-ts` function with all its type safe goodness. There's nothing special about this, just function composition.
* `map()` - applies an operation over a stream.
* `filter()` - applies an filter over a stream

To facilitate this type of composition, each combinator should return a unary function of type `Stream => Stream`. For example, the type of `map()` should be:

```ts
declare function map<T,X,TClose>(fn: (item: T) => Operation<X>): (stream: Stream<T, TClose>) =>  Stream<X, TClose>;
```

cowboyd:
closed by #701

PR #701, opened 2023-05-01
composable stream combinators

dagda1:
## Motivation

Implementation of #699 

## Approach

Introduce a `pipe` function which is a slightly altered fp-ts pipe to allow the creation of a pipeline of combinators.

This PR also introduces a `lift` function to make any normal function an operation.


cowboyd (commented):
As nice as it is to be able to filter with both a simple function, or operation function, there is a payoff in that combinators you make have to use this convention in order to feel idiomatic. I'm kind of tempted to force lifting to be manual at first, and see how it plays out because we can always auto-lift in the future, but we can't go back.

In other words, rather than saying:

```ts
let stream = pipe(messages, map(msg => msg.data);
```

You would have to say:

```ts
let stream = pipe(messages, map(lift(msg => msg.data));
```

It's a bit more verbose, but it has the advantage of being really simple. Plus, TypeScript is our friend here and will prevent bad usages.

When [generator arrow syntax](https://github.com/tc39/proposal-generator-arrow-functions/issues/2) arrives, this would just be:

```ts
let stream = pipe(messages, map(*(msg) => msg.data));
```
(Or some variant thereof. The syntax isn't finalized yet)

Also, what do you think about calling it `op` or `fn`? That would make it both convenient, and maybe a bit less esoteric than lift which can scare off the non FP crowd.

That would make it:

```ts
let stream = pipe(messages, map(op(msg) => msg.data));
```

or

```ts
let stream = pipe(messages, map(fn(msg) => msg.data));
```

dagda1:
> ```ts
> let stream = pipe(messages, map(op(msg) => msg.data));
> ```
> 
> or
> 
> ```ts
> let stream = pipe(messages, map(fn(msg) => msg.data));
> ```

I really like `lift`, but I get the point it is a bit esoteric.

I agree with the points and have updated the code.

lib/op.ts
```diff
@@ -0,0 +1,27 @@
...
+  ...args: Args
+) => Operation<R>;
+
+export function isOperation<Args extends unknown[], R>(
+  fn: AnyFunction<Args, R> | OperationFunction<Args, R>,
+): fn is OperationFunction<Args, R> {
+  // deno-lint-ignore no-prototype-builtins
+  return fn.hasOwnProperty("prototype") && "next" in fn.prototype &&
+    "return" in fn.prototype && "throw" in fn.prototype;
+}
```
cowboyd: Is this used anywhere anymore?

test/stream-combinators.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        return item.toUpperCase();
+      });
+
+      let shorts = filter(function* (a: string) {
+        return a.length < 4;
+      });
+
+      let subscription = yield* pipe(channel.output, shorts, upCase);
+
+      expectType<Subscription<string, string>>(subscription);
```
cowboyd: If we're going to add a type expectation, it would be cool to test it on a mapping that actually changes the type, like say from `string` -> `number`: `map(op(str => str.length))`

lib/filter.ts
```diff
@@ -0,0 +1,31 @@
...
+            while (true) {
+              let next = yield* subscription.next();
+
+              if (next.done) {
+                return next;
+              } else {
+                if (yield* predicate(next.value)) {
+                  return next;
+                }
+              }
```
cowboyd: Should this be an `else if`?

lib/op.ts
dagda1: nope

test/stream-combinators.test.ts
dagda1: The following comes back as `Subscription<number, string>`.  I take it, this is correct?

```ts
let shorts = filter(function* (a: string) {
  return a.length < 4;
});

let length = map(function* (item: string) {
  return item.length;
});

let subscription = yield* pipe(channel.output, shorts, length);
```

cowboyd: 🥳 hooray for `yield*`

cowboyd: (and hooray for pipe)

cowboyd (approved):
🥳 Awesome work!

dagda1:
I loved this PR.  What fun!

PR #702, opened 2023-05-28
chore(npm): generate separate types for esm

neurosnap:
Reference: https://github.com/denoland/dnt/issues/307

## Motivation

- https://arethetypeswrong.github.io/?p=effection%403.0.0-alpha.7

## Approach

All we need to do is upgrade `dnt` and it will automatically generate separate types for cjs and esm.


cowboyd (approved):
🙏 Thanks for tracking this down!

cowboyd:
@neurosnap Looks like there is an issue with the `deno.lock` Do we need to update it?

neurosnap:
The issue is that `esm.sh` dependencies need to be pinned: https://esm.sh/#faq

<blockquote><div><strong><a href="https://esm.sh">ESM>CDN</a></strong></div><div>A fast, smart & global CDN for modern(es2015+) web development.</div></blockquote>

Issue #712, opened 2023-06-16
Update v3 docs to have feature parity with v2 docs


cowboyd:
- [x] ~~breadcrumbs along the top (Home > getting started > introduction)~~ unecessary
- [x] right nav to highlight sections https://unifiedjs.com/explore/package/@jsdevtools/rehype-toc/ https://github.com/rehypejs/rehype-autolink-headings
- [x] responsive templates for 3 break points
- [x] Fix title on the nav bar (there is no text)
- [x] Hero section for the index
- [x] Doc search
- [x] Create a separate plan for migrating the content.
- [x] Previous and Next sections
- [x] Footer
- [x] Style the left navigation sidebar
  - [x] position fixed
  - [x] hover effects for links
  - [x] active effects for current page
- [x] left nav for docs is usable on mobile
- [x] inline code language for inline code blocks

PR #713, opened 2023-06-22
✨v3 website

cowboyd:
## Motivation

As part of the migration to v3 we need to be able to update the documentation as well be able to have a method to embed playgrounds into our websites. This will be crucial for Effection, but also in derivative libraries and future products.

Goals:

1. Fast deploys, easy previews.
2. Zero build
3. Just HTML™️
4. Explicit vs Implicit Render

## Approach

I wasn't able to achieve (2), but the other points work well, and there is hope that (2) will be achievable if any of the v8 isolate based runtimes (Vercel, Netlify, Deno) will support dynamic import.

For now, this is just simple Deno server that serves HTML templates. The server definition is in `www/server.ts`. It uses a route recognizer to parse path and query parameters, and then calls an HTML template that is expressed in JavaScript (The data structure is
defined by https://github.com/cowboyd/html) and then serialized to a DOM using [deno_dom](https://github.com/denoland/deno_dom) and Google's [incremental-dom
library](https://github.com/google/incremental-dom).

Tailwind CSS is used as the atomic CSS framework.

The documentation, like the v2 website is based on MDX. Ideally, we would like to not require a build step for this, and in development mode we can, by compiling the MDX into a JavaScript module. However, becaue Deno Deploy [does not support dynamic
import](https://github.com/denoland/deploy_feedback/issues/1), this is not an option at the moment.

As such, you need to run `deno task build` in the `www/` directory to generate the `MDX` files and then commit them to source control. The dev script does this automatically.

Future directions:

### TODOs and Open Questions

- [ ] We need to figure out navigation for where to host and how to browse back to the `v2` docs
- [ ] Resolve all the outstanding issues in https://github.com/thefrontside/effection/issues/712
- [ ] after merge: move the deno deploy branch to point to `v3` instead of `v3-www`
- I think we can simplify things considerably by using [hast](https://github.com/syntax-tree/hastscript) as the structural representation of the HTML. This means several things: We can apply rehypejs plugins at any level, not just MDX, and also we do not need to use `deno_dom` nor `incremental-dom` for server side rendering since there is a direct transform from `hast` -> `string`
- It would be nice to have autocompletion and type checking for tailwind classes.
- Because we're based on Effection, we can use really nice helpers to interrupt a request mid-flight to redirect, render 404, etc...


www/docs/build.ts
```diff
@@ -0,0 +1,89 @@
...
+      modules.push({
+        filename,
+        url: new URL(`./${filename}`, import.meta.url),
+        varname: `doc${moduleIdx++}`,
+        importPath: `./${filename}.js`,
+      });
+    }
+  }
+
+  yield* expect(esbuild.build({
```
dagda1, line 65: oh nice, this could open the door for vanilla-extract

cowboyd: Yes, and if we can get support for dynamic modules, could be done at runtime.

www/docs/build.ts
```diff
@@ -0,0 +1,89 @@
...
+import rehypeAutolinkHeadings from "https://esm.sh/rehype-autolink-headings@6.1.1";
+import rehypeAddClasses from "https://esm.sh/rehype-add-classes@1.0.0";
+import rehypeSlug from "https://esm.sh/rehype-slug@5.1.0";
+import rehypeToc from "https://esm.sh/@jsdevtools/rehype-toc@3.0.2";
+
+import builder from "https://esm.sh/@mdx-js/esbuild@2.3.0";
+
+import structure from "../docs/structure.json" assert { type: "json" };
+
+const mdx = builder({
```
dagda1, line 18: THis is almost identical to my setup, the only  other one i use is [rehype-preset-minify](https://www.npmjs.com/package/rehype-preset-minify)

www/docs/build.ts
dagda1: fresh uses esbuild too.  It has an undocumented hook you can use

dagda1 (approved):
Great work!  Should the compiled `*.mdx.js` files be source control?

cowboyd:
> Great work! Should the compiled `*.mdx.js` files be source control?

If we want to not have a build step on deploy, then yes, they need to be in source control. I followed the Fresh model of having your dev script rebuild as you work, so tome thing similar is what fresh does with their `fresh.gen.ts` file. Ideally, we could make esbuild output a single module with keys rather than a bunch of files. That way, we would need only one generated file under source control, but I could figure out how to do it.

Longer term, I wish there would be support for async import()

PR #715, opened 2023-06-28
🔥Get rid of view data structure and just render templates

cowboyd:
## Motivation

Eventually, we'd like to get to a point where we can automatically map the url path to the hierachy of templates that get rendered, but for now all it does is obfuscate things.

It's not clear what this is doing with it's esoteric lookups in a "view" data structure.

```ts
 yield* Outlet.set(yield* view[1]["/docs/:id"](doc));
 yield* Outlet.set(yield* view[0]({ title: `${doc.title} | Effection` }));
 return yield* Outlet;
 ```

This accomplishes the same thing and is a lot clearer (render the
documentation template inside the app template:

```ts
 return yield* render(
   AppHtml({ title: "Effection" }),
   IndexHtml(),
  );
```

Eventually, we can figure out how to render based on the url, but
until that day arrives, let's opt for clarity and simplicity.

## Approach

To accomplish this, we define a "render" function that takes a list of
template operations and renders them in reverse order taking each
result and making it the `outlet` of the next one.


dagda1 (approved):
Looks good, one nit....should `view.ts` be renamed `render.ts`?

cowboyd:
> Looks good, one nit....should `view.ts` be renamed `render.ts`?

Maybe? But it is also where you import `outlet` from.

What do you think about

```ts
import { outlet } from "freejack/render";
```

If that makes sense, then let's change it.

dagda1:
I did not consider the `outlet`, might be fine the way it is.

PR #716, opened 2023-07-02
♻️Use Rehype AST as internal HTML format.

cowboyd:
## Motivation

Our javascript templates were using our own custom html which meant that we had to write the serialization logic ourselves (we used deno-dom + google's incremental-dom to do this). It's a nifty hack, but really there is no need to have a full-fledged DOM on the server. It turns out there is already a full dom-as-data representation out there called [hast][hast] and it is what's used by [rehype][rehype] to enable its pipeline. Essentially each rehype plugin is just a `hast -> hast` function.

By using this intermediate format, we get all the benefits of the rehype ecosystem, and we don't have to maintain a custom library to do it.

In the future, we can move all our rehype plugins outside of the doc store and just make them as a series of transformations.

## Approach

This replaces the custom "html" package with [hastx](https://github.com/thefrontside/hastx) that creates a hast value from JSX. This means that we can inherit the serialization code, and that we don't have to munge the HTML as much, and whatever transforms we do in the future can be done as rehype plugins (In fact, our tailwind support should be implemented as rehype plugin)

[hast]: https://github.com/syntax-tree/hast
[rehype]: https://github.com/rehypejs/rehype


dagda1 (approved):
🚢

PR #717, opened 2023-07-03
✨Add main() method as top-level entry point.

cowboyd:
### Motivation

The docs make extensive use of the `main()` method, which is why it should exist in order to publish them.

Also, the `run()` method serves as a bare-bones entry point into any Effection operation. However, when you are using it to control your entire applications there are certain things that you always want to happen. They are this:

On all environments:

✅ print out an error message if an unhandled error occurs
✅ provide a mechanism to "exit immediately" from the program, and yet still gracefully perform all shutdown.

On servers:

✅ initiate graceful shutdown on SIGINT.

On browsers:

✅ initiate graceful shutdown on the `unload` event.

If you don't do this, you need to write a lot of boiler-plate code to make sure that you shutdown gracefully whenever possible, and that you handle errors properly. However, with a single entry point that contains everything baked in, you get all of it for free. In a word, `CTRL-C` just works, and on the browser abruptly killing a tab exits gracefully.

## Approach

This carries over the "main" package from `v2` with the notable exception that you can now use an `exit()` operation to immediately halt the program.

It uses the new main method to implement the web server for the website, which demonstrates how much clarity it brings.

Other changes of note:

- removed deno.lock file from the top repository because this is a library and doesn't need a lock file
- added jsdocs. I think we should start doing this from here on out now that the APIs are becoming stable.
- the main tests require starting a child process, and so you need to run `deno test -A` to run all the tests.

### Alternate Designs

There is an argument to be made that "less is more" and that this could be a separate package, or a separate entry point. In other words, you would import in Deno or Browser from:

```ts
import { main } from "https://deno.land/x/main.ts";
```
And in Node:

```ts
import { main } from "effection/main";
```

It is also potentially annoying that you will download the code for all platforms even though you only need it for the one you're running on.


lib/main.ts
```diff
@@ -0,0 +1,144 @@
...
+  let hardexit = (_status: number) => {};
+
+  let result = await run(() => action<Exit>(function* (resolve) {
+    // action will return shutdown immediately upon resolve, so stash
+    // this function in the exit context so it can be called anywhere.
+    yield* ExitContext.set(resolve);
+
+    try {
+      let interrupt = () => resolve({ status: 130, signal: "SIGINT " });
+      return yield* withHost({
```
neurosnap: I really like this pattern matching style!

lib/main.ts
```diff
@@ -0,0 +1,144 @@
...
+): Promise<void> {
+  let hardexit = (_status: number) => {};
+
+  let result = await run(() => action<Exit>(function* (resolve) {
+    // action will return shutdown immediately upon resolve, so stash
+    // this function in the exit context so it can be called anywhere.
+    yield* ExitContext.set(resolve);
+
+    try {
+      let interrupt = () => resolve({ status: 130, signal: "SIGINT " });
```
neurosnap: Why did you need a space in `SIGNINT `?

lib/main.ts
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      console.error(result.message);
+    }
+  }
+
+  if (result.error) {
+    console.error(result.error);
+  }
+
+  hardexit(result.status);
```
neurosnap, line 118: I'm curious, if `effection` properly shuts the task tree down, why do we need a `hardexit`?

.github/workflows/verify.yaml
```diff
@@ -29,4 +29,4 @@ jobs:
         run: deno lint
 
       - name: test
-        run: deno test
+        run: deno test -A
```
neurosnap: nit: Do you think we should be more specific here with the permissions we allow?  In the context of a test runner inside CI, it's probably okay, just feels like we are removing a major benefit to using `deno`.

lib/main.ts
```diff
@@ -0,0 +1,144 @@
...
+        },
+      });
+    } catch (error) {
+      resolve({ status: 1, error });
+    }
+
+    yield* exit(0);
+  }));
+
+  if (result.message) {
```
neurosnap, line 106: What do you think about returning a `Result` type to be internally consistent?

test/main.test.ts
```diff
@@ -0,0 +1,112 @@
+import { describe, expect, it, useCommand } from "./suite.ts";
+import { run, sleep, expect as $ } from "../mod.ts";
```
neurosnap: nit: `$` is weird and confusing to read, especially with my past trauma from jquery; maybe this is the exposure therapy I need?

neurosnap (commented):
Overall looks good!

I don't see any tests for `node` or `browser` code paths.  I'm sure it would add a lot of complexity to test them, just curious of your thoughts on leaving them out.

lib/main.ts
cowboyd: @neurosnap It's in the case where the exit status is non-zero you have to actually have to transmit the exit code to any caller of the process. I guess we don't need it when the exit code is `0`.

I was also toying with the idea of making main an operation so that it could be fully embedded, but it means you would have to have some more boilerplate:

```ts
let result = await run(main(function*(args) {
  ///  your program here
}));
if (result.status !== 0) {
  Deno.exit(result.status);
}
```

the advantage of the above is that your main method is fully embeddable, and you could have mains embedded with other mains within mains...

The disadvantage is that you have to do your explicit check.

lib/main.ts
cowboyd: oops! nice catch 😁

.github/workflows/verify.yaml
cowboyd: It's a good point, and I agree that it's best to always constrain the permissions to the smallest possible.

I've added a task `test` that only enables the run permission on the `deno` executable. Also, I simplified the simulated daemon to just use a really long sleep. That way it doesn't need any special permission, and all we're testing is the `SIGINT` behavior.

test/main.test.ts
cowboyd: Yeah, I'm not a super fan of it, and it was just there to get rid of the conflict for `expect()`

What do you think of this 😈 

https://github.com/thefrontside/effection/pull/717/files#diff-2608d3bd5af09d726fe4183133e684f22d2a52739a65f893e54a170bcabfa578R7-R16

Only appropriate for an end-user application like a test-suite (not a library) but it's nice in that it let's us yield* to promises directly.

lib/main.ts
cowboyd: It's worth considering as it is very similar to a `Result`, but I'm not sure if the `Result` type fits because it is a value that represents a JavaScript outcome which has a clean boundary between success (`value: T`) and failure(`error: Error`). On the other hand, the type of `result` here needs to represent a _POSIX process outcome_ and the line between success and failure is a bit more blurry in my mind. For example, is a signal exit an error? what value of `Error` would we use to represent that.

neurosnap (approved):

PR #718, opened 2023-07-05
Update documentation for `Operation`

cowboyd:
# Pages to inspect
- https://effection--v3-www-operations.deno.dev/docs/operations
- https://effection--v3-www-operations.deno.dev/docs/actions
- https://effection--v3-www-operations.deno.dev/docs/resources

## Motivation

Operations have changed considerably perhaps the most of anything between v2 -> v3. Instead of being a union of seven different types, it is now just a single unified interface. We need to document all of Effection v3, but especially operations.

## Approach

We're adopting an approach of constantly leaning on the likenesses and the differences between `async/await` and operations. This is because in truth, Effection is really a superset of `async/await` that provides structural concurrency guarantees.

|    Async      |     Effection       |
|---------------|---------------------|
|Promise        | Operation           |
|await          | yield*              |
|async funciton | function*           |
|AsyncIterable  | Stream              |
|`for await`    | `for yield* iterate`|

As such, this tries to constantly highlight while Operations are like promises but better.


taras:
I only read the beginning, but it reads great. I like how it's introduced relative to Promises.

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+await main(function*() {
+  let text = yield* makeSlow('Hello World!');
+  console.log(text);
+});
+```
+
+There is no limit to the number of operations that can be nested within
+each other, and in fact, composition is so core to how operations work that
+_every operation in Effection eventually boils down to a combination of
+just three primitive operations_: `action()`, `resource()`, and `suspend()`.
```
taras, line 140: We should probably mention or link to pages where this is covered

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+}
+
+await main();
+```
+
+This logs `"Hello World!"` to the console after a delay of 1s, and so it works
+as expected when we run it... sort of.
+
+But what happens when you want to halt your sleep operation? Or more commonly,
+what happens when you no longer care about the outcome of your sleep operation?
```
taras: Since `halt` is not a well known terminology in async/await, It might be easier if you just say `what happens when you no longer care about the outcome of your sleep operation?`

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+To understand why this might be, let's ask ourselves how long the program should
+take to execute in Node JS?
```
taras: This is a little meta; how about this

```
Let's take a hypothetical example of racing two sleep operations against each other. How long do you think this program takes to execute in Node.js?
```

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+async function sleep(milliseconds) {
+  await new Promise(resolve => setTimeout(resolve, milliseconds));
+}
+
+await Promise.race([sleep(10), sleep(10_000)]);
+```
+
+If you guessed 10 seconds, then you are technically correct! But if your
+intuition told you that the answer _should_ be 10 milliseconds, then
+congratulations, you are also correct!
```
taras: I understand what you're trying to do here, but it's confusing because of two conflicting statements. I would suggest removing it and going straight into the next paragraph.

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+Our intuition tells us that the answer should be 10 milliseconds because when
```
taras: This paragraph doesn't tell them the answer, it kinda leaves them hanging until the last sentence and even then it doesn't mention the number. How about this?

```
It takes 1000ms even though our intuition says it should be 10ms. It takes 1000ms because `setTimeout` installs a callback and the run loop can not exist until that callback is removed. 

In practice, the application will be able to continue past the await statement after 10ms but your program will not finish for another 990ms until that callback is cleared.

By contast, the equivalent code in Effection will exit immidiately after 10ms as we expect.
```

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+import { sleep, race, main } from "effection";
+
+await main(function*() {
+  yield* race([sleep(10), sleep(10_000)]);
+});
+```
+
+This is because every operation in Effection can be interrupted at any moment
+and as a result any cleanup code associated with that operation will be executed
+immediately.
```
taras: We need to introduce the cleanup separately. It overloads the statement in a way that makes it difficult to parse.
 
```md
By contrast, the equivalent code in Effection will exit immediately after 10ms, as we expect, because every operation in Effection can be interrupted at any time.
```

www/docs/operations.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+If we look at the implementation of the `sleep()` operation in Effection, we can
```
taras, line 212: Before you dive into this, we might want to repeat the guess-the-outcome excercise with try/finally. That will provide the 2 pieces of context necessary to understand the code below.

www/docs/operations.mdx
cowboyd: I'm going to add a section on "writing operations" that will cover these.

www/docs/operations.mdx
cowboyd: I don't quite follow. Are you suggesting before I show the implementation of `sleep()`? Maybe if you can outline what you think the rhythm would be, I can fill in the details.

cowboyd:
@taras I integrated your feedback, and really beefed up the section on actions and resources, can you have another look when you have some free time?

www/docs/operations.mdx
cowboyd: Have a look now. I really unpacked the try/finally in depth.

taras:
@cowboyd the new version is much better, but it's really long now. We should split it up into separate pages with discreet take away for each page. It feels like a lot to get through now.

www/docs/actions.mdx
```diff
@@ -0,0 +1,287 @@
+---
+id: actions
+title: Actions and Suspensions
```
taras, line 3: Maybe `Suspect and Actions` to be symmetrical with the topic order?

cowboyd:
> @cowboyd the new version is much better, but it's really long now. We should split it up into separate pages with discreet take away for each page. It feels like a lot to get through now.

@taras Ok, I split them into the three pages: operations / actions / resources. Check out the links in the original issue comment.

www/docs/actions.mdx
taras: Should this be 2 separate pages?

taras (approved):
let's ship it and we can iterate from here

www/docs/actions.mdx
cowboyd: There’s a bit of a paradox because suspend is a little light and doesn’t really make sense outside of the context of other things, but you also kind of need to know how it works to understand actions…

I can have a play with it.

PR #719, opened 2023-07-06
✨Add all()

cowboyd:
## Motivation
The docs use `all()` and `race()` extensively, so we need to have at least some plausible versions of these functions.

## Approach
This ports over the tests from `v2`, and is implemented as about the simplest version of `all()` that I could get working. However, there is a failing test, and while I think that we can ship it while it still fails, we do need to fix it before releasing `3.0.0`.

### TODOs and Open Questions

- [ ] Our test fails because all() is spawning resources that fail, but any resource that fails while `all()` is running should cause the `all()` to raise an exception error in a resource while it is initializing. We could have any number of resources running, and if any of them fails before the all is completed, we need to be able to catch it.

### DIFF
--- a/lib/all.ts
+++ b/lib/all.ts
@@ -1,0 +1,58 @@
+// deno-lint-ignore-file no-explicit-any
+import type { Operation, Task } from "./types.ts";
+import { useScope } from "./run/scope.ts";
+
+/**
+ * Block and wait for all of the given operations to complete. Returns
+ * an array of values that the given operations evaluated to. This has
+ * the same purpose as
+ * [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all).
+ *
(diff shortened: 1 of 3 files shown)


neurosnap (approved):
Looks good, this is the one I'm using which gives us some more functionality.  Not sure what your temperature is on adding these: https://github.com/neurosnap/starfx/blob/main/fx/parallel.ts

cowboyd:
@neurosnap My take is that functions like `parallel` are necessary to build better concurrency abstractions. However, I think that we should offer safe equivalents of vanilla JS constructs wherever we can. This will help lower the activation energy of adoption if everything piece of knowledge that a JS developer has out of the box can be immediately leveraged to use its structured concurrency equivalent.

For me, it's much the same case as the idea of how we need to have the block-iteration syntax. Whether we have `parallel()` or not, I still think we should have `all()` and `race()`

PR #724, opened 2023-07-10
📄Add docs for `suspend()` and `action()` operations

cowboyd:
## Motivation

Documentation good.

## Approach

I wanted to make sure that folks knew `action()` is equivalent to `new Promise()` but not go to into too much depth since there will be a more complete discussion in the guides.


taras (approved):
👌

PR #727, opened 2023-07-15
attempt at for-of-each

neurosnap:
## Motivation

https://discordapp.com/channels/700803887132704931/1126453685061431316

### DIFF
--- a/test/channel.test.ts
+++ b/test/channel.test.ts
@@ -6,8 +6,28 @@ import {
   it as $it,
 } from "./suite.ts";
 
-import type { Operation, Port, Stream } from "../mod.ts";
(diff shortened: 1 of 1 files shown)


cowboyd:
@neurosnap I was able to get it working. The secret is that the iterator must always have the answer at hand... It can't be an Effection operation. It has to be just a normal iterator of synchronous values. The `each()` and `next()` operations do the async work, and then set the result so that it can be retrieved with a simple `iterator.next()`. 

I went ahead and made it a stack so that it would support nested iterations. I think other enhancements that we should consider to make this production ready are:

1. detecting if user forgot to call `next()` by somehow marking the iteration context as stale and fresh.
2. consider subscribing in an explicit scope that is closed at the end of the iteration.

Issue #728, opened 2023-07-17
Power Scopes


cowboyd:
Scopes are a powerful API to "contain" a set of running operations. They are programmatic interface to the tree of effects, and are  absolutely critical when interfacing with non-Effection code such as web servers or native UI events. However, after some usage, it's clear that the semantics of the scope API as it is in v3 is not quite right. This is an effort to document the problems with things as they currently stand and propose some solutions.

## Error handling

There is no way to protect code from a scope that it is using. In other words, it is impossible to catch the following error from within the current operation.

```ts
let scope = yield* useScope();
scope.run(function*() {
  yield* sleep(10_000);
  throw new Error("time bomb!");
});
```

While it is possible to create an error boundary above the current operation, it is often awkward because it is the code itself that knows how errors should be handled because it is in actual proximity of the code that raises the error.

An example of this is that it is currently impossible to catch an error from an `all()` operation that has a failing resource that fails before the all operation completes.

```ts
try {
  yield* all([failingResource, sleep(1000)]);
} catch (error) {
  // never hit if resource fails before the sleep
}
```
In other words, any errors that happen before the transaction completes should be catchable.

## Shared Context

Operations run within the same context stomp all over each other's context. 

```ts
const Count = createContext<number>("count");

function* increment() {
  let count = yield* Count;
  yield* Count.set(count + 1);
}

function* run() {
  let scope = yield* useScope();
  yield* Count.set(1);
  scope.run(function*() {
    yield* increment();
  })
  scope.run(function*() {
    yield* increment();
  });
  console.log(yield* Count); //=> logs `3` to the console.
}
```

## Proposal

`useScope()` will accept an optional `link` operation that defines, among other things, how the lifetime of a child frame is connected to the current frame. 

```ts
declare function useScope(link?: (error: Error) => Operation<void>): Scope;
```

> The term "link" is derived from [erlang nomenclature](https://www.erlang.org/doc/man/erlang.html#spawn_link-1)

If no `link` operation is provided, then the default link operation will be to run the child in isolation from the current frame. E.g. 

```ts
yield* useScope(function*(error) {
  // do nothing! it is your responsibility to deal with all errors;
});
```

But if we wanted to re-route and log errors, this also allows us to establish a very natural boundary.

```ts
let errors = createChannel<Error>();
yield* useScope(function*(error) {
    yield* errors.input.send(error);
});
```

[...]

PR #729, opened 2023-07-18
♻️ Make each task run in its own frame.

cowboyd:
## Motivation
There are a lot of problems with the way execution is currently handled that are outlined in this issue
https://github.com/thefrontside/effection/issues/728 owing to the fact that there can be many tasks executing against the same frame.

## Approach
This removes that capability, and now the `Block` interface is really private to the `Frame` object. This makes the code a lot simpler.

However, there is one tradeoff that was necessary to make in order for this to work. You can now no longer start "bare resources" with `scope.run()` because they will shut down the frame containing them as soon as the instruction that allocates the resource is finished. Instead, if you want to run a resource from a scope, you will have to suspend it.

```js
scope.run(function*() {
  yield* useResource();
  yield* suspend();
});
```

### DIFF
--- a/lib/instructions.ts
+++ b/lib/instructions.ts
@@ -9,7 +9,6 @@ import type {
 } from "./types.ts";
 
 import { reset, shift } from "./deps.ts";
-import { createFrameTask } from "./run/frame.ts";
 import { Err, Ok } from "./result.ts";
 
 /**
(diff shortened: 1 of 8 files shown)


lib/instructions.ts
```diff
@@ -152,27 +146,16 @@ export function spawn<T>(operation: () => Operation<T>): Operation<Task<T>> {
...
-            let destruction = yield* child.destroy();
-            if (!destruction.ok) {
-              yield* frame.crash(destruction.error);
-            } else if (
-              blockResult.aborted &&
-              !blockResult.result.ok
-            ) {
-              yield* frame.crash(blockResult.result.error);
-            } else if (!blockResult.result.ok) {
-              yield* frame.crash(blockResult.result.error);
```
neurosnap, line 171: what a refactor 🎉

neurosnap (approved):
The reason for this change makes a lot of sense to me.  `scope.run()` should be spiritually similar to `spawn()` in that is creates a new task with a new frame.

No comments really, you've spent a lot of time thinking about this and didn't need to introduce new concepts.

lib/instructions.ts
cowboyd: This was definitely something that made me feel that this was the right direction!

PR #735, opened 2023-07-29
V3 www collections

cowboyd:
## Motivation

Update the `spawn()` and collections APIs for v3

### Approach

* spawn https://effection--v3-www-collections.deno.dev/docs/spawn
* collections https://effection--v3-www-collections.deno.dev/docs/collections

### DIFF
--- a/www/docs/collections.mdx
+++ b/www/docs/collections.mdx
@@ -3,98 +3,99 @@ id: collections
 title: Streams and Subscriptions
 ---
 
->⚠️ These docs have not been updated from version 2 of Effection, and do not
-> apply to version 3. The information you find here may be of use, but may
-> also be outdated or misleading.
+For every tool in the `async/await` toolbox, there is an equilavent in
+Effection. We were already introduced to the most important of these in
+[the introduction][introduction].
 
-Effection ships with a powerful library for working with streams of values.
-This allows you to create complex systems where evented code, stateful code and
-concurrent asynchronous code all are working together seamlessly.
+|    Async        |     Effection         |
+|-----------------|-----------------------|
+|`Promise`        | `Operation`           |
+|`await`          | `yield*`              |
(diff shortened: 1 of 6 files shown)


www/docs/collections.mdx
```diff
@@ -3,94 +3,98 @@ id: collections
...
+subscription acts like a queue that provides a sequence of values via a
+`next()` method. The only difference is that instead of returning a
+`Promise` that _fulfills_ to an iterator result, it returns an
+[`Operation`][operation] that _yields_ an iterator result.
+
+```ts
+function* forEach(subscription) {
+  let next = yield* subscription.next();
+  while (!next.done) {
+    console.log(`value: ${next.value}`);
```
taras, line 46: This is confusing because you're calling `yield* subscription.next()` only once but you're using `while (!next.done)`, to me this reads like you're going to show `value: ${next.value}` for infinity because you're not pulling another value again.  Does reading `next.done` have a side effect, or is line 44 supposed to be inside the while loop?

cowboyd: Nah, it's a straight up bug. If you're doing a manual iteration like this, then you have to manage the reference to the next result yourself which was never being done. It should be:

```ts
function* forEach(subscription) {
  let next = yield* subscription.next();
  while (!next.done) {
    console.log(`value: ${next.value}`);
    next = yield* subscription.next();
  }
  console.log(`done: ${next.value}`);
}
```

The initialization of `next` happens outside the loop so that the value of the last iteration result (where `done === true`) is visible.

cowboyd: Nice catch though. Hopefully the updated example is more clear.

taras: Yeah, much better.

taras (approved):
👍

PR #737, opened 2023-08-01
update the guide on v3 error handling.

cowboyd:
## Motivation

Update the guide on error handling and error boundaries.

## Approach

- https://effection--v3-docs-errors.deno.dev/docs/errors
- Once the scope API is finalized we'll need to incorporate it into this

### DIFF
--- a/www/docs/errors.mdx
+++ b/www/docs/errors.mdx
@@ -54,24 +54,38 @@ nor will it ever raise an error. And, because it will produce neither
 a positive nor negative outcome, it is an error to `await` the result
 of a halted operation.
 
-When  this happens, the promise is rejected with a special halt error:
+When  this happens, the promise is rejected with a special halt error.
+In this example, we show a very long running task that is stopped in the
+middle of its work even though it would eventually return a value if we
+waited long enough.
 
 ``` typescript
-import { run, suspend } from 'effection';
+import { run, sleep } from 'effection';
 
 async function runExample() {
+  // this task takes a long time to return
+  let task = run(function*() {
+    yield* sleep(10_000_000);
+    return "hello world";
+  });
+
+  await task.halt();
+
   try {
-    let task = run(supsend);
-    await task.halt();
-    await task;
+    let value = await task // 💥 throws "halted" error;
+
+    // will never reach here because halted tasks do not produce values
+    console.log(value);
   } catch(err) {
-    console.log("got error", err.message) // => "got error halted"
+    console.log(err.message) // => "halted"
   }
 }
 ```
 
-Notice how it is not an error to `await` the halt operation itself, only to
-`await` the outcome of a halted operation.
+An important point is that `task.halt()` is an operation in its own
+right and will only fail if there is a failure in the teardown. Thus
+it is not an error to await the `task.halt()` operation; it is only an
+error to await the outcome of the operation which has been halted.
 
 ## Error propagation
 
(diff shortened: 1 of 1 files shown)


www/docs/errors.mdx
```diff
@@ -53,51 +49,59 @@ async function runExample() {
...
-    let task = run();
-    task.halt();
+    let task = run(supsend);
+    await task.halt();
     await task;
   } catch(err) {
     console.log("got error", err.message) // => "got error halted"
   }
 }
 ```
```
taras: I was following along up until this point. A lot is going on in this example, and something unexpected can be distracting. 

1. `run(suspend)` was confusing because it's unclear what this is doing. `run(function*(){})` might be less confusing because at least it's visible.
2. You're calling `task.halt()` in the same try/catch block as creating the task and awaiting for task result. It's a lot to process; it's unclear what the error will be because the code is not executable. If there were an inline runner, then it would make the result a bit clearer. The comment is not apparent. 

It might be helpful to break up the code somehow or show where the error will occur.

cowboyd: @taras Have a look now. I think I made it a lot more explicit about what's going on.

taras: This is much better. Does `await task.halt()` throw an error? If it does, can we add this comment `await task.halt(); // 💥 throws halted error`

cowboyd: It actually doesn't, and this is an important point. `task.halt()` is an operation that only fails if there is a failure in the teardown of the task.

This is why I added the text right after the code.

> Notice how it is not an error to await the task.halt() operation, only to await the outcome of the operation which has been halted.

```ts
try {
  yield* suspend();
} finally {
  throw new Error('this halt operation will fail');
}
```

In other words "halting" is an operation that can succeed or fail in its own right. However, once a task is halted, then it will never return a value ever, and so an attempt to wait for its value is an error.

taras: > task.halt() is an operation that only fails if there is a failure in the teardown of the task.

Is more clear than `Notice how it is not an error to await the task.halt() operation, only to await the outcome of the operation which has been halted.`

```ts
  try {
    // halt it. It will no longer result in "hello world"
    await task.halt();

    await task; // 💥 throws halted error
  } catch(err) {
    console.log(err.message) // => "halted"
  }
  ```
  
 Can you move `await task.halt()` outside of the try catch?

Comment (cowboyd) [reply www/docs/errors.mdx]: Ok, update to include that language. It's a good idea to demonstrate that `task.halt()` is not an error by moving it out of the `try/catch`

Comment (taras) [reply www/docs/errors.mdx]: Oh yeah, that's much better. What do you think about removing the comment and showing it in code.

```ts
import { run, sleep } from 'effection';

async function runExample() {
  // this task takes a long time to return
  let task = run(function*() {
    yield* sleep(10_000_000);
    return "hello world";
  });

  await task.halt();

  try {
    const value = await task // 💥 throws "halted" error;

    // will never reach here because halted task will never produce a value
    console.log(value);
  } catch(err) {
    console.log(err.message) // => "halted"
  }
}
```

cowboyd: I like improving the flow of the code, but I don't think it's something we should leave unsaid. Repeating the same concept three times is a good thing, especially when it is very subtle.

I've updated the code to your example, but left the commentary in.

taras (approved):
👍

PR #739, opened 2023-08-07
📄Fix invocation syntax of `on()`

cowboyd:
## Motivation

The invocation of `on()` did not take the correct arguments.

# Approach

fix it


taras (approved):
👍

PR #742, opened 2023-08-10
[v3] switchMap operator on Stream

nandin-borjigin:
## Motivation

[switchMap](https://rxjs.dev/api/index/function/switchMap) operation brings higher order control flow to reactive programming. 

Conceptually, a stream transformation flow can be perceived as an asynchronous computation pipeline (program). And `switchMap` gives us the power to switch between different pipeline (program) based on the upstream value, while gracefully cleaning up the previous, if any, pipeline.

## Approach

`swtichMap` accepts a mapper that maps each value from the upstream (source) stream to a `Operation<Stream<B>>`.

Internally, `switchMap` subscribes to the upstream (source) stream and runs the mapper on each value. Then it spawns a new task and executes the operation to get the mapped (inner) stream and subscribes to it within the task. Any values arrives on the mapped (inner) stream is then forwarded to the output stream of `switchMap` operator.

Before spawning a new task to execute the mapped operation, `switchMap` halts the previous task so that the subscription to previous mapped (inner) stream goes out of scope and "cancelled".

### TODOs and Open Questions

- [ ] Tests

### DIFF
--- a/lib/switch-map.ts
+++ b/lib/switch-map.ts
@@ -1,0 +1,45 @@
+import type { Operation, Stream, Task } from "./types";
+import { createChannel } from "./channel";
+import { spawn } from "./instructions";
+
+export function switchMap<A, B>(op: (a: A) => Operation<Stream<B, unknown>>) {
+  return function <AClose>(stream: Stream<A, AClose>): Stream<B, AClose> {
+    return {
+      *[Symbol.iterator]() {
+        const channel = createChannel<B, AClose>();
+
+        yield* spawn(function* () {
+          let innerTask: Task<unknown> | undefined;
+          const final: AClose = yield* forEach(stream, function* (outer) {
+            if (innerTask) {
+              yield* innerTask.halt();
+            }
+            innerTask = yield* spawn(function* () {
+              const innerStream = yield* op(outer);
+              yield* forEach(innerStream, (inner) => channel.input.send(inner));
+            });
+          });
+          yield* channel.input.close(final);
+        });
+
+        return yield* channel.output;
+      },
+    };
+  };
+}
+
+// Can be replaced with for yield* each once it's available.
+function* forEach<T, TClose>(
+  stream: Stream<T, TClose>,
+  fn: (a: T) => Operation<void>
+): Operation<TClose> {
+  const subcription = yield* stream;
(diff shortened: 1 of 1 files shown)


cowboyd:
This is really cool! I have a couple of questions / feedback:

1. Does `op` need to return an operation that yields a stream, or can it just return a stream? 
2. Can switchMap be used to implement almost anything?
 

General notes: While explicit task management is something that you can do in Effection, it's something that you rarely have to do. Instead, you can just lean on normal function scope rules. I haven't tested this, but here is a sketch that just "escapes" to the next iteration of a while loop in order to make sure everything is properly torn down.

```ts
import type { Operation, Stream } from "./types.ts";
import { createChannel } from "./channel.ts";
import { spawn } from "./instructions.ts";
import { call } from "./call.ts";

export function switchMap<A, B>(op: (a: A) => Operation<Stream<B, unknown>>) {
  return function <AClose>(stream: Stream<A, AClose>): Stream<B, AClose> {
    return {
      *[Symbol.iterator]() {
        const { input, output } = createChannel<B, AClose>();

        yield* spawn(function* () {
          type Tick = IteratorResult<A, AClose>;
          let subscription = yield* stream;

          let next: Tick = yield* subscription.next();

          while (!next.done) {
            let inner = yield* op(next.value);
            next = yield* call(function* () {
              yield* spawn(() => forEach(inner, (value) => input.send(value)));
              return yield* subscription.next();
            });
          }
          yield* input.close(next.value);
        });

        return yield* output;
      },
    };
  };
}
```

As you can see, we wrap the spawned `forEach` in `call()`. When the called operation returns, the spawned resources are shut down and the old subscription is destroyed.

cowboyd:
Hi @nandin-borjigin! I'm going to go ahead and close this since it would be very appropriate for a stream helpers library for Effection, but probably won't make it into core. In fact, we removed our own stream helpers library in the hope that a good community solution would emerge. This is because Effection's reason for being is as a tiny library for providing structural concurrency and effects with maximum interop with the wider JS ecosystem.

The good news is that with the [release of Effection 3.0.0](https://frontside.com/blog/2023-12-18-announcing-effection-v3/) today, the API should be very stable for whoever wants to take on this task.

PR #744, opened 2023-08-11
🐞Fix NodeJS detection logic

cowboyd:
## Motivation

The node detection logic was broken on certain Node environments causing browser code to be executed on NodeJS

## Approach

Use the detection logic from https://www.npmjs.com/package/detect-node?activeTab=code


taras (approved):
👍 🚢

PR #748, opened 2023-08-12
[v3] fix: context inheritence between parallel tasks

neurosnap:


lib/run/frame.ts
```diff
@@ -12,14 +12,15 @@ let ids = 0;
...
+  context?: Record<string, unknown>;
   parent?: Frame;
 }
 
 export function createFrame<T>(options: FrameOptions<T>): Frame<T> {
   let { operation, parent } = options;
   let children = new Set<Frame>();
   let block = createBlock(operation);
-  let context = Object.create(parent?.context ?? {});
+  let context = Object.create(options.context ?? parent?.context ?? {});
```
cowboyd: If there is a parent, then I still think that the parent context should be the prototype of this context. 

I think what we want is:

```ts
Object.create(parent?.context ?? {}, Object.getOwnPropertyDescriptors(options.context ?? {}))
```

Issue #751, opened 2023-08-24
Scope simplification


cowboyd:
The first simplification to the scopes API was to make each invocation of `run()` use its own frame. I propose two more:

- [x] Remove `Scope` as `Operation`. Right now, `Scope` implements `Operation`, but this is more easily handled by a task grouping construct such as [`parallel` in starfx](https://github.com/neurosnap/starfx/blob/main/fx/parallel.ts#L12)
- [x] remove `close()` from the `Scope` API, and pass it explicitly to the caller of `useScope()`. The following makes no sense.
```js
function* op() {
  let scope = yield* useScope();
  yield* scope.halt(); 
}
```
instead, `close` is only for the creator of the scope:
```js
let [scope, destroy] = createScope();
await scope.run(function*() {
  yield* sleep(100);
});
await destroy();
```
- [x] add a `set()` method to `Scope` to allow for setting context directly on the frame. This allows programmatically setting a context from outside Effection:
```js
scope.set(StoreContext, store);
scope.get(StoreContext) === store //=> true
```

In summary, scope becomes the way to interact with Effection from _outside_, and that is its only responsibility. The new interface for `Scope` will be:

```ts
export interface Scope {
  run<T>(operation: () => Operation<T>): Task<T>;
  get<T>(context: Context<T>): T;
  set<T>(context: Context<T>, value: T): T;
}
```

PR #753, opened 2023-08-24
♻️🔥Simplify the Scope API by removing "group-ish" functionality.

cowboyd:
## Motivation

Step one of https://github.com/thefrontside/effection/issues/751

## Approach

This rips out all the child tracking which is already the job of a frame. Also, because there is now a 1:1 relationship between frame and scope, there is also, by extension, a 1:1 relationship between frame and task, and also scope and task. We can make the task lazy, and just use it as our shutdown logic.


neurosnap (approved):
very nice refactor

PR #754, opened 2023-08-25
♻️Move the `close()` function from scope off of the Scope API.

cowboyd:
## Motivation

Part two of the scope improvements in #751 That is, closing a scope is an option available to the code that created the scope. Those who merely use the scope, do not need to be able to close it.

```js
let [scope, close] = createScope();
try {
  await scope.run(someOperation);
} finally {
  await close();
}
```

## Approach

Move the close function off of the scope

### Alternate Designs

I did consider also making the tuple returned by `createScope` an interface so that it could be captured as a single object and/or destructured.

```js
let newScope = createScope();
await newScope.run(myOp);
await newScope.close();
```

It would still allow restructuring:

```js
let { scope, close } = createScope();
```

Funny, now I'm kinda leaning towards this... The main reason to have the tuple instead of the record is that it is very easy to rename the function references in your destructuring assignment.

```js
let [testScope, teardown] = createScope();
```
vs
```js
let { scope: testScope, close: teardown } = createScope();
```

### DIFF
--- a/lib/run/scope.ts
+++ b/lib/run/scope.ts
@@ -1,4 +1,4 @@
-import type { Frame, Operation, Scope } from "../types.ts";
+import type { Frame, Future, Operation, Scope } from "../types.ts";
 import { create } from "./create.ts";
 import { createFrame } from "./frame.ts";
 import { getframe, suspend } from "../instructions.ts";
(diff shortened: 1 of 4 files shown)


neurosnap (approved):
I like the tuple style so I say we try it.

PR #755, opened 2023-08-25
✨get/set context values from `Scope` API

cowboyd:
## Motivation

> This is part 3 of 3 in #751 

Whenever you are creating a scope from scratch, you need to be able to interact with the context to make sure that things are available for children of that scope.

## Approach

This adds the `get/set` method to scope, and then switches Context to use them to do its job.


neurosnap (approved):
Very nice refactor

PR #757, opened 2023-08-25
✨ `race()`

cowboyd:
## Motivation

Ports race from v2

### DIFF
--- a/lib/mod.ts
+++ b/lib/mod.ts
@@ -21,3 +21,4 @@ export * from "./each.ts";
 export * from "./queue.ts";
 export * from "./signal.ts";
 export * from "./ensure.ts";
+export * from "./race.ts";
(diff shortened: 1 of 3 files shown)


neurosnap (commented):
One thing that is really nice with `race()` is the ability to pass an object of operations, that way you can use conditional logic for the result:

```ts
const result = yield* race({
  create: createUser,
  timeout: sleep(1000)
})

if (result.timeout) {
  console.log('this operation timed out!');
}
```

What are your thoughts on supporting a feature like that?

cowboyd:
I actually prefer this object-based API, but my tendency is to keep strictly to parity with the `Promise.race()` API and let enhanced APIs be built on top of that. This is something that we can always add later if we decide that there are no drawbacks.

neurosnap (approved):

PR #759, opened 2023-08-28
🔥Get rid of v2 packages that won't be part of 3.0.0

cowboyd:
## Motivation

We've ported the packages one by one while standing up version 3. However, now most of the time we're inspecting the code, it is sufficient to do it from a direct checkout of the `v2` branch, either via github, or on the local file system.

## Approach

This removes all the old code which will not be part of the `v3` release. There are elements that will come with 3.x, but as mentioned before, they can be brought over from other sources.


taras (approved):
🔥 let it burn!

PR #760, opened 2023-08-30
🐞Fix arguments passed to NodeJS

cowboyd:
## Motivation
The process argument handling is different from Deno and NodeJS. The following script in Deno:

``` shellsession
$ deno script.ts hello world
```

will set `Deno.args` to `["hello", "world"]`. By contrast, the following in NodeJS:

``` shellsession
$ node script.mjs hello world
```

will set `process.argv` to
`["/path/to/node","/abs/path/to/script.mjs", "hello", "world"]`

The NodeJS way is similar to the argv of a `Bash` script, but the Deno way is more practical.

## Approach

This normalizes Node to behave like Deno, only passing the user arguments to the script. We can explore how to get the script name and the executable name later.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -2,7 +2,7 @@
   "lock": false,
   "tasks": {
     "test": "deno test --allow-run=deno",
-    "test:node": "deno task build:npm 0.0.0 && node test/main/node.mjs",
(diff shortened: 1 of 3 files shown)


taras (approved):
👍

PR #771, opened 2023-09-29
Don't create a new frame just to wait for subscription.next()

cowboyd:
## Motivation
We were experiencing serious performance degradations when delivering a mere hundreds of messages in a short time. The reason was due to incessant garbage collection. This was because It was creating a "throw away" `Frame` whose only purpose was to sit there until the next message was received at which point it was dropped, and then a new "throw away" frame would be created the next time you pulled from the subscription.

```js
while (true) {
  let next = yield* subscription.next(); // <-- new Frame every time
  ...
}
```

But a `Frame` is a very heavyweight object just to hold up the current execution until it can be resumed. In cases where you have many message deliveries at once, you can end up creating and throwing away thousands of frames in a matter of milliseconds.

## Approach

With this change, the `Queue` subscription uses a new `pause` operation which wraps the _current frame_'s continuation (rather than allocating a completely new one). This means that there is one frame, and one generator per subscription, not thousands of frames, and thousands of generators per subscription.

```js
while (true) {
  let next = yield* subscription.next(); // <-- just pause the current generator
  ...
}
```

The `pause()` operation is left private to inside Effection for now because it's unclear what form its API will ultimately take, but it has several key limitations that make it different from an action:

1. it takes a synchronous function as its callback, not an operation.
2. It returns a function that reverses the effects of that function.

All it does is just pause and resume the _current_ iterator from the _current_ frame.


neurosnap (approved):
Massive improvement!

Issue #767, opened 2023-09-22
Plan for Migrating Effection Documentation from v2 -> v3


cowboyd:
With the upcoming release of v3, there are some significant changes to the way that the documentation will work. For example, the guides have been completely re-written, and now use a simple HTML server instead of doc-o-saurus, and so it is not just a matter of updating that source code. Additionally, we don't want to just pull down the v2 docs and replace it with the v3 docs, we want to make sure that all the docs are available to all users of both major versions of the software.

In short, we need a plan to house both sites at the same time and make sure that we can navigate between them, and that navigation links are not broken. Below are the tasks that are needed to make this happen.

- [x] https://frontside.com/effection points to v3 website.
  - [x] v3 website contains an item in the header `v2` to link to the v2 website.
- [x] https://frontised.com/effection-v2 points to the v2 website
  - [x] v2 website is updated with a banner that indicates that the v2 website is here for historical purposes, but will not be updated after January 2024
- [x] Update v2 api docs 
  - [x] fix links for existing site https://github.com/thefrontside/effection/issues/766
  - [x] new place for site is effection/v2/apidocs/
- [x] update the Open Source switcher to point to effection v3
  - [x] update switcher on interactors
  - [x] update switcher on effection
  - [x] update switcher on BigTest

In order to offload some of the complexity of maintaining typescript documentation, we are going to just use the API docs that are already [generated automatically by deno.land](https://deno.land/x/effection)

PR #786, opened 2023-10-05
Mount V2 docs at /V2/ subpath in Effection V3 site

cowboyd:
## Motivation

As part of https://github.com/thefrontside/effection/issues/767, we want to make sure that V2 documentation is always accessible to users as they migrate.

## Approach

Rather than maintain three websites and three deployments (V3, V2, and V2-apidocs), two on netlify, and the other deno.com, this consolidates all three into a single, simple server on deno.com

The html for the V3 site is served via html templates, the `V2` sites are served by downloading [pre-built versions from Github Releases](https://github.com/thefrontside/effection/releases/tag/docs-v2-r3) as `.tgz` files. Once downloaded, they are kept in memory until the server is restarted so serving is as fast, or faster than from the file system.

There is a slight wrinkle in Docusaurus (the docs tool used for v3) in that it cannot handle being mounted at different base urls without having to re-build, so there is a `local.tgz` version which assumes `/V2` as the basis for the V2 site, and `prod.tgz` which is built with `/effection/V2/` as its base for the `v2 site` since it will be served from `https://frontside.com/effection/V2`

### Previews

- **V2** https://effection--mount-v2-docs-onto-v3.deno.dev/V2/
- **V2 API** https://effection--mount-v2-docs-onto-v3.deno.dev/V2/api/
- **V3** https://effection--mount-v2-docs-onto-v3.deno.dev/

### Implementation Notes

The portion that actually serves the tar file, was taken directly from the [`serveDir()`](https://deno.land/std@0.203.0/http/file_server.ts?s=serveDir)  function from the Deno standard library, and just adapted to access the tarball in every place where it would access the file system.

### DIFF
--- a/www/deno.json
+++ b/www/deno.json
@@ -1,6 +1,6 @@
 {
   "tasks": {
-    "dev": "deno run -A lib/watch.ts html docs lib -- deno run -A --lock-write main.ts",
+    "dev": "deno run -A lib/watch.ts html docs lib hooks -- deno run -A --lock-write main.ts",
     "build": "deno run -A docs/build.ts"
   },
   "lint": {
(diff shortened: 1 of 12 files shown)


taras (approved):
👍

PR #787, opened 2023-10-06
Add OSS project selector to v3 website

cowboyd:
## Motivation
We need a way to navigate from Effection V3 to our other open source projects, especially the Effection V2 for now.

## Approach
This adds an open source swither that is based very closely on the open source switcher that is on the main site. It links to Interactors, Effection V2, and the Auth0 Simulator.

It doesn't highlight V2, but it does at least provide navigation to it. We can remove the V2 link after some time.

### Screenshots

image

image


taras (approved):
👍

PR #789, opened 2023-10-06
📄Add more typedocs

cowboyd:
## Motivation

Now that interfaces are getting more defined, we should add typedocs for them, which is necessary for a public release

## Approach

This also removes the `log.ts` APIs which were always very theoretical.


lib/types.ts
```diff
@@ -42,9 +44,86 @@ export interface Operation<T> {
...
+ *
+ * A task can also be created from within an operation by using the
+ * {@link spawn()} operation.
+ *
+ * @example
+ * ```javascript
+ * import { run, spawn, sleep } from "effection";
+ *
+ * await run(function*() {
+ *   let task = yield* spawn(function*() {
```
neurosnap, line 86: I'm not sure if this is mentioned elsewhere, but it needs to be made very clear that tasks **must** be yielded/awaited-to or else when the parent task completes it will shut down the child task whether the child task is completed or not.

I'm not sure if there's a design principles doc but that should be a bullet point in it.

cowboyd: Yeah, it's definitely a subtlety to it. What about this section? https://effection.deno.dev/docs/scope

I think we could add it to that section (which is incomplete) and then link to it from the type docs. Would that work?

cowboyd: @neurosnap I added a quick reference to the scoping guide. I think we can update the scoping guide too, but that would be in a later PR.

taras (approved):

PR #790, opened 2023-10-10
Cleanup remaining V2 docs issues.

cowboyd:
## Motivation

We really want to launch the V3 doc site.

## Approach

This updates several things to be in line with the existing doc site:

1. fix an issue with the OSS links where they were overlaying the
2. Use the normal Frontside fonts
3. Make the footer nice.
4. make the Previous and Next navigation nice.
5. Make the topic navigation more like the old site with the hover effects

### Screenshots

before:

image

after:

image


taras (approved):
👍 looks pretty 😍

PR #801, opened 2023-10-23
Add cause to error.ts

lc-mm:
## Motivation

Maintain a reference to the original error

## Approach

Add the `cause` attribute

### DIFF
--- a/.changes/add-cause-property.md
+++ b/.changes/add-cause-property.md
@@ -1,0 +1,4 @@
+---
+"@effection/core": patch
+---
+Add the `cause` property to all wrapped errors
--- a/.github/workflows/covector-status.yaml
+++ b/.github/workflows/covector-status.yaml
@@ -14,7 +14,8 @@ jobs:
       - run: yarn install --immutable
       - run: yarn prepack
       - name: covector status
-        uses: jbolda/covector/packages/action@covector-v0.9
+        uses: jbolda/covector/packages/action@main
         id: covector
         with:
           command: "status"
+          comment: true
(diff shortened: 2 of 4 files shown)


cowboyd:
@lc-mm Before I write a test case for this, have you confirmed that this will work for you by linking it into your codebase and attempting to run it?

lc-mm:
Yes I am running this code in my production app now.

cowboyd (approved):

cowboyd:
> Yes I am running this code in my production app now.

@lc-mm Fantastic! If I may ask, how long have you been using Effection, and how do you find using it?

lc-mm:
> > Yes I am running this code in my production app now.
> 
> @lc-mm Fantastic! If I may ask, how long have you been using Effection, and how do you find using it?

I chose Effection 2-3 months ago to build out a new feature. I wanted to use generators because I had good experiences in the past using redux saga. So far I have enjoyed using it. I did choose to go with v2 until v3 is out of beta.

On a side note I am using it in react native and did have to add this in order to start using it:

```
window.addEventListener = (x: unknown) => x;
window.removeEventListener = (x: unknown) => x;
```

cowboyd:
Interesting. I'm guessing it's because the main() method? https://github.com/thefrontside/effection/blob/v2/packages/main/src/browser.ts#L33

What would we do to detect the react native app unloading?

Also, please let us know if there's any questions you have, or feedback. We're always looking for ways to improve.

lc-mm:
There is no way that I know of to detect the app closing or unloading, You can detect application states such as foreground or background. When the user closes an app the execution stops without any notice

PR #808, opened 2023-10-30
fix links in typedocs

cowboyd:
## Motivation
We had a bunch of incorrect usages of the `@link` tag. You can't use it for arbitrary links. Also, only simple, direct JavaScript symbols.

## Approach

This fixes the bad usages, and replaces the external links with normal markdown links.

## Screenshots

example of a malformed link 👇🏻 

Image showing malformatted links


taras (approved):
👍

PR #823, opened 2023-11-07
precompute list of topics ahead of time

cowboyd:
## Motivation

The order of topics on the left sidebar was non-deterministic since the topic became defined by the very first doc that was successfully parsed.

## Approach

This pre-computes the list of topics and their order ahead of time from the structure module, and then the content of the topic is filled in as docs are loaded.


taras (approved):
👍

PR #825, opened 2023-11-08
Express `each()` as a function

cowboyd:
## Motivation

We were expressing the `each()` operation as a `const` reference. Honestly, I can't recall why this was done other than some vague connection to TypeScript. However, it makes the typedocs weird because `each` is represented as a `var`, not a `function`.

## Approach

This switches `each()` to be a function and fixes the typedoc at the same time.


neurosnap (approved):

taras (approved):
👍

PR #826, opened 2023-11-09
feat(signal): allow custom `Queue` impl when creating a `Signal`

neurosnap:
## Motivation

`starfx` needs a way to customize the implementation for a `Signal`'s `Queue` so we can control how actions are dispatched and subscribed to, primarily for performance reasons.

It is not uncommon for a FE app to have hundreds of listeners on the action `Stream` so this is a hot path that is causing performance issues for [app-ui](https://github.com/aptible/app-ui).

## Approach

We need a way to allow users to customize the `Queue` implementation to support features like buffering, dropping messages, filtering messages, etc.  Ideally, we would be able to implement these downstream of a `Stream` but we are noticing performance implications with that implementation.  

So this is not ideal, but until we dig into why there's such a big performance impact for sending and listening to a lot of messages, we opted to allow users to monkey-patch `Queue` with something custom.  As a result, we decided to create `Context` object that defaults to `createQueue` but can be overridden when necessary.

See discussion: https://discord.com/channels/700803887132704931/1169297115609059411

The biggest drawback here is we are not really solving the root issue with performance, instead we came up with a solution that doesn't change the API contract while still letting end-users monkey-patch `Queue`.

### TODOs and Open Questions

- Why is the performance poor when having a lot of subscribers on a `Signal`?

## Screenshots

Listed below are screenshots from a production FE application using `effection` and `starfx`.  As you can see, this PR removes all "slow tasks" in the profiler.

### BEFORE
Screenshot 2023-11-09 at 10 50 31 AM

### AFTER
Screenshot 2023-11-09 at 10 46 35 AM

### DIFF
--- a/lib/signal.ts
+++ b/lib/signal.ts
@@ -2,6 +2,7 @@ import type { Stream, Subscription } from "./types.ts";
 
 import { createQueue, type Queue } from "./queue.ts";
 import { resource } from "./instructions.ts";
+import { createContext } from "./context.ts";
 
 /**
  * Convert plain JavaScript function calls into a {@link Stream} that can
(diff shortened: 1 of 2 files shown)


cowboyd (approved):
Looks good!

PR #829, opened 2023-11-12
Revolutionize the website

cowboyd:
## Motivation

I've extracted the HTML framework that is driving the Effection website and the frontside.com website into a separate repo. That way we don't have to copy and paste between the two. It's also this framework that will, in all probability, drive the v3 inspector.

## Approach

It's a middleware stack that is divided into three layers: `jsx`, `html`, and `http`. The JSX stack maps `Request` to JSX elements, then, once you have your html document, the `html` stack transforms the  HTML as a syntax tree, and finally there is the `http` layer which lets you operate on `Request` -> `Response`

By having a proper middleware stack, this will let us have interactive components, and it will also let us perform caching at the HTTP layer.

It's a big change, with plenty of deletions, but here is the core of it:

```ts
  let app = createRevolution({
    jsx: [
      route("/", indexRoute()),
      route("/docs/:id", docsRoute(docs)),
    ],
    html: [
      rebaseMiddleware(),
      twindRevolution({ config }),
    ],
    http: [
      route("/V2(.*)", v2docsRoute(v2docs)),
      baseUrlMiddleware(),
    ],
  });
```

## Screenshots

image


taras (commented):
Can we combine `main.tsx` into `main.ts`?

cowboyd:
@taras deno deploy only allows one entry point at a time, so once this is merged we can get rid of `main.ts`

taras (approved):
🔥

PR #832, opened 2023-11-16
feat: `call` accepts plain functions

neurosnap:
## Motivation

As library developers, we might want to require the user to understand delimited continuations or any of the constructs created by `effection`.  Sometimes we want to let the end-user provide anything reasonable we `call()` knows how to reflect what was passed into the function and it will handle it automatically.  The most obvious example is `Promise`.  We don't have to require the user to convert a `Promise` into an `Operation`.

Being able to pass a vanilla function into `call()` would also be nice since it is trivial for us to detect and accept.

So in summary, we want to support the following variables to `call()`:

```ts
call(() => thing);
call(async () => thing);
call(function*() {});
```

## Approach

We try to inspect the type of the variables passed into `call()` and figure out how to evaluate it.

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -1,20 +1,23 @@
-import type { Operation } from "./types.ts";
+import type { Instruction, Operation } from "./types.ts";
 import { action } from "./instructions.ts";
 import { pause } from "./pause.ts";
 
 /**
- * Pause the current operation, and run a promises, async functions, or
- * operations within a new scope. The calling operation will resumed (or errored)
+ * Pause the current operation, then runs a promise, async function, plain function,
+ * or operation within a new scope. The calling operation will be resumed (or errored)
  * once call is completed.
  *
  * `call()` is a uniform integration point for calling async functions,
- * evaluating promises; as well as generator functions and operations. It can be
- * used to treat a promise as an operation.
+ * evaluating promises, generator functions, operations, and plain
+ * functions.
+ *
+ * It can be used to treat a promise as an operation:
  *
  * @example
  * ```js
  * let response = yield* call(fetch('https://google.com'));
  * ```
+ *
  * or an async function:
  *
  * @example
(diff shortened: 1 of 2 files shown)


lib/call.ts
```diff
@@ -43,19 +62,28 @@ import { pause } from "./pause.ts";
...
-export function call<T>(operator: () => Operation<T>): Operation<T>;
-export function call<T>(operator: () => Promise<T>): Operation<T>;
-export function call<T>(operator: Operation<T>): Operation<T>;
-export function call<T>(operator: Promise<T>): Operation<T>;
-export function call<T>(operator: Operator<T>): Operation<T> {
+export function call<T>(callable: Operation<T>): Operation<T>;
+export function call<T>(callable: () => Operation<T>): Operation<T>;
+export function call<T>(callable: () => Promise<T>): Operation<T>;
+export function call<T>(callable: () => T): Operation<T>;
+export function call<T>(callable: Callable<T>): Operation<T | Iterable<T>> {
```
cowboyd: Why might this return an Operation<Iterable<T>>? If you have `call(() => [])` which override will it select? The one on line 68 or 69?

neurosnap: You're right, removed.  Probably old code from when I was experimenting.

lib/call.ts
```diff
@@ -2,6 +2,25 @@ import type { Operation } from "./types.ts";
 import { action } from "./instructions.ts";
 import { pause } from "./pause.ts";
 
+function isFunc(f: unknown): boolean {
+  return typeof f === "function";
+}
+function isPromise(p: unknown): boolean {
+  if (!p) return false;
+  return isFunc((p as PromiseLike<unknown>).then);
+}
```
cowboyd: Can we use guards where possible? It saves a little casting when you use them.

```suggestion
function isPromise<T>(p: unknown): p is Promise<T> {
  if (!p) return false;
  return isFunc((p as Promise<T>).then);
}
```

lib/channel.ts
```diff
@@ -56,8 +56,8 @@ export function createChannel<T, TClose = void>(): Channel<T, TClose> {
   let signal = createSignal<T, TClose>();
 
   return {
-    send: lift(signal.send),
-    close: lift(signal.close),
+    send: (p) => call(() => signal.send(p)),
+    close: (p) => call(() => signal.close(p)),
```
cowboyd: Since we can't get rid of `lift`, let's continue to use it here. We can revisit the best way to do this later, but `lift` is not going to create a frame which is preferable.

cowboyd (commented):
I think this is the best option we have. As noted, there is a small discrepancy in the types and the runtime behavior in the event that you `call(() => instructions)` where instructions is an array of `Instruction`.

This is a very tiny edge case.

The only remaining decision is whether to support bare values. In other words, should callable be:

```ts
| (() => T))
| (() => Promise<T>))
| (() => Operation<T>))
```
or

```ts
| T
| Operation<T>
| Promise<T>
| (() => T))
| (() => Promise<T>))
| (() => Operation<T>))
```

It feels like we should support one or the other, but not somewhere in between.

test/call.test.ts
```diff
@@ -48,4 +58,42 @@ describe("call", () => {
...
+  it("evaluates a vanilla function that returns an iterator", async () => {
+    function* eq() {
+      return "value";
+    }
+    await expect(run(function* () {
+      return yield* call(() => eq);
+    })).resolves.toEqual(eq);
+  });
+
+  it("evaluates a string iterator", async () => {
```
cowboyd: I wouldn't call this a string iterator, but just a function returning a string.

lib/call.ts
```diff
@@ -43,35 +62,69 @@ import { pause } from "./pause.ts";
...
 
-type Operator<T> =
+export type Callable<T> =
   | Operation<T>
+  | Iterator<T>
   | Promise<T>
+  | T
   | (() => Operation<T>)
-  | (() => Promise<T>);
+  | (() => Iterator<T>)
```
cowboyd: Shouldn't this be `Iterator<Instruction, T>`?

lib/call.ts
```diff
@@ -1,20 +1,42 @@
...
+function isInstructionIterator<T>(it: unknown): it is Iterator<Instruction, T> {
+  if (!it) return false;
+  return isFunc((it as Iterator<Instruction, T>).next) &&
+    isFunc((it as Iterator<Instruction, T>).throw);
+}
+function isIterable<T>(it: unknown): it is Iterable<T> {
+  if (!it) return false;
+  return typeof (it as Iterable<T>)[Symbol.iterator] === "function";
+}
+
```
cowboyd: Can we move these private functions to the bottom of the file and lead with the implementation of `call()`?

lib/call.ts
```diff
@@ -1,20 +1,42 @@
...
 /**
  * Pause the current operation, and run a promises, async functions, or
  * operations within a new scope. The calling operation will resumed (or errored)
```
cowboyd: Looks like there's a slight grammatical error here. Might as well fix this now.

```suggestion
 * Pause the current operation, and run a promise, async function, or
 * operation within a new scope. The calling operation will resumed (or errored)
```

cowboyd (approved):
Looks good!

test/call.test.ts
```diff
@@ -22,6 +23,15 @@ describe("call", () => {
     )).resolves.toEqual(42);
   });
 
+  it("evaluates async function", async () => {
```
cowboyd:
```suggestion
  it("evaluates an async function", async () => {
```

Issue #833, opened 2023-11-17
Finalize v3 docs


cowboyd:
Some gaps still prevent a coordinated publicity push to raise people's awareness about Effection in particular and structured concurrency in general. 

This issue is a clearing house for all those gaps. They should be the minimal list of things that we want to have in place before officially announcing Effection v3.

- [x] is the organization of the material correct
- [x] Reorganize navigation 
- [x] update the section on TypeScript: explain how it mostly works but that you will see `Generator<Instruction, T, any>` and correcting via `Operation<T>` 
- [x] Move Rosetta Stone section to another location, includes integrating with `async/await`
- [x] fetch examples should use abort signal (since we no longer export a global fetch)
- [x] Update all code samples for V3
- [x] Remove inspector page https://github.com/thefrontside/effection/pull/843
- [x] Add installation page https://github.com/thefrontside/effection/pull/844
- [x] Review TypeScript page
- [x] Review Processes page
- [x] Review and update Testing Page
- [x] Update the introduction tutorial
- [x] Add a message that Effection can be used in Node, Deno and Browser
- [x] Verify Github Readme links to correct documentation website
- [x] Update Main Navigation

Descoped:
-  Revisit how to explain the `TClose`, `TDone`, `TReturn` type parameter.
-  Add React Example
-  upgrading from `V2`
-  Add code examples to Features and Benefits of Effection
- Add Context to the documentation
- Add Scope Section to documentation

cowboyd:
Some feedback from early testers of V3

from @nelstrom:
> To make the documentation clearer, maybe you could replace TClose with "the value of the IteratorResult when done: true"? Or alternatively, add an entry to the docs that defines TClose?
> Just noticed that TClose is defined as a type parameter in a couple of places:
> https://deno.land/x/effection@3.0.0-beta.1/mod.ts?s=Signal
> https://deno.land/x/effection@3.0.0-beta.1/mod.ts?s=Queue
>
>Another confusing thing about the documentation for each:
>
> Note that just as with async iterators, there is no way to consume the TClose value of a stream using the for-each loop.
> What's a `TClose` in this context? It's not mentioned anywhere else. 
> If the documentation generator doesn't make it easy to write that documentation once, I'd be in favour of repeating the information where needed. As coders, we put a high value on keeping things DRY. But when it comes to prose, you've got to repeat yourself sometimes to be understood.

from @bdougherty

[...]

PR #834, opened 2023-11-20
Adding basic fetch docs

taras:
## Motivation

As part of #833, we must provide a documented fetch alternative to update all the docs where Fetch is used incorrectly.

## Approach

Started implementing a fetch module, and wrote initial tests and documentation.

1. `createFetchOperation` takes `fetch` and returns an operation that automatically takes care of handling graceful abort.
2. Wrote tests for `createFetchOperation`
3. Added a new section to the docs called Standard Library with page for fetch
4. Wrote docs about `createFetchOperation`

## TODO

* [ ] Implement useFetch
* [ ] Provide FetchContext
* [ ] Write tests for both
* [ ] provide documentation on how to stub fetch when using `useFetch`

### DIFF
--- a/lib/fetch.ts
+++ b/lib/fetch.ts
@@ -1,0 +1,27 @@
+import { useAbortSignal } from "./abort-signal.ts";
+import { action } from './instructions.ts';
+import { call } from './call.ts';
+
+export function createFetchOperation(fetchFn: typeof fetch) {
+  return function* fetchOperation(
+    input: RequestInfo | URL,
+    init?: Omit<RequestInit, "signal">,
+  ) {
+    return yield* action<Response>(function* (resolve, reject) {
+      const signal = yield* useAbortSignal();
+
+      try {
+        let result = yield* call(
+          fetchFn(input, {
+            ...(init ?? {}),
+            signal,
+          }),
+        );
+        resolve(result);
+      } catch (e) {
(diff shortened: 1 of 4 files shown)


cowboyd:
I really like the concept of making a "fetch middleware" to automatically apply authorization tokens that are context sensitive. However, I'm hesitant to add that in without a lot of thought.

An under-pinning philosophy of `v3` is that the apis should feel as 1:1 of a mapping from vanilla javascript to effection flavored javascript so that the transition in thinking. For example, that is why all of the streaming apis had to change in order that they align with the native async iterable apis, and could be consumed with simple iteration.

One of the nice things about the W3C fetch API is that it's just there, and you can use it with a single step. This makes that a two-step process of first instantiating the fetch operation, and then calling it. Is there a way that we can have this happen in a single step?

```ts
import { main, useFetch } from "effection";

await main(function*() {
  let response = yield* useFetch("https://frontside.com");
  console.log(yield* response.text());
});
```

```ts
yield* FetchFunction.set(function*(request, init, next) {
  let token = yield* AuthTokenContext;
  return yield* next(request, {
    ...init,
      headers: {
        ...(init?.headers ?? {}),
        ...(token ? { "Authorization": `token ${token}` } : { })
      }
  });
});
```

PR #845, opened 2023-12-13
🔥Remove stream combinators

cowboyd:
## Motivation

Stream transformations are not our core competency. If Effection is measured by that yard stick, it'll come up short because what we have is 5% implementation of a not-strongly-opinionated stream helpers library. One could very reasonably ask "well where is flatMap(), merge(), reduce(), reduceRight()" and on and on. The answer is that we aren't in that business, and it's better to not play that game at all, than to enter the race and get lapped 20 times.

Our mantra is "Structured Concurrency and Effects for JavaScript". It's a minimalist approach that seeks to excel via maximum compatibility and minimum friction over the long term. Having these as part of our API surface feels like it runs counter to that principle.

## Approach
In the spirit of keeping things "core", remove first(), filter(), map(), pipe(), and associated documentation.

They should be published as a separate library where the optimal implementation (if there is one) can be discovered.


taras (approved):
👍 Let's 🪓

taras (approved):
👍

neurosnap (approved):
We aren't using any of these in starfx so go for it!

PR #846, opened 2023-12-13
Thinking in Effection page and reorganize page

taras:
## Motivation

We want to give people what they need to know to start Effection.

## Approach

Create "Thinking in Effection" page and moved Async Roseta Stone to that page.

## Preview

https://effection--tm-reoganize-sections.deno.dev/docs/thinking-effection

## Questions

* [x] I'm not happy with the example in `Every function returns or throws` section. It can be said that the same can be accomplished with async await.


cowboyd:
How about this as an intro?

> When we say that Effection is "Structured Concurrency and Effects for Javascript" we mean three things:
> 
> 1. No operation runs longer than its parent.
> 2. Every operation exits fully.
> 3. It's just JavaScript, and except for the guarantees derived from (1) and (2), it should feel familiar in every other way.
>
> Developing a new intuition about how to leverage Structured Concurrency, while leaning on your existing intuition as a JavaScript developer will help you get the most out of Effection and have you attempting things that you would never have even dreamed anymore.

taras:
> 1. No operation runs longer than its parent.
> 2. Every operation exits fully.
> 3. It's just JavaScript, and except for the guarantees derived from (1) and (2), it should feel familiar in every other way.

Are these the "Three Laws of Structured Concurrency with Effection"?

cowboyd (approved):

cowboyd:
I'd remove the `.gif` if it isn't going to be used

PR #847, opened 2023-12-13
🔥Remove transform example from event docs.

cowboyd:
## Motivation

We had several defunct references to the stream helpers.

## Approach

🔥🔥


taras (approved):
👍

PR #857, opened 2023-12-15
Tweak CSS

taras:
## Motivation

Make it look better

## Approach

1. Made footer show up under article
2. Made footer smaller
3. Adding padding between sections
4. Made TOC wrap


cowboyd:
I'm going to go ahead and approve because CSS is really, really difficult to get right and you can fall down a rabbit hole of reorganizing page structure, but we do need to revisit the footer since it isn't really a footer anymore.

image

In order to match with the page footer it should:

1. span all content on the page
2. appear on every page

cowboyd (approved):

PR #859, opened 2023-12-16
♻️ Return to Stream as Operation

cowboyd:
## Motivation

Now that we have` each()`, you end up writing subscriptions more often than you use them. The idea behind subscribe() was that yielding directly to a stream to instantiate a subscription was confusing, which at first blush is still true.

However, the decision to add a subscribe() method directly was made on the assumption that you would be using subscriptions by hand more often and so a little more ceremony around authoring subscriptions was warranted.

That doesn't seem to hold anymore now that we have each() since it's almost always the right thing to do instead of using to the subscription directly. The only case is when you're doing meta-level stuff in which case, understanding the underlying stream/subscription mechanics as a precursor is ok.

For example, adapting to this event emitter goes from:

```js
export function onEmit(emitter, name) {
  return {
    subscribe() {
      return resource(function*(provide) {
        let signal = createSignal();
        emitter.on(name, signal.send);
        try {
          yield* provide(yield* signal.subscribe());
        } finally {
          emitter.off(name, signal.send);
        }
      });
    }
  }
}
```

to this:

```js
export function onEmit(emitter, name) {
  return resource(function*(provide) {
    let signal = createSignal();
    emitter.on(name, signal.send);
    try {
      yield* provide(yield* signal);
    } finally {
      emitter.off(name, signal.send);
    }
  });
}
```

## Approach

Optimize for making streams easy to define


taras (approved):
👍 very nice

PR #868, opened 2023-12-21
Fix bad example codeblock in call.ts

wms:
## Motivation

Allows `call` example to render correctly on docs browser

## Approach

Self-explanatory


taras (approved):
thank you for your contribution

cowboyd (approved):

PR #871, opened 2023-12-23
chore: update sleep.ts example

janeklb:
## Motivation

The jsdoc `@example` snippet is probably leftover from v2 since it's using `yield` instead of `yield*`. Same goes for the [scope](https://frontside.com/effection/docs/scope) and [streams and subscriptions](https://frontside.com/effection/docs/collections) (for `subscription.next()`) docs.

### DIFF
--- a/lib/sleep.ts
+++ b/lib/sleep.ts
@@ -9,7 +9,7 @@ import { action, suspend } from "./instructions.ts";
  * import { main, sleep } from 'effection';
  *
  * await main(function*() {
- *   yield sleep(2000);
+ *   yield* sleep(2000);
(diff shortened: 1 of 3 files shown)


cowboyd (approved):
@janeklb Thanks for taking the time to correct this!

janeklb:
My pleasure! Awesome library - really appreciate all your thoughts (posts) around structured concurrency.

ps. Had to push again as I hadn't been signing my commits

cowboyd:
The fixes for the guides are live. The type docs are versioned with the release of the code, and so will appear with the next release. Again, thanks!

PR #879, opened 2024-01-06
Lift a simple function into an operation safely.

cowboyd:
## Motivation

There is the potential for a function, if it is a continuation, to cause the scope in which it is called to be destroyed. In this case, nothing after that function should ever be called. However, because of the way `lift()` is currently implemented, this will not be the case.

## Approach

This unpacks the `lift()` function just a bit to return an operation with a single instruction that immediately returns the result of invoking the function. However, because there is a `yield` point in the computation, it is an opportunity to not continue which will happen if corresponding frame is closed as a result of calling the function.

You can see in the test case how when the spawned producer closes the signal, it cause its parent to finish up and return, meaning that the producer should shut down immediately and not continue executing.

At the same time, it removes the named types since they don't really add much clarity when you look at them from your language server.

### DIFF
--- a/lib/lift.ts
+++ b/lib/lift.ts
@@ -1,3 +1,4 @@
+import { shift } from "./deps.ts";
 import { type Operation } from "./types.ts";
 
 /**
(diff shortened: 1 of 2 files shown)


lib/lift.ts
```diff
@@ -20,9 +21,12 @@ export function lift<TArgs extends unknown[], TReturn>(
 ): LiftedFn<TArgs, TReturn> {
   return (...args: TArgs) => {
     return ({
-      [Symbol.iterator]() {
-        let value = fn(...args);
-        return { next: () => ({ done: true, value }) };
+      *[Symbol.iterator]() {
+        return yield () => {
+          return shift(function* (k) {
+            k({ ok: true, value: fn(...args) });
```
neurosnap: should this be `k.tail` since you don't care about the return value of `k`?

cowboyd: Nice catch!

neurosnap (approved):

Issue #869, opened 2023-12-22
Effection examples


taras:
We must provide examples of how Effection can be used for different things. 

For frontend the conders are,
* form validation
* autocomplete

1. React
    - [Loading spinner example](https://github.com/taras/effection-react-loadingspinner) 
    - [Autocomplete example](https://github.com/taras/effection-react-autocomplete)
3. Vue
4. CLI
5. Express service

**What examples would you like to see?**

Atrue:
@taras What do you think about adding this file to the library https://github.com/taras/effection-react-autocomplete/blob/main/src/hooks/useTask.ts? I suppose it's a common case for React to run a task in the `useEffect`. Maybe it's worth adding something like a common library for React, like `effection/react`.
One more benefit is to add an eslint rule for the deps. Actually, the only thing is needed is:
```json
"rules": {
      "react-hooks/exhaustive-deps": ["warn", {
        "additionalHooks": "useTask"
      }]
    }
```
It makes the linter to check all the deps the same way as it does for useEffect and etc.

<blockquote><img src="https://opengraph.githubassets.com/80187b72d49b16096121dbd721e625e7294307ab1c4b68aeef4d8044e06eacd5/taras/effection-react-autocomplete" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/taras/effection-react-autocomplete/blob/main/src/hooks/useTask.ts">effection-react-autocomplete/src/hooks/useTask.ts at main · taras/effection-react-autocomplete</a></strong></div><div>Contribute to taras/effection-react-autocomplete development by creating an account on GitHub.</div></blockquote>

Atrue:
@taras I’ve made a repo comparing effection with async approach in some react examples https://github.com/Atrue/react-concurrency-examples/tree/main

Currently, I see 3 additional hooks that may be useful for react developers:

`useTaskEffect` - https://github.com/Atrue/react-concurrency-examples/blob/main/src/hooks/useTaskEffect.ts. It’s similar to https://github.com/taras/effection-react-autocomplete/blob/main/src/hooks/useTask.ts, but I think it’s better to have `effect` in the name as it’s run in the `useEffect` hook.

[...]

PR #881, opened 2024-01-08
Retry backoff example

taras:
## Motivation

As part of #869, I wanted to convert an example by [@thr1ve. in Discord](https://discord.com/channels/795981131316985866/1125094089281511474/1189299794145988748) to Structured Concurrency with Effection.

> * only retry if the error matches specific error/s (otherwise fail)
> * begin with a delay of startDelay ms between retries
> * double the delay each retry attempt up to a max of maxDelay ms
> * retry a maximum of 5 times

This is what it looks like in Effect.ts

```ts
const startDelay = 200
const maxDelay = 500

export const retryPolicy = pipe(
  Schedule.intersect(
    Schedule.linear(startDelay).pipe(Schedule.union(Schedule.spaced(maxDelay))),
    Schedule.recurs(5)
  ),
  Schedule.whileInput(
    (error: ParsedArangoError) =>
      error._tag === 'LockTimeout' ||
      error._tag === 'ConflictDetected' ||
      error._tag === 'TransactionNotFound'
  )
);
```

image

It was pretty straightforward and fun to implement in Effection. 

The logic is this:
1. call the operation, if we get an unknown error then throw which will stop the operation
2. otherwise, start retrying up to 5 times. After every attempt, double the delay before the next attempt. 
3. To add a maxDelay, I created a timeout function that I'm racing against the retry mechanism.

This is what it looks like in Effection

```ts
export function* retryBackoffExample<T>(
  op: () => Operation<T>,
  { attempts = 5, startDelay = 5, maxDelay = 200 } = {},
): Operation<T | undefined> {
  try {
    return yield* op();
  } catch (e) {
    if (!isKnownError(e)) {
      throw e;
    }
  }
  
  let lastError: Error;
  function* retry() {
    let delay = startDelay;
    for (let attempt = 0; attempt < attempts; attempt++) {
      yield* sleep(delay);
      delay = delay * 2;

      try {
        return yield* op();
      } catch (e) {
        if (isKnownError(e)) {
          lastError = e;
          if (attempt + 1 < attempts) {
            continue;
          }
        }
        throw e;
      }
    }
  }

  function* timeout(): Operation<undefined> {
    yield* sleep(maxDelay);

    throw lastError || new Error('Timeout');
  }

  return yield* race([retry(), timeout()]);
}
```

It's longer, but it doesn't require tears.

## Approach

1. Created examples directory with `retry-backoff.ts` example. 
2. Created a test file

## TODO:

* [ ] I don't know how to test the backoff mechanism without module stubbing which I really don't want to do.

### DIFF
--- a/examples/retry-backoff.test.ts
+++ b/examples/retry-backoff.test.ts
@@ -1,5 +1,5 @@
-import { Task, run } from "../mod.ts";
-import { beforeEach, describe, expect, it } from "../test/suite.ts";
+import { Task, run, sleep } from "../mod.ts";
+import { afterEach, beforeEach, describe, expect, it } from "../test/suite.ts";
 import { CustomError, retryBackoffExample } from "./retry-backoff.ts";
 /**
  * Requirements:
@@ -9,7 +9,7 @@ import { CustomError, retryBackoffExample } from "./retry-backoff.ts";
  *  * retry a maximum of 5 times
  * Source: https://discord.com/channels/795981131316985866/1125094089281511474/1189299794145988748
  */
-describe("retry-backoff", () => {
+describe("retry-backoff-example", () => {
   it("fails for unknown errors", () => {
     expect(run(function* () {
       yield* retryBackoffExample(function* () {
@@ -22,17 +22,20 @@ describe("retry-backoff", () => {
     describe("known errors", () => {
       let task: Task<void>;
       let retries: number;
-      beforeEach(() => {
-        retries = 0;
+      beforeEach(async () => {
+        retries = -1;
         task = run(function* () {
           yield* retryBackoffExample(function*() {
             retries++;
             throw new CustomError('LockTimeout');
           });
         });
+        try {
+          await task;
+        } catch {}
       });
-      it("rejects to a known error", () => {
-        expect(task).rejects.toEqual({
+      it("rejects to a known error", async() => {
+        await expect(task).rejects.toEqual({
           _tag: 'LockTimeout'
         });
       });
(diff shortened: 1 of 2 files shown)


cowboyd:
I'd omit the race since it can just be part of the flow of the operation:

```ts
export function* retryBackoffExample<T>(
  op: () => Operation<T>,
  { attempts = 5, startDelay = 5, maxDelay = 200 } = {},
): Operation<T> {

  yield* spawn(function* () ) {
    yield* sleep(maxDelay);
    throw new Error('timeout');
  });

  let currentDelay = startDelay;
  let lastError: Error;
  for (let attempt = 0; attempt < attempts; attempts++) {
    try {
      return yield* op();
    } catch (e) {
      if (!isKnownError(e)) {
        throw e;
      } else {
        lastError = e;
        yield* sleep(currentDelay);
        currentDelay = currentDelay * 2;
      }
  }
  throw lastError;
}
```

taras:
I initially wrote it this way but it didn't match the requirements

1. The maxDelay would apply to the entire operation, not only the error handling
2. Retry attempts would include the initial operation

cowboyd:
Ah, ok. I misunderstood then. There is no timeout on the operation as a whole, but just a cap on the delay to wait between operations. In that case, I think it can be expressed even more succinctly with a single loop:

```ts
import { Operation, sleep } from "effection";

export function* retryBackoffExample<T>(
  op: () => Operation<T>,
  { attempts = 5, startDelay = 5, maxDelay = 200 } = {},
): Operation<T> {
  
  let current = { attempt: 0, delay: startDelay, error: new Error() };

  while (current.attempt < attempts) {
    try {
      return yield* op();
    } catch (error) {
      if (!isKnownError(error)) {
        throw error;
      } else {
        yield* sleep(current.delay);
	current.attempt++;
        current.error = error;
        current.delay = Math.min(current.delay * 2, maxDelay);
      }
    }
  }
  throw current.error;
}
```

If we wanted to, we could easily imagine writing a `withTimeout()` to limit the time allowed for the entire operation. Decoupling total operation timeout from retry logic seems like the proper factoring.

cowboyd:
@taras 

What about this? shorter _and_ no tears.

```ts
import { call, type Operation, race, sleep } from "effection";

export function* retryBackoffExample<T>(
  op: () => Operation<T>,
  { maxAttempts = 5, startDelay = 5, maxDelay = 200 } = {},
): Operation<T> {
  return race([
    timeout(maxDelay),
    call(function* () {
      let attempt = 0;
      for (let delay = startDelay;; delay = delay * 2) {
        try {
          return yield* op();
        } catch (error) {
          if (!isKnownError(error) || attempt++ >= attempts) {
            throw error;
          }
          yield* sleep(delay);
        }
      }
    }),
  ]);
}

function* timeout(afterMillis: number): Operation<never> {
  yield* sleep(afterMillis);
  throw new Error("timeout");
}
```

taras:
We need to move this to contrib.

cowboyd:
Moved to https://github.com/thefrontside/effection-contrib/pull/30

PR #882, opened 2024-01-09
Minor doc fixes

minkimcello:
## Motivation

I was reading through the docs and noticed some typos and formatting inconsistencies.

## Approach

- Fixed typos
- Added missing import in an example
- Comments in shell start with `#`
- ~~Set OSS link to effection in the header to `/` and now displays `v3` instead of `v2`~~
- Capitalized H in GitHub for the header link
- There were two places where an operation was written as `function *foo` instead of `function* foo`. although it works either way, it was written as `function* foo` everywhere else so I fixed the two for consistency
- An example for streams was missing `channel.` for sending messages

### DIFF
--- a/www/components/header.tsx
+++ b/www/components/header.tsx
@@ -33,7 +33,7 @@ export function Header() {
                   <IconGithHub />
                 </span>
                 <span class="hidden md:inline-flex">
-                  Github
+                  GitHub
                 </span>
               </a>
             </li>
--- a/www/deno.lock
+++ b/www/deno.lock
@@ -103,8 +103,8 @@
         "integrity": "sha512-/kYRxGDLWzHOB7q+wtSUQlFrtcdUccpfy+X+9iMBpHK8QLLhx2wIPYuS5DYtR9Wa/YlZAbIovy7qVdB1Aq6Lyw==",
         "dependencies": {}
       },
-      "@types/hast@2.3.8": {
-        "integrity": "sha512-aMIqAlFd2wTIDZuvLbhUT+TGvMxrNC8ECUIVtH6xxy0sQLs3iu6NO8Kp/VT5je7i5ufnebXzdV1dNDMnvaH6IQ==",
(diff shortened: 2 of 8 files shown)


www/components/project-select.tsx
```diff
@@ -81,8 +81,8 @@ const projects = [
   {
     title: "Effection",
     description: "Structured Concurrency for JavaScript",
-    url: "/V2/",
-    version: "v2",
+    url: "/",
+    version: "v3",
```
cowboyd: This was so that you could switch from v3 to v2. There's no need to switch back to v3 if you're on v3 is there?

Is it confusing? If so, maybe we take it out of the switcher altogether?

cowboyd (commented):
Everything looks good, although I do have one question about the project selector?

www/components/project-select.tsx
minkimcello: Ohh I see. It does look like it was forgotten to be updated. On interactors and bigtest, they also have links that routes back to its own page so I don't think it's that weird to have effection v3 in the oss dropdown.

If there's a user that lands on v3 docs and is looking for v2, I'm not sure if their instinct would be to navigate to the OSS dropdown to find the v2 docs. For users to be able to find V2, we could introduce a versions dropdown:

Screenshot 2024-01-09 at 10 17 10 AM

But for now I'll just revert it back to `V2`.

cowboyd (approved):

minkimcello:
Someone's going to have to merge this for me.

cowboyd:
@minkimcello Can you set up commit signing?

minkimcello:
> @minkimcello Can you set up commit signing?

🫡

cowboyd:
@minkimcello I went ahead and rebased it for you.

minkimcello:
> @minkimcello I went ahead and rebased it for you.

Thank you. I don't have write access so if the PR looks good you can go ahead and merge it for me.

PR #884, opened 2024-01-10
Bring race() types closer to Promise.race()

bdougherty:
## Motivation

The existing types for `race()` required that all Operations had the same return type. This should not be necessary.

## Approach

Moves the `Yielded` type that was previously added for `all()` to the global types file and updates the `race()` return type to use it. That way it will return `Operation<union type of all the individual operations>`.

### DIFF
--- a/lib/all.ts
+++ b/lib/all.ts
@@ -1,4 +1,4 @@
-import type { Operation, Task } from "./types.ts";
+import type { Operation, Task, Yielded } from "./types.ts";
 import { spawn } from "./instructions.ts";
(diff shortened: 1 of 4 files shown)


cowboyd (approved):
Looks great! Awesome turn around, and fun that we have a `Yielded<T>` type now. I moved it up in the file where we have the "public", documented types and added a link to the `Awaited` builtin type in typescript.

cowboyd:
Thanks @bdougherty, we'll make sure to get this out with 3.0.2!

Issue #890, opened 2024-02-08
RFC: Embed API documentation into website


taras:
# Motivation

Deno website is hosting our API docs. Although convenient, it might send the wrong message to those unfamiliar with Deno.   

It would be nice to have API docs hosted on our website.

# Approach

We could use the same mechanism we use for v2 docs to serve API docs for different versions. This requires having the generated HTML somewhere we can download it. We could run a workflow on any `effection-v*` tag that generates HTML using Deno CLI, commit it, and push it to the tag.

This would allow us to read the docs for each version from the tag's tarball.

jbolda:
This is complete for both v3 and v4.

PR #891, opened 2024-02-09
Workflows for creating releases with docs

taras:
Related #890

## Motivation

This PR introduces two GitHub Actions Workflows:
* create-release.yaml
* rebuild-releases.yaml

### create-release.yaml

This workflow runs anytime a tag is created that matches "effection-v*". It'll checkout the tag, generate docs with `deno doc`, create a tarball and create a release.

### rebuild-releases.yaml

This workflow can only be triggered manually. It can be used to rebuild all of the releases. It has two jobs: _setup_ & _rebuild_. The setup job gets a list of all tags that match the "effection-v*" pattern and sticks the result into outputs. The _rebuild_ job uses outputes from _setup_ as matrix. The actual workflow is same as **create-release.yaml**

## TIL

1. GitHub has a new mechanism for managing tokens generated by users in an organization. If you want to interact with an organization's resources, you'll need to create a token specifically for that organisation. There are additional configuration options under TheFrontside github org settings
2. You can have [dynamic matrixes in GitHub Actions](https://thekevinwang.com/2021/09/19/github-actions-dynamic-matrix) at the bottom of the blog post there are good examples like [this one](https://github.com/suzuki-shunsuke/example-github-actions-dynamic-matrix/blob/master/.github/workflows/test.yml)
3. `act -s GITHUB_TOKEN="<YOUR_TOKEN>" -W .github/workflows/rebuild-releases.yml workflow_dispatch` can be used to run `rebuild-releases.yaml` on your machine

## TODO

* [x] How do I test this?
* [x] How do I run this for all existing tags?

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -36,12 +36,12 @@ jobs:
         env:
           VERSION: ${{steps.vars.outputs.version}}
           
-      - run: tar cfv docs.tar -C docs .
+      - run: tar cfvz api-docs.tgz -C docs .
 
       - name: Create Release
         uses: softprops/action-gh-release@v1
         with:
-          files: docs.tar
+          files: api-docs.tgz
           name: ${{steps.vars.outputs.version}}
 
       
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -31,8 +31,6 @@ jobs:
         id: vars
         run: echo ::set-output name=version::$(echo ${{github.ref_name}} | sed 's/^effection-v//')
 
-      - run: sed -i '/docs/d' .gitignore
-
       - name: Generate Docs
         run: deno doc --html --name=effection@$VERSION mod.ts
         env:
@@ -44,6 +42,6 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           files: docs.tar
-          token: ${{ github.token }}
+          name: ${{steps.vars.outputs.version}}
 
       
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -21,7 +21,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{github.ref}}
-          fetch-depth: 0 
 
       - name: setup deno
         uses: denoland/setup-deno@v1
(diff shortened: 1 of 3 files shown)


cowboyd (commented):
I'm not a huge fan of having workflows commit directly to the main branch. Why not just create a release associated with the tag that contains the docs and make the docs part of the release?

If we want to go the route of having our workflows commit to the repository, then I think it should be part of a "merge release" which pushes to a release branch, and then contains

- changelogs
- docs

Then _that_ thing gets tagged. 

However, if our website needs to be able to switch between API docs versions, then it doesn't make sense to have it in the repository.

taras:
Using a release is a good idea and it might be easier. I'll do that

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: deno doc --html --name=effection@$VERSION mod.ts
+        env:
+          VERSION: ${{steps.vars.outputs.version}}
+          
+      - run: tar cfv docs.tar -C docs .
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: docs.tar
```
cowboyd: Two suggestions: 

1. let's call them api-docs in case we decide to release other kinds of togs
2. let's compress them so that they can be quickly transferred across the network

```suggestion
      - run: tar cfvz api-docs.tgz -C docs .

      - name: Create Release
        uses: softprops/action-gh-release@v1
        with:
          files: api-docs.tgz
```

cowboyd (commented):
Question: Do we need the rebuild releases? At the most, it seems like we would need them for `3.0.0`, `3.0.1`, and `3.0.2`

taras:
> Do we need the rebuild releases? At the most, it seems like we would need them for 3.0.0, 3.0.1, and 3.0.2

I guess that's true. I was thinking in case we want to update the releases to a specific format. Like now, I need to rerun, to rebuild at least 3.0.0, 3.0.1 and 3.0.2. In the future, there will be more. I can change the filter to only include non-preview releases. 

What do you think?

cowboyd:
> > Do we need the rebuild releases? At the most, it seems like we would need them for 3.0.0, 3.0.1, and 3.0.2
> 
> I guess that's true. I was thinking in case we want to update the releases to a specific format. Like now, I need to rerun, to rebuild at least 3.0.0, 3.0.1 and 3.0.2. In the future, there will be more. I can change the filter to only include non-preview releases.
> 
> What do you think?

I'm torn, on the one hand, I feel like releases are releases, and that means a binary artifact corresponding to a specific git SHA. On the other, you're right. What if we need to update urls, or change style?

In that case, maybe it's best to have an executable task in the repo that we can do manually?

taras:
> In that case, maybe it's best to have an executable task in the repo that we can do manually?

The setup I have allows us to rerun the job via GitHub UI. Would you like me to rewrite it to Deno?

PR #892, opened 2024-02-13
Caching docs in Deno.Kv and testable file system helpers

taras:
## Motivation

I started rewriting the GitHub Actions workflows from #891 into Deno tasks. In the process, I realized that I would like to cache the contents of the docs directory so I can load it from Deno.Kv instead of untarballing the files in the Deno instance. As it was getting more complicated, I wanted to start building up helpers and ensure they were tested. I kept ballooning the requirements, but it's good practice.

## Approach

1. Created `useCwd` hook that uses `CwdContext` for storing the current context.
2. Created `useReadDir` and `useMkdir` that are `useCwd` compatible
3. Started writing test

## TODO

1. [ ] My `useCwd` doesn't actually work, it's something about the way I'm using scope. `useReadDir` and `useMkdir` don't actually get the set value.


cowboyd:
I think we're going to go the staticalize route and just snapshot everything to disk and call it a day.

Issue #893, opened 2024-02-15
imports not used when using multi-root workspaces for website development in VSCode


taras:
[VSCode Deno plugin does not read `deno.json`](https://github.com/denoland/vscode_deno/issues/938) when used in multi-root workspace configuration. `www/deno.json` has imports path for Effection in other dependencies which Deno LSP does not know how to resolve which causes the following error

<img width="878" alt="image" src="https://github.com/thefrontside/effection/assets/74687/6560fa81-7b3e-48cc-bb04-c7763aa80e38">

cowboyd:
@taras Does [the workaround](https://github.com/denoland/vscode_deno/issues/938#issuecomment-1734106460) in that issue work?

taras:
No, because it's not the same issue. Our issue isn't that `deno.json` is in a folder. The problem is that it only reads the `deno.json` of the root workspace. The workaround in #894 is to make `www` root, which automatically reads the `deno.json` with imports relying on the fact that the core doesn't have any imports.

jbolda:
The docs site has been moved into a separate repo (and also I think deno now supports this, but either way) so this can be closed,

PR #894, opened 2024-02-15
Make it easier to contribute to www with VSCode

taras:
Related #893 

## Motivation

We want it to be easy for people to contribute to the website using VSCode. 

## Approach

Due to https://github.com/denoland/vscode_deno/issues/787, imports are only read from the workspace's default _deno.json_ file. Fortunately, we can trick VSCode Deno Plugin into reading the `deno.json` file from 'www' directory by creating a `.code-workspace` file that has `www` directory as the first in the list of folders. Since the core doesn't have dependencies, this issue will not effect the core folders.

1. created `www/www.code-workspaces` file with `www` as first and main folder
2. Added `www/.vscode/settings.json` with `deno.enable: true` to activate deno when it's root
3. Added `www` directory to excludes `.vscode/settings.json` 
4. Added instructions to README.md that encourage users to read `www/README.md` for website contribution instructions
5.  Added instructions on how to use the `www/www.code-workspaces` file for contributing to the website.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-  "deno.enable": true
+  "deno.enable": true,
+  "files.exclude": {
+    "www": true,
+  }
 }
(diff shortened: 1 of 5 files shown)


cowboyd (approved):
:shipit:

PR #905, opened 2024-08-30
Added context documentation

taras:
## Motivation

Context is a killer feature of Effection that we don't have documented on our website.

## Approach

* I "borrowed" way too much inspiration from [React Context](https://react.dev/learn/passing-data-deeply-with-context) docs.
* Added a section about naming conventions that someone asked me about in Discord.


taras:
What's up with this? 
image

taras:
@cowboyd why is Lint failing here? https://github.com/thefrontside/effection/actions/runs/10638858208/job/29495674024?pr=905


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection">Added context documentation · thefrontside/effection@5a86e90</a></strong></div><div>Structured concurrency and effects for JavaScript. Contribute to thefrontside/effection development by creating an account on GitHub.</div></blockquote>

taras:
@cowboyd I fixed the TOC sidebar but lint is not passing.

cowboyd:
@taras It looks like maybe line wrap is disabled and so it's wrapping to the same line? not sure.

As for the content, this is really good, and I like cribbing from React. I do think that we should mention these pratical pieces of information:

1. Context serves the same purpose as "dependency injection" except that it is coupled to lexical scope and not 
2. You can use a resource in combination with a context to initialize a service and "inject it", which is what folks use.
3. Mabye some discussion around the difference between `yield* Context` vs `yield* Context.get()`

taras:
Let's merge what we have and add to it

cowboyd:
> Let's merge what we have and add to it

The thing is that I don't think the advice to wrap in a `use` helper makes much sense unless it's understood that there are multiple ways to access the context each with different behaviors and that it is generally a good idea to abstract around the access depending on your use-case. By the same token, an `initContext()` operation doesn't make any sense outside the discussion of services and resources. Why not just call `Context.set()`? It's because there may be long running things that need to happen as part of setting it up. What might those things be?

I'd be happy to add this as this makes it really easy to slot in.

taras:
@cowboyd I removed the naming convention section. DI should be a separate page in the advanced section.

cowboyd (approved):

PR #906, opened 2024-08-31
Fix lint errors using deno lint --fix

taras:
## Motivation

#905 is blocked because of lint errors, so I will fix them here and then rebase.

## Approach

1. Ran `deno lint --fix`
2. Using `npm:process` instead of `globalThis.process`
3. Pinned to 1.46.x in CI

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -20,12 +20,12 @@ jobs:
       - name: setup deno
         uses: denoland/setup-deno@v1
         with:
-          deno-version: v1.x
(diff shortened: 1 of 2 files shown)


cowboyd:
Might have been caused by the Deno default linting rules changing in between point releases of Deno. We may want to pin our Dino version in the set up Dino action so that this doesn't happen in the future.

taras:
@cowboyd ~I'll do this in a separate PR~. I'll do it here.

cowboyd (approved):

PR #914, opened 2024-10-25
Race was failing because result of holt is a promise, but `yield*` wants an operation

taras:
## Motivation

image

## Approach

Wrap each shutdown promise in call


cowboyd:
This would be a good fix, but we actually need to make a `Task` an operation, not just a promise for two reasons. 1) we want working with tasks to be fluent inside other operations and forcing a call() around them would get really tedious, really quick. 2) and this is really important, we need operations to resolve synchronously if possible, and always consuming a task as a promise will force it to delay to the until the next tick of the run loop. This could be detrimental for task supervision.

That's why we needed to go with #915

PR #916, opened 2024-11-05
Moved useScope to own file to eliminate circular dependency

taras:
## Motivation

I was encountering the following error.

image

A circular dependency caused it

1. scope.ts imports Children, Generation from contexts.ts
2. contexts.ts imports createContext from context.ts
3. context.ts import useScope from scope.ts

## Approach

Moved `useScope` to own file and import from that file.

The result is

1. scope.ts imports Children, Generation from contexts.ts
2. contexts.ts imports createContext from context.ts
3. context.ts import useScope from use-scope.ts


cowboyd:
context was going to be refactored to not have `useScope()` anyway, so I went ahead brought that change forward with #918  which fixes this issue.

PR #917, opened 2024-11-05
⬆️ Upgrate to Deno 2

cowboyd:
## Motivation
Deno 2 is out, and we're not going back.

## Approach
Deno 2 formats yaml files, and it uses a stricter mode of TypeScript which treat the argument to a catch clause as `unknown` rather than `any`. In general, this is a good thing, but we have been treating errors as errors thus far with little bad effect. This adds an explicit cast to each catch clause to assume that it is an `Error` like before.

It highlights the danger which is good, and we should seriously consider moving away from this assumption (although a pox upon the house of anyone who throws anything but an error). Nevertheless, now is not the time.

### DIFF
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -20,14 +20,13 @@ jobs:
         uses: actions/checkout@v3
 
       - name: setup deno
-        uses: denoland/setup-deno@v1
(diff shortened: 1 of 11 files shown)


taras (approved):
👍

PR #918, opened 2024-11-05
✨Introduce Do() operation

cowboyd:
## Motivation

Many of the Operations are comprised of a single Effect, and very often a synchronous Effect at that. There is no need to pay any of the overhead for creating a generator when we could just make a simple iterator.

## Approach

This introduces a `Do()` function which takes any effect and converts it automatically into an operation (an effect iterator) so that it can be used anywhere that an operation can be used. The iterator just returns a wrapping effect that captures its result, and then the next time it is called, it returns that result. This gives us better performance and also solves the circular dependency problem inside `context.ts`

### DIFF
--- a/lib/context.ts
+++ b/lib/context.ts
@@ -1,25 +1,17 @@
 import { Context, Effect, Operation, Scope } from "./types.ts";
 import { Ok } from "./result.ts";
-import { useScope } from "./scope.ts";
+import { Do } from "./do.ts";
 
 export function createContext<T>(name: string, defaultValue?: T): Context<T> {
   let context: Context<T> = {
     name,
     defaultValue,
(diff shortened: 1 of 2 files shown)


taras (approved):
How do you know that it improves performance?

cowboyd:
> How do you know that it improves performance?

I'm using this performance suite created by @rixtox https://github.com/cowboyd/async-benchmark/tree/effection-v4



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/cowboyd/async-benchmark">GitHub - cowboyd/async-benchmark at effection-v4</a></strong></div><div>Contribute to cowboyd/async-benchmark development by creating an account on GitHub.</div></blockquote>

PR #930, opened 2024-12-14
✨introduce v4 action API

cowboyd:
## Motivation
The new action API is changing in order to be more in line with the new Promise() constructor, so we want to communicate that change.

The new api is:

```ts
action((resolve, reject) => {
  let timeout = setTimeout(() => {}, 1000);
  return () => { clearTimeout(timeout) };
})
```
As you can see, it is just like `new Promise()` except the return value is the "uninstall" function. This brings us closer to the Rosetta Stone promise that `action()` is just like promise constructor.

## Approach

This adds the new action API to the old action API in a backwards compatible manner via an overload, and then deprecates the old API.


taras (approved):
👍

PR #931, opened 2024-12-14
Added effection-contrib section

taras:
## Motivation

Contribs is going to be awesome! Let's add it to the site.

## Approach

* Took `proxyRoute` from frontside.com, but it uses an older version of effection and revolution so I just did some copy/pasta

### DIFF
--- a/www/routes/proxy-route.ts
+++ b/www/routes/proxy-route.ts
@@ -1,0 +1,58 @@
+import type { HTTPMiddleware } from "revolution";
+import { call, Operation } from "effection";
+
+export interface ProxyRouteOptions {
+  website: string;
+  prefix: string;
+  root?: string;
+}
+
+export function proxyRoute(options: ProxyRouteOptions): HTTPMiddleware {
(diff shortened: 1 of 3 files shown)


www/routes/proxy-route.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    target.hostname = website.hostname;
+    target.port = website.port;
+    target.protocol = website.protocol;
+
+    let base = new URL(`/${options.prefix}`, request.url);
+
+    let headers: Record<string, string> = {
+      "X-Base-Url": base.toString(),
+    };
```
cowboyd, line 27: Are we using this base header in the contrib website?

I think we may want to start statically generating the effection website at this point. because it's happening in the context of us also needing to generate the frontside website as well. Let's pair up to get that done

cowboyd (commented):
Can we call it contrib to be consistent with the library name? Even though contrib is singular it has the connotation of encompassing all contributions.

PR #932, opened 2024-12-16
🔥 Remove v2 docs.

cowboyd:
## Motivation

Maintaining an ancient Docusaurus site that is doubly nested is untenable.

## Approach

We'll take down the v2 docs site. However, T hey are still available to download in TGZ format from previous github release assets


taras (approved):
We're probably the only people who'll need v2 docs. We know where we can find them.

PR #933, opened 2024-12-17
✨Add `scoped()` higher order operation

cowboyd:
## Motivation

One of the many overlapping concerns of `action()` and `call()` were to establish context boundaries, error boundaries, and concurency boundaries, but this was problematic because it required the creation of a new frame and thus a new iterator every time you wanted to do one of those things, so stack traces were lost. It also was problematic because sometimes you would use `call()` for an aysnc function, other times, you wanted to use it to encapsulate an operation, so when reading code you never knew why exactly `call()`/`action()` etc. were being used.

However v4 has the ability to separate all these concerns: error handling, concurrency delimiting, and context delimiting, but they are all separate functions and a lot to teach. Instead, we want a single function that we can use for encapsulating all effects.

## Approach

This introduces the `scoped()` operation which takes any operation, traps errors that may emit from it, and wraps a new internal scope around it to prevent any effects from leaking out of it.

### DIFF
--- a/lib/mod.ts
+++ b/lib/mod.ts
@@ -22,3 +22,4 @@ export * from "./abort-signal.ts";
 export * from "./main.ts";
 export * from "./with-resolvers.ts";
 export * from "./async.ts";
+export * from "./scoped.ts";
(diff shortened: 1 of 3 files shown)


neurosnap (approved):
nice!

PR #938, opened 2024-12-20
🔥 Remove v2 docs.

taras:
They are still available to download in TGZ format from previous GitHub release assets

Related #932


cowboyd:
Nice catch!

PR #939, opened 2024-12-20
Publish to JSR

taras:
## Motivation

I want to update revolution and such to e4, but I'd like to consume it from JSR. 

## Approach

1. Since we're not tracking version in deno.json, so I created a script that adds the version and other needed properties to deno.json before running `npx jsr publish`.
2. I changed `https://` imports to jsr since JSR doesn't allow these.
3. I added explicit types to eliminate "slow types"
4. I had to add `--allow-dirty` to `jsr publish` because deno.json becomes dirty after changes.
5. Upgraded dnt

### DIFF
--- a/tasks/build-jsr.ts
+++ b/tasks/build-jsr.ts
@@ -1,0 +1,9 @@
+await Deno.writeTextFile(
+  "jsr.json",
+  JSON.stringify({
+    name: "@effection/effection",
+    version: Deno.env.get('VERSION'),
+    exports: "./mod.ts",
+    include: ["lib", "mod.ts", "README.md"],
+  }),
+);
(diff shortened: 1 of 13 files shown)


tasks/build-jsr.ts
```diff
@@ -0,0 +1,15 @@
+import jsonDeno from "../deno.json" with { type: "json" };
+
+await Deno.writeTextFile(
+  new URL("../deno.json", import.meta.url),
+  JSON.stringify({
+    ...jsonDeno,
+    name: "@effection/effection",
+    version: Deno.env.get("VERSION"),
```
cowboyd, line 7: We can include everything in the source `deno.json` except for the version since it will never change, right?

taras: Done.

cowboyd (approved):
🎉

PR #943, opened 2024-12-21
Use JSR Token instead of OIDC

taras:
## Motivation

OIDC is the recommended way of authenticating from GitHub Actions to JSR, but it requires that the person who triggered the workflow (i.e., the person who merged the code that triggered the workflow) be a member of the package scope in JSR. We want FrontsideJack to be the only one with scope membership, so we can't use OIDC.

## Approach

Created a token for FrontsideJack in JSR that's scoped to only publish to `effection` package, added the token as a secret to the effection repository and passing the token via `jsr publish --token`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -61,4 +61,4 @@ jobs:
         run: deno task build:jsr ${{steps.vars.outputs.version}}
 
       - name: Publish JSR
-        run: npx jsr publish --allow-dirty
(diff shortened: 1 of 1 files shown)


cowboyd (approved):
I would add, that it is a feature that the maintainers list of the affection repository does not need to be perfectly in sync with the org admin of JSR for the release process to work.

As it stands if we we have a maintainer of the GitHub repository who is not also an admin on the JSR.io our release process won't work guaranteed.

PR #947, opened 2024-12-21
Publish v4 to JSR

taras:
## Motivation

I'm going to start migrating dependent projects to effection@4.0.0-alpha.3, so I'd like to have v4 available on JSr.

## Approach

Cherry-picked commits from v3


cowboyd (approved):
Nice! I do wonder why isn't that our linting rules can't catch all of the extra constraints required by JSR.  Is there a setting we can apply?

taras:
These are JSR-specific rules mostly to avoid slow types.

PR #949, opened 2024-12-24
Add Contrib link to the header

taras:
## Motivation

I forgot to include a link to the contrib page in #948 

## Approach

Add missing link to the header


cowboyd (approved):
Awesome! Also, we need to make sure and add the contrib sitemap

taras:
Do we need to do something special for the sitemap? I thought it was automatic

cowboyd:
I don't think we updated the proxy site map.

cowboyd:
@taras is this still necessary?

taras:
Nope

Issue #946, opened 2024-12-21
Add equivalent to shared `Promise` to FAQ


SebastienGllmt:
In typescript, a very common pattern is to share a promise in multiple places. For example:

```typescript
class Foo {
  data: Promise<number>;

  constructor() {
    this.data = (async () => 5)();
  }

  async getData() {
    return await this.data;
  }
}

const foo = new Foo();
console.log(await foo.getData());
```

However, this can't be done directly with Operation as they can only be computed once:

```typescript
const func = function* () {
  console.log("compute");
  return 5;
}

const started = func();

console.log(await run(() => started)); // returns 5
console.log(await run(() => started)); // returns undefined
```

# Solution

A solution to this is to use the same `run` (aka the same `Task`) multiple times instead of creating a new task every time
```ts
const func = function* () {
  console.log("compute");
  return 5;
}

const started = func();
const started2 = run(() => started) // save the `Task` to reuse it

console.log(await started2); // returns 5
console.log(await started2); // returns 5 again!

main(function* () {
  // and this can even be used as a generator!
  console.log(yield* started2); // still 5
  console.log(yield* started2); // still 5
});
```

That is to say, I think `Task` is the go-to when you want this kind of Promise-like behavior. The only caveat to this approach is that it doesn't work when you try to chain the result of the task as described in solution (3) in #944

# Other options

I noticed that `effect-ts` handles this differently by having a `fork` and `join` command: https://effect.website/docs/additional-resources/effect-vs-promise/#faq

cowboyd:
This might be a way that does not require creating a task, but uses `withResolves()` to store result of a computation. It can convert any operation into a shared operation:

```ts
function shared<TArgs extends unknown[], T>(
  compute: (...args: TArgs) => Operation<T>,
): (...args: TArgs) => Operation<T> {
  return (...args) => {
    let result: WithResolvers<T> | undefined;
    return {
      *[Symbol.iterator]() {
        if (result) {
          return yield* result.operation;
        } else {
          result = withResolvers<T>();
          try {
            let value = yield* compute(...args);
            result.resolve(value);
            return value;
          } catch (error) {
            result.reject(error as Error);
            throw error;
          }
        }
      },
    };
  };
}
```

E.g.

```ts
const func = shared(function* () {
  console.log("compute");
  return 5;
});

await run(func) //=> 5, console.log("compute");
await run(func) //=>  5
```

Issue #944, opened 2024-12-21
Add equivalent to `Promise.then` to FAQ


SebastienGllmt:
In JS, it's common in a function that isn't `async` to chain promises together using `.then` and `.catch`

However, it's not clear what the effection equivalent to this is (what happens if you're in a function that isn't a generator? Do you just convert everything to promises using `run(myGenerator).then((...) => { ... })` and escape effection temporarily)?

There were two suggestions by Charles on Discord for this:

1. A simple approach that does not chain together well
```ts
function* then<A,B>(operation, fn: (value: A) => Operation<B>): Operation<B> {
  return yield* fn(yield* operation);
}
```

2. An approach that chains together well as long as you have a `pipe` function from somewhere (lodash, etc. or write your own)
```ts
function then<A,B>(fn: (value: A) => Operation<B>): (operation: Operation<A>) => Operation<B> {
  return function*(operation) {
    return yield* fn(yield* operation);
  }
}

// usage
pipe(
  operation,
  then(lift((val) => val * 2)),
  then(lift((val) => String(val)),
  then(lift((val) => val.toUpperCase())),
)
```

_Differences with `Promise.then`_: typically you're allowed to chain non-async steps (ex: `(async () => 5)().then(a => a+1)`). Internally (assuming this is a Promise and not a promise-like), this can be implemented by checking if the return type of `fn` is `instanceof Promise`. One can do maybe achieve something similar by checking if `fn` has a generator symbol, but another approach is to just require using `lift` instead

3. Another option is to leverage `Task`

Note that effection already defines a `Task` interface that is _both_ a Promise and an Operation. In this sense, we already have a path for implementing this pattern via `run()` which returns a Task. The problem though is that the implementation of `then`/`catch`/`finally` on the result of `run()` do NOT return a Task, but rather return just a `Promise` which means you can't chain it together.

If we instead had an implementation of `Task` that allowed chaining, this would also solve the problem nicely. Note that `effect-ts`, for example, 

## Other projects

Note that the `effect-ts` library has a similar design decision. They decided that their equivalent to tasks can only be piped, and that once you convert to a promise-like API, you cannot go back

```ts
import { Effect } from "effect"

const task1 = Effect
    .succeed(1)  // create a task
    .pipe(Effect.delay("200 millis"));  // okay to combine with pipes

Effect
    .runPromise(task1) // convert to promise-like API
    .then(console.log);
```

[...]

PR #950, opened 2024-12-29
Allow yielding a chain of `then` on `Task`

SebastienGllmt:
## Motivation

There are cases where you need to do the following steps:
1. Create a `Task` (using `run`)
2. Do some computation on the `Task` using `then`
3. `yield` the result of the task

example that shows how this can be used to:
- implement #946 without requiring to use `async/await` explicitly anywhere
- easily chain things without having to awkwardly wrap results with `call` as in #944  
```ts
class Foo {
    task: Future<number>;
    
    constructor() {
       this.task = run(() => ...).then(val => val+1);
    }

    *doWork() {
        yield* this.task;
    }
}
```

This was a bit tedious to do previously (see #944), and this PR simplifies this case a bit by making that calling `then` on a `Task` returns a `Future` (which you can yield) instead of just a `Promise`

## Approach & Alternate Designs

You can find some background in #944  and inline comments in this PR

### TODOs and Open Questions

- [ ] Do we want to make creating `Future`s more easier in general? This has a chance of overlapping conceptually with the `pipe` concept that was recently removed, and also overlaps with the idea of `call` (although `call` returns an `Operation` instead of a `Future`)
- [ ] Is there a way to make that `then` returns a `Task` instead of a `Future`? I tried this initially, but ran into some complications: https://github.com/thefrontside/effection/issues/944#issuecomment-2564596508

### DIFF
--- a/lib/run/task.ts
+++ b/lib/run/task.ts
@@ -115,7 +115,6 @@ export function createTask<T>(
             });
           }
         },
-        // then: (...args) => getHaltPromise().then(...args),
         then: async <Result1=void, Result2=never>(onfulfilled?: ThenFulfilledType<void, Result1>, onrejected?: CatchType<Result2>) => {
           type NewResult = Result1 | Result2;
           const newPromise = getHaltPromise().then(onfulfilled, onrejected);
--- a/lib/run/task.ts
+++ b/lib/run/task.ts
@@ -6,6 +6,10 @@ import { action } from "../instructions.ts";
 
 import type { FrameResult } from "./types.ts";
 import { create } from "./create.ts";
+import { call } from "../call.ts";
+
+type ThenFulfilledType<T, Then=T> = ((value: T) => Then | Promise<Then>) | undefined | null;
+type CatchType<Catch=never> = ((reason: any) => Catch | Promise<Catch>) | undefined | null;
 
 export function createTask<T>(
   frame: Frame<T>,
@@ -32,7 +36,7 @@ export function createTask<T>(
     return promise;
   };
 
-  let task = create<Task<T>>("Task", {}, {
+  let task: Task<T> = create<Task<T>>("Task", {}, {
     *[Symbol.iterator]() {
       let frameResult = evaluate<FrameResult<T> | void>(() => frame);
       if (frameResult) {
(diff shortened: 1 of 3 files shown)


lib/run/task.ts
```diff
@@ -86,10 +115,37 @@ export function createTask<T>(
...
+          const newPromise = getHaltPromise().then(onfulfilled, onrejected);
+          const future: Future<NewResult> = create<Future<NewResult>>("Future", {}, {
+            *[Symbol.iterator]() {
+              return yield* call(() => newPromise);
+            },
+            then: (...args) => newPromise.then(...args),
+            catch: (...args) => newPromise.catch(...args),
+            finally: (...args) => newPromise.finally(...args),
+          });
+          return await future;
```
SebastienGllmt, line 129: note: the code here for handling the abort future is mostly the same as the `then` in the parent `Task`

This is why I think maybe it's better if we have a more general way to create a `Future` that is exposed somewhere that is used in both these places

There is, to some extent, a conceptual overlap between what this block of code is trying to do and what `call` is doing. That is to say, if `call` returned a `Future` instead of an `Operation`, using it here could also be an option

lib/types.ts
```diff
@@ -53,7 +53,30 @@ export interface Operation<T> {
...
+   */
+  catch<TResult = never>(onrejected?: ((reason: any) => TResult | PromiseLike<TResult>) | undefined | null): Future<T | TResult>;
+
+  /**
+   * Attaches a callback that is invoked when the Promise is settled (fulfilled or rejected). The
+   * resolved value cannot be modified from the callback.
+   * @param onfinally The callback to execute when the Promise is settled (fulfilled or rejected).
+   * @returns A Promise for the completion of the callback.
+   */
+  finally(onfinally?: (() => void) | undefined | null): Future<T>;
```
SebastienGllmt, line 78: Note: these three are copy-pasted from the definition of `Promise`, but have the return type changed from `Promise` → `Future`

lib/run/task.ts
```diff
@@ -86,10 +115,37 @@ export function createTask<T>(
             });
           }
         },
-        then: (...args) => getHaltPromise().then(...args),
```
SebastienGllmt, line 89: Note: there is one very subtle change introduced by this PR: the result of these is no longer a `Promise` and, instead, is now a `PromiseLike`. In practice, this *should* work everywhere (the goal of PromiseLike is that they're meant to be supported by everything magically), but it _is_ possible to write code that only works on Promises and not PromiseLike (ex: any code that uses `instanceof Promise`)

cowboyd:
I'm wondering about the corner cases here.

1. Will the returned promises be lazy? I.e. 
2. What if the return value is an operation vs a promise vs a regular value?

I.e. what would be the expectation around this:

```ts
run(fn).then(function*(value) { return value + 1 })
```

SebastienGllmt:
1. The laziness is not changed by this PR. Nothing happens (lazy) until you use `await` just like before this PR
2. Currently, the behavior is the same as promises in generally. That is to say,
    1. if you return a value, it stays a value
    2. if you return a promise, it gets flattened to a value (`.then(() => Promise.resolve(3))` becomes just `3`)
    3. If you return a generator, it doesn't get flattened (it gets treated as case 1 and just returns a generator)

You could say that may we would like `then` to flatten generators as well for usability reasons, but this is not possible because the type definition of `then` comes from `Promise` (we cannot just arbitrarily change it). You could introduce a new function in the `Future` interface that flattens *both* promises in generators if we want

cowboyd:
This should be addressed by https://github.com/thefrontside/effection/pull/1065 and https://github.com/thefrontside/effectionx/pull/112

Please let us know if there is anything we missed.

PR #951, opened 2025-01-02
📝 Document the new action interface in v4

cowboyd:
## Motivation

The `action()` interface has been greatly simplifield for v4 bringing it almost completely in line with the `new Promise()` constructor.

## Approach

This adds docs for both the function and also the rosetta stone entry.

### DIFF
--- a/lib/action.ts
+++ b/lib/action.ts
@@ -1,11 +1,50 @@
 import { Err, Ok } from "./result.ts";
 import type { Effect, Operation } from "./types.ts";
 
-interface Resolver<T> {
+/**
+ * A function used to define how an {@link action} will be resolved. This
+ * function is passed two arguments: a `resolve` callback used to resolve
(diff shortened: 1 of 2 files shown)


lib/action.ts
```diff
@@ -1,11 +1,50 @@
...
+ * A function used to define how an {@link action} will be resolved. This
+ * function is passed two arguments: a `resolve` callback used to resolve
+ * the action with a value and a `reject` callback used to reject the
+ * action with a provided error.
+ *
+ * @param executor - the function to run whenever an action is evaluated.
+ * @return a "finally" function that is _always_ run whether the
+ *           action is resolved, rejected, or discarded.
+ */
+interface Executor<T> {
```
cowboyd, line 14: This was renamed to `Executor` because  that's the terminology used by the `Promise` constructor

taras (approved):
👍 very nice!

PR #954, opened 2025-01-03
📝 docs for v4 call()

cowboyd:
## Motivation

We changed and simplified `call()` to be more aligned with `Function.prototype.call()`, and we need to make sure that the docs reflect this.

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -32,8 +32,9 @@ export interface Callable<
 }
 
 /**
- * Pause the current operation, async function, plain function, or operation function.
- * The calling operation will be resumed (or errored) once call is completed.
+ * Pause the current operation and evaluate an async function, plain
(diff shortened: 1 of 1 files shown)


lib/call.ts
```diff
@@ -56,10 +57,14 @@ export interface Callable<
...
  * yield* call(() => "a string");
  * ```
- * @param callable the operation, promise, async function, generator funnction,
+ *
+ * The function will be invoked anew every time that the `call()` operation is evaluated.
+ *
+ * @param callable - the operation, promise, async function, generator funnction,
  * or plain function to call as part of this operation
+ *
+ * @returns an {link @Operation} that evaluates to the result of executing the function to completion
```
taras: The link formatting is wrong here, it should be `{@link Operation}`

taras (approved):

PR #957, opened 2025-01-05
📝 update the `suspend()` docs to account for v4 API

cowboyd:
## Motivation

The suspend operation is no longer fundamental. It is just an action that does nothing, so we had to update the docs to reflect that.

## Approach
We had to change not only the API docs, but I also noticed that there were several places where the old `action+suspend` trick was being used that are no longer valid.

### DIFF
--- a/www/docs/actions.mdx
+++ b/www/docs/actions.mdx
@@ -1,51 +1,24 @@
-In this section, we'll cover the first two fundamental operations in Effection:
-`suspend()` and `action()`, and how we can use them in tandem to serve as a safe
-alternative to the [Promise constructor][promise-constructor].
+In this section, we'll cover one of the most fundamental operations in
+Effection: `action()`, and how we can use it as a safe alternative to
+the [Promise constructor][promise-constructor].
 
-## Suspend
+## Example: Sleep
 
-Simple in concept, yet bearing enormous practical weight, the `suspend()`
-operation is fundamental to Effection. It pauses the current
-operation until it [passes out of scope][scope], at which point it will return
-immediately.
-
-Let's revisit our simplified sleep operation from the [introduction to
+Let's revisit our sleep operation from the [introduction to
 operations](./operations):
 
+
 ```js
-export function sleep(duration) {
-  return action(function* (resolve) {
+export function sleep(duration: number): Operation<void> {
+  return action((resolve) => {
     let timeoutId = setTimeout(resolve, duration);
-    try {
-      yield* suspend();
-    } finally {
-      clearTimeout(timeoutId);
-    }
+    return () => clearTimeout(timeoutId);
   });
 }
 ```
 
 As we saw, no matter how the sleep operation ends, it always executes the
-`finally {}` block on its way out; thereby clearing out the `setTimeout`
-callback.
-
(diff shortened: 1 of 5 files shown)


www/docs/actions.mdx
```diff
@@ -80,22 +53,13 @@ await Promise.all([sleep(10, signal), sleep(1000, signal)]);
...
-context of an action as the pivot between that action's setup and teardown.
-
-## Actions
+## Action Constructor
 
-The second fundamental operation, [`action()`][action], serves two
-purposes. The first is to adapt callback-based APIs and make them available as
-operations. In this regard, it is very much like the
-[promise constructor][promise-constructor]. To see this correspondance, let's
+The [`action()`][action] function provides a callback based API to create Effection operations. You don't need use it all that often, but when you do it functions almost exactly like the
```
davidbrochart:
```suggestion
The [`action()`][action] function provides a callback based API to create Effection operations. You don't need it all that often, but when you do it functions almost exactly like the 
```

www/docs/actions.mdx
```diff
@@ -147,119 +113,26 @@ await Promise.race([
...
     xhr.open("GET", url);
     xhr.onload = () => resolve(xhr.responseText);
     xhr.onerror = () => reject(xhr.statusText);
     xhr.send();
-    try {
-      yield* suspend();
-    } finally {
-      xhr.abort();
-    }
+    return () => { xhr.abort(); }; //called in all cases
```
davidbrochart:
```suggestion
    return () => { xhr.abort(); }; // called in all cases
```

www/docs/operations.mdx
```diff
@@ -205,48 +205,34 @@ await main(function*() {
...
-3. uninstall the timeout callback
-
-If we always follow those steps, then we'll always resume at the right point,
-and we'll never leak a timeout effect.
-
->💡As an alternative to using a `finally {}` block to express your cleanup, you
-> can also use the [`ensure()`][ensure] operation which in some cases can add
-> clarity. Which method you use is up to you!
+Like every operation in Effection, the logic to enter a sleep is
+bundled right alongside the logit to exit it. As a result, and we'll
```
davidbrochart:
```suggestion
bundled right alongside the logic to exit it. As a result, and we'll
```

www/docs/operations.mdx
```diff
@@ -205,48 +205,34 @@ await main(function*() {
...
+never leak a timeout effect.
 
 ## Computational Components
 
 It's amazing to think how so many operations in the Effection
-ecosystem can be broken down into a combination of the fundamental:
-`action()`, `resource()`, and `suspend()` operations.  Composability
-is the key to everything.
+	ecosystem can be broken down into a combination of the fundamental:
```
davidbrochart:
```suggestion
ecosystem can be broken down into a combination of the fundamental:
```

cowboyd:
Thanks as alway @davidbrochart! I really appreciate the time you have taken to proofread changes to our docs in this PR and so many others.

davidbrochart:
Thank you for merging, I hope I will contribute in a more significant way in the future :)

PR #960, opened 2025-01-07
📝 document error boundaries and the `scoped()` function

cowboyd:
## Motivation

The way error boundaries are created has changed from `v3` to `v4`. Instead of using `call()` or `action()` to establish an error boundary, there is now a dedicated `scoped()` function which not only traps any errors, but also creates a resource boundary that does not let any resource or task escape.

## Approach
This updates the error documentation, and also adds documentation to the `scoped()` function. Finally, there was an old deno link that I updated to JSR.

### DIFF
--- a/lib/scoped.ts
+++ b/lib/scoped.ts
@@ -9,6 +9,18 @@ import { useCoroutine } from "./coroutine.ts";
  * shut down, and all contexts will be restored to their values outside
  * of the scope.
  *
+ * @example
+ * ```js
+ * import { useAbortSignal } from "effection";
+
+ * function* example() {
+ *   let signal = yield* scoped(function*() {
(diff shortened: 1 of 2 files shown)


lib/scoped.ts
```diff
@@ -9,6 +9,18 @@ import { useCoroutine } from "./coroutine.ts";
...
+ * @example
+ * ```js
+ * import { useAbortSignal } from "effection";
+
+ * function* example() {
+ *   let signal = yield* scoped(function*() {
+ *     return yield* useAbortSignal();
+ *   });
+ *   return signal.aborted; //=> true
+ * }
```
taras, line 21: This is a beautiful little piece of code 😍

taras (approved):

PR #961, opened 2025-01-09
📝 document scope, spawn

cowboyd:
## Motivation

Get the Scope jsdocs up to speed in preparation for the v4 release. 

## Approach

In addition to the jsdoc, the `ScopeInternal` class was moved to its own file so that it can be imported internally, but is not exported from the main module.

### DIFF
--- a/lib/spawn.ts
+++ b/lib/spawn.ts
@@ -6,8 +6,8 @@ import type { Effect, Operation, Task } from "./types.ts";
 /**
  * Run another operation concurrently as a child of the current one.
  *
- * The spawned operation will begin executing immediately and control will
- * return to the caller when it reaches its first suspend
+ * The spawned operation will begin executing at the next available
+ * opportunity.
  *
  * @example
  * ```js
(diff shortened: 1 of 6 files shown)


lib/spawn.ts
```diff
@@ -1,8 +1,57 @@
...
+ *    throw new Error('boom!');
+ *   });
+ *
+ *   console.log('this code will run and probably should not');
+ *
+ *   yield* suspend(); // <- error is thrown after this.
+ * });
+ * ```
+ * @param operation the operation to run as a child of the current task
+ * @typeParam T the type that the spawned task evaluates to
```
taras, line 30: This should say `@typeParam {T}`

taras (changes requested):
Just need to write `{T}` so I'll have something to match against in the doc generator.

lib/spawn.ts
cowboyd: Should it always be the very first word following `@typeParam`?

lib/spawn.ts
```diff
@@ -1,8 +1,57 @@
...
+ * being executed that should not.
+ *
+ * @example
+ * ```js
+ * import { main, suspend, spawn, useScope } from 'effection';
+ *
+ * await main(function*() {
+ *   yield* useScope();
+ *
+ *   scope.run(function*() {
```
taras: @cowboyd this looks like a mistake. Should this be yielding?

cowboyd: Yes, this is outdated actually, so I was able to just remove it.

lib/spawn.ts
cowboyd: After discussing in discord, we're opted to try identical handling to `@param` which uses the first word following the annotation followed by an  `"-"`

taras (approved):

PR #966, opened 2025-01-22
Run tests in CI for Node 18, 20, 22

taras:
## Motivation

I found a bug when running Effection v4 in Node 18 that was not there in Node 20. I tracked it to `toReversed`. I wanted to fix it but I found that we don't actually run tests in Node.

## Approach

1. Run tests in Node in CI for 18, 20, and 22.
2. I had to replace useCommand with tinyexec because Deno.Command doesn't have a shim for Node.


tasks/built-test.ts
```diff
@@ -0,0 +1,37 @@
...
+  importMap: "deno.json",
+  package: {
+    // package.json properties
+    name: "effection-tests",
+    version: "0.0.0",
+    sideEffects: false,
+  },
+});
+
+await Deno.copyFile("README.md", `${outDir}/README.md`);
```
cowboyd: Not strictly necessary, right?
```suggestion
```

cowboyd (commented):
This is a major enhancement in our coverage, awesome!

 Since we now are running the full test suite including `maint.test.ts` in NodeJS, I think we can get rid of the `test:node` task in `deno.json` and the corresponding `.mjs` file which just ensured that the `main()` function worked, but nothing else.

taras:
@cowboyd Repurposed node:test but I removed the node.mjs file.

cowboyd (approved):

PR #971, opened 2025-02-01
Fix typo in context docs

just-be-dev:
I was reading through the v4 docs and noticed a small typo.


cowboyd (approved):
Nice catch! Thanks 👍🏻

PR #973, opened 2025-02-10
🐛 resolve operation overload before simple function overload

### LINKED ISSUE #972: Incorrect Type Inference for call in Effection 4.0.0-alpha.6
There appears to be an issue with the type definitions of **`call`** in Effection v4 where the overload for `() => Operation<T>` is overshadowed by the overload for `() => T`. 
This causes unexpected type errors when returning an Operation<string>, and it may affect other types as well.

Steps to Reproduce
1. Define an operation that returns a string:

```ts
function* computeHashString(): Operation<string> {
  yield* sleep(1000);
  return "a string";
}
```
2.  Attempt to use call:
```ts
import { call, Operation } from "effection";

const hashString = yield* call(() => computeHashString());
console.log('the hash string is:', hashString); // correctly logs "a string"

// Type of hashString is Operation<string>, so this comparison causes a TS error:
if (hashString === "a string") {
  // Type error
}
```

3. Alternatively, calling `call(computeHashString)`  triggers a different type error:
```txt
No overload matches this call.
Argument of type 'Operation<string>' is not assignable to parameter of type '() => Operation<unknown>'
...
```

Expected Behavior

  `call(computeHashString)` should match the overload that accepts () => Operation<T> rather than failing to match any overload.

Actual Behavior

   The overload that accepts () => T is prioritized before the one that accepts () => Operation<T>, causing the compiler to complain.

as @cowboyd  pointed out on Discord conversation:

the order of overloads in Effection is:
```ts
export function call<T, TArgs extends unknown[] = []>(
  fn: (...args: TArgs) => Promise<T>,
): Operation<T>;
export function call<T, TArgs extends unknown[] = []>(
  fn: (...args: TArgs) => T,
): Operation<T>;
export function call<T, TArgs extends unknown[] = []>(
  fn: (...args: TArgs) => Operation<T>,
): Operation<T>;
```
and reversing the second and third overloads should fix the problem:
```ts
export function call<T, TArgs extends unknown[] = []>(
  fn: (...args: TArgs) => Promise<T>,
): Operation<T>;
export function call<T, TArgs extends unknown[] = []>(
  fn: (...args: TArgs) => Operation<T>,
): Operation<T>;
export function call<T, TArgs extends unknown[] = []>(
  fn: (...args: TArgs) => T,
): Operation<T>;
```
Effection version: v. 4.0.0-alpha.6
Typescript version: 5.2.2

i.e.
Swapping the overloads so that () => Operation<T> is checked before () => T resolves the type error.

cowboyd:
## Motivation

fix #972

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -69,11 +69,12 @@ export function call<T, TArgs extends unknown[] = []>(
   fn: (...args: TArgs) => Promise<T>,
 ): Operation<T>;
 export function call<T, TArgs extends unknown[] = []>(
-  fn: (...args: TArgs) => T,
-): Operation<T>;
-export function call<T, TArgs extends unknown[] = []>(
   fn: (...args: TArgs) => Operation<T>,
 ): Operation<T>;
+export function call<T, TArgs extends unknown[] = []>(
+  fn: (...args: TArgs) => T,
+): Operation<T>;
+
 export function call<T, TArgs extends unknown[] = []>(
   callable: Callable<T, TArgs>,
   ...args: TArgs


taras (approved):
Is this only a problem in V4?

cowboyd:
@taras good question. I don't think so, although e3 prioritizes `Operation` functions over async functions 🤷🏻 

https://github.com/thefrontside/effection/blob/v3/lib/call.ts#L95-L102

PR #979, opened 2025-02-28
Effect-TS benchmarks

joshamaju:
## Motivation

Adds benchmarks for the Effect library

## Approach

Recursion and events benchmarks


cowboyd:
Thanks @joshamaju 🙏🏻

joshamaju:
Always happy to help

PR #981, opened 2025-03-06
Effect benchmark

joshamaju:
## Motivation

Compares start-up times of effect and effection

## Approach

Exit operation early and capture time at moment of exit

### Alternate Designs

Use an annotation context provider that can capture more than just start-up time

## Screenshots

Screenshot 2025-03-06 at 15 09 23


joshamaju:
Making some changes to this

cowboyd:
@joshamaju Looks like you might be using Prettier in your IDE instead of Deno format?

joshamaju:
Yes, sorry about that. My deno setup isn't working properly.

PR #982, opened 2025-03-06
🎒backport `each()` 🐛 fixes to v3

cowboyd:
## Motivation

There were two problems with `each()` in v3 that were fixed in v4.

1. EachLoops are stacked so that you can iterate within an iteration, within an iteration, and so on. This is appropriate when all of the iterations are part of the same coroutine because they will be executed in serial. In other words, you can only be in a single iteration at a time. The problem arises when you have the scope of one coroutine inheriting from the scope of another, but those coroutines are executing concurrently. This can result in the loop stack being changed during the middle of an execution and all kinds of mayhem can ensue.
2. The subscription for an each loop is never being released because it is not running in its own resource. That means if you have an iteration within an iteration, it pile up subscriptions which could lead to memory leaks.

## Approach

This makes sure that every unique scope (coroutine) has its own stack of each loops, and that they run in a resource which is exited when the loop is completed.

### DIFF
--- a/lib/each.ts
+++ b/lib/each.ts
@@ -1,5 +1,8 @@
 import type { Operation, Stream, Subscription } from "./types.ts";
 import { createContext } from "./context.ts";
+import { useScope } from "./run/scope.ts";
+import { resource, spawn } from "./instructions.ts";
+import { withResolvers } from "./with-resolvers.ts";
 
 /**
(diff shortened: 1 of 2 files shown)


taras (approved):
Very nice. Great to have this fixed in v3

PR #985, opened 2025-03-28
Fixed run function example

garyhe0780:
## ISSUE

example code in run function comment is wrong, we should wrapper async function inside call before yield async function


cowboyd (approved):
🙏🏻  Thank you!

taras (approved):

PR #988, opened 2025-04-03
Introducing until helper to replace call(promise)

taras:
## Motivation

It was very convenient to use `call(promise)` but it created disharmony with `call` function in JavaScript. We wanted to keep that functionality with `until` function. 

## Approach

Added `until` which does `call(() => promise)`


cowboyd (approved):
LGTM!

PR #990, opened 2025-04-03
Adding until to v4 to replace call(promise)

taras:
## Motivation

Same as #988 but for v4

## Approach

* Added until
* Updated docs

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -101,6 +101,22 @@ export function call<T, TArgs extends unknown[] = []>(
   };
 }
 
+/**
+ * It can be used to treat a promise as an operation. This function
+ * is a replacement to the v3 deprecated `call(promise)` function form.
+ *
+ * @example
+ * ```js
+ * let response = yield* until(fetch('https://google.com'));
+ * ```
+ * @template {T}
+ * @param promise
+ * @returns {Operation<T>}
+ */
+export function until<T>(promise: PromiseLike<T>): Operation<T> {
+  return call(async () => await promise);
+}
+
 function isPromise<T>(
   target: Operation<T> | Promise<T> | T,
 ): target is Promise<T> {
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -101,22 +101,6 @@ export function call<T, TArgs extends unknown[] = []>(
   };
 }
 
-/**
- * It can be used to treat a promise as an operation. This function
- * is a replacement to the v3 deprecated `call(promise)` function form.
- *
- * @example
- * ```js
- * let response = yield* until(fetch('https://google.com'));
- * ```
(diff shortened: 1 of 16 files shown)


lib/call.ts
```diff
@@ -101,6 +101,22 @@ export function call<T, TArgs extends unknown[] = []>(
...
+ * let response = yield* until(fetch('https://google.com'));
+ * ```
+ * @template {T}
+ * @param promise
+ * @returns {Operation<T>}
+ */
+export function until<T>(promise: PromiseLike<T>): Operation<T> {
+  return call(async () => await promise);
+}
+
```
cowboyd: Let's implement `until()` directly with an `action()`. There is no need to go through the full `call()` logic every time we want to await a promise.

Also, `until()` is its own API independent from `call()` that belongs in its own file with its own test. That way, `call()` can depend on `until()` and not the other way around.

test/call.test.ts
```diff
@@ -22,6 +22,10 @@ describe("call", () => {
     )).resolves.toEqual(42);
   });
 
+  it("evaluates a promise directly with `until`", async () => {
+    await expect(run(() => until(Promise.resolve(42)))).resolves.toEqual(42);
+  });
+
```
cowboyd: We should put this in its own file and check both the negative, positive, and halt cases.

cowboyd (dismissed):

lib/call.ts
taras: 👍

test/call.test.ts
taras: @cowboyd what would you expect to happen in the halt case?

lib/until.ts
```diff
@@ -0,0 +1,21 @@
+import { action } from "./action.ts";
+import type { Operation } from "./types.ts";
+
+/**
+ * It can be used to treat a promise as an operation. This function
+ * is a replacement to the v3 deprecated `call(promise)` function form.
```
cowboyd: I think the callout to the v3 usage makes sense in v3, but in v4, I think we can leave that part out.
```suggestion
 * Treat a promise as an {@link Operation}
```

lib/until.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @example
+ * ```js
+ * let response = yield* until(fetch('https://google.com'));
+ * ```
+ * @template {T}
+ * @param promise
+ * @returns {Operation<T>}
```
cowboyd:
```suggestion
 * @returns {Operation<T>} that succeeds or fails depending on the outcome of `promise`
```

cowboyd (commented):
A few documentation suggestions, but LGTM.

PR #991, opened 2025-04-10
Fixed run example

taras:
## Motivation

Code signing contributions

Closes #985


cowboyd (approved):
Do we also need to patch this in to v4?

PR #995, opened 2025-04-15
🚛 Backport `Scope.spawn()` to v3

cowboyd:
## Motivation

In V4, `scope.spawn()` is encouraged when starting background tasks from an existing operation (as opposed to `scope.run()`). 

In order to make forward compatible apps, this API should be in v3.

## Approach

This adds a trivial implementation of spawn that delegates to `scope.run()`


taras (approved):
So simple!

PR #1004, opened 2025-06-19
🔥 get rid of unused dependency on assert.

cowboyd:
## Motivation

This code is completely unused and removing it has zero effect on the test and packaging of code.

## Approach

🔪 cut it!


taras (approved):
👍

PR #1005, opened 2025-06-19
✨ add interval() helper to consume a stream of intervals

cowboyd:
## Motivation

[`setInterval()`][1] is a core JavaScript API that allows you to setup global state. As such, there really isn't any reason that you shouldn't consume it using structured concurrency.

## Approach
This adds an `interval()` function that constructs a stream that emits an item every time that the interval is triggered.

```ts
let startTime = Date.now();

for (let _ of yield* each(interval(10))) {
  let elapsed = Date.now() - startTime;
  console.log(`elapsed time: ${elapsed} ms`);
  yield* each.next();
}
```

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Window/setInterval

### DIFF
--- a/lib/interval.ts
+++ b/lib/interval.ts
@@ -1,0 +1,31 @@
+import { resource } from "./instructions.ts";
+import { createSignal } from "./signal.ts";
+import type { Stream } from "./types.ts";
+
+/**
+ * Consume an interval as an infinite stream.
+ *
+ * ```ts
+ * let startTime = Date.now();
+ *
+ * for (let _ of yield* each(interval(10))) {
+ *   let elapsed = Date.now() - startTime;
+ *   console.log(`elapsed time: ${elapsed} ms`);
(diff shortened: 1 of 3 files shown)


taras (commented):
Do you know if this change remains compatible with https://docs.deno.com/examples/mocking_tutorial/#faking-time?

cowboyd:
@taras It definitely should because the interval helper is just another caller to `setInterval()` in the same way that `sleep()` is just a client of `setTimeout()`. It doesn't patch in any way.

We could actually use it to write a test to find out! .... but we'd have to make sure that it was compatible with nodejs

taras:
🤦 I'm sorry, I read this in passing. I didn't notice that this is a contrib update - it doesn't change the underlying sleep command.

taras (approved):
👍

jbolda (approved):
Seems great with me!

PR #1007, opened 2025-07-18
Fix webpack compatibility

rrthomas:
This is related to PR #935. Unfortunately, webpack follows even dynamic
imports unless told not to. This one-line change adds the magic comment to
tell webpack not to try to import ‘node:process’.

## Motivation

This PR fixes webpack compatibility.

## Approach

This PR adds a single comment to tell webpack not to try to import
node:process.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -91,8 +91,10 @@ export async function main(
             }
           },
           *node() {
+            // Annotate dynamic import so that webpack ignores it.
+            // See https://webpack.js.org/api/module-methods/#webpackignore
             let { default: process } = yield* call(() =>
-              import("node:process")
+              import(/* webpackIgnore: true */ "node:process")
             );
             hardexit = (status) => process.exit(status);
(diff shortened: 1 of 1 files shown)


jbolda:
Appreciate the PR!

Two quick asks, and we can get it merged a published.
- The comment previous to this has a preview package that you can test it out with. Give it a quick test to confirm. (Feel free to use this until the updated version comes out as well.)
- Do you have a link or reference to documentation that notes this? If you do, we could add that as a comment in the code. It would be helpful to reference in continuing to maintain support with future / past webpack versions.

rrthomas:
Many thanks for the quick review. I have added a comment with a link to webpack documentation. I will test the preview package and report back.

rrthomas:
Confirmed that the PR preview package fixes working with webpack for me.

jbolda (approved):
Perfect, thanks! Will get it published right quick too!

PR #1020, opened 2025-08-10
Clean up actions page in v3

taras:
## Motivation

A new Effection user reported actions page being incorrect. It turned out that it was using examples with deprecated APIs.

## Approach

1. Replaced actions page with content from v4
2. Renamed `Actions and Suspension` to `Actions` and move it to the bottom of the sidebar section
3. Move Resources after Spawn
4. Removed broken badges from installation page
5. Fixed the JSR url on the installation page


cowboyd (approved):
Can we also do the rename on the v4 page?

taras:
@cowboyd https://github.com/thefrontside/effection/pull/1021

PR #1024, opened 2025-08-20
📝 Docs migration guide

### LINKED ISSUE #909: API Changes and Deprecation List for V4
Certain APIs are going to need to change, although not very much, for v4. We should make a list of them.

- `call()` This is very overused for noting scope, and also for evaluating promise. We are going to deprecate the bare promise
  - add notes on the what is changing about `call()`, and how to upgrade
  - add `scoped()` to V3. It will be just an alias to `call()`. One option for upgrade. Upgrade would involve replace call() -> scope() then fix TS errors.
- `Scope#run` -> `run(op, { scope })`
  - ideally replace `Scope#run` -> `Scope#spawn`
- `action()`: Action will become more like the promise constructor and will not take an operation function, but instead will take a simple function that receives a resolver to be called when the effect resolves. It is more a general effect resolver than anything else.

- [ ] update action jsdocs on the website to be curret

### LINKED ISSUE #1018: Create Migration Guide
- [ ] differences in Callable()
  - [ ]  WAS
    - [ ] promise, constant, operation, generation function, async function, sync function
    - [ ] encapsulate tasks by creating a new scope 
    - [ ] created a new task / iterator (not directly part of API, just a side-effect)
    - [ ] created an error boundary
  - [ ] IS
    - [ ] operation function, asynchronous function, sync function i.e. call a function. Only pass functions to call (nothing else)
      - [ ] promise, use `until()` helper `yield* call(promise)` -> `yield* until(promise)`
      - [ ] constant, use `constant()` helper `yield* call(5)` -> `yield* constant(5)`
      - [ ] operation, use the operation: `yield* call(operation)` -> `yield* operation`
      - [ ] use `lift` for an option to call(constant)
    - [ ] use `scoped()` helper for encapsulating tasks and creating error boundary.
    - [ ] does not create a new iterator.
- [ ] differences in `action()`
  - [ ] an action callback is always synchronous
  - [ ] old API has been deprecated, if you fix all deprecations then it will work in v4
- [ ] explain task priority
- [ ] add more context to `lift`, see discord thread

cowboyd:
## Motivation

Resolve https://github.com/thefrontside/effection/issues/1018 #909

### DIFF
--- a/docs/structure.json
+++ b/docs/structure.json
@@ -4,7 +4,8 @@
     ["typescript.mdx", "TypeScript"],
     ["thinking-in-effection.mdx", "Thinking in Effection"],
     ["async-rosetta-stone.mdx", "Async Rosetta Stone"],
-    ["tutorial.mdx", "Tutorial"]
+    ["tutorial.mdx", "Tutorial"],
+    ["upgrade.mdx", "Upgrading from V3"]
   ],
   "Learn Effection": [
     ["operations.mdx", "Operations"],
--- a/docs/upgrade.mdx
+++ b/docs/upgrade.mdx
@@ -1,0 +1,225 @@
+For the most part, the changes between Effection versions 3 and 4 are
+internal, and while the public API remains largely untouched , there
+were some places where it was appropriate to make breaking changes.
+However, in the cases where we did, it was guided by our simple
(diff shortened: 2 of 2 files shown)


taras (approved):
👍

PR #1028, opened 2025-08-29
Bring the website back into this repository

taras:
## Motivation

We wanted to bring the website back into effection repository since Deno now supports having multiple projects in one repository. In the process, we also added the ability to render the website by reading content from git repositories instead of via GitHub API.

## Approach

It was actually quite a bit of work to be able to read content from git repositories, but it looks like the result was worth while. I'll explain how it works. There are 2 primary abstractions to do most of the heavy lifting,

## Fetch API

The fetch API is based on middleware. I was originally going to use `@effectionx/context-api` but I couldn't because the website uses effection 3, so I copy and pasted it into this repository. The fetch API provides the base abstraction for reading contents from URLs. This is helpful when we need to read data from GitHub API or GitHub website URLs. For example, JSR references GitHub respository URLs. By using the fetch context API, we're able to use a standard interface for fetching data and configure that interface separately.

## Repository

Repository is an abstraction for working with repositories. There are two kinds: Git and Octokit. Git read from the file system and Octokit from GitHub API. Both providers are available for reading files, but only Git provider is used for reading files. Octokit provider is used for fetching stars. We should really publish the Git API as a separate package.


cowboyd (commented):
Can we put `AGENTS.md` into a `.agents` or something? I'd like to maintain root dir hygiene if possible

www/README.md
```diff
@@ -0,0 +1,20 @@
+## Effection Website
+
+The Effection website contains Guides and API documentation.
```
cowboyd: We should include a phrase about how it introspects the Effection repository at various branches and tags as the source of documentation.

cowboyd (commented):
😱 😱 😱 

Nice!

taras:
> Can we put AGENTS.md into a .agents or something? I'd like to maintain root dir hygiene if possible

@cowboyd unfortunately, according to https://agents.md, it has to be in the root. 

<blockquote><div><strong><a href="https://agents.md/">AGENTS.md</a></strong></div><div>AGENTS.md is a simple, open format for guiding coding agents. Think of it as a README for agents.</div></blockquote>

www/README.md
taras: I added more information

cowboyd:
> Commits must have verified signatures.
You might want to scrub Claude off the commit authorship, or else sign on its behalf (not quite sure how to do that)

cowboyd (approved):

cowboyd:
Checks appear borked.

PR #1031, opened 2025-10-04
Do not bind SIGTERM in Windows

taras:
## Motivation

Problem 1: Binding SIGTERM in windows in Deno triggers this warning `Windows only supports ctrl-c (SIGINT), ctrl-break (SIGBREAK), and ctrl-close (SIGUP)`. We don't want `main` to produce any warnings.

To verify that this is working on Windows, I ran tests on windows. Which lead to Problem 2: `TypeError: Windows only supports ctrl-c (SIGINT), ctrl-break (SIGBREAK), and ctrl-close (SIGUP), but got SIGTERM` (note the _but got SIGTERM_)

## Approach

The solution to problem 1 is to avoid binding SIGTERM in Deno and Node using their platform flags.

The solution to problem 2 is to use ctrlc-windows instead of sending SIGINT to terminate in Windows.

The test produced stderr which caused the test to fail in pwsh. In bash, it would hang because bash is capturing command's stderr which disrupted ctrlc-windows function. In `@effectionx/process` we use ctrlc-windows with a process in detached mode which bypasses this problem. I used `cmd` shell which doesn't disrupt stderr like bash, nor fails the process when stderr is produced.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,19 @@
+# Set default line ending behavior
+* text=auto
+
+# Explicitly set line endings for text files
+*.ts text eol=lf
+*.js text eol=lf
+*.json text eol=lf
+*.md text eol=lf
+*.yml text eol=lf
+*.yaml text eol=lf
+
+# Binary files
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.woff binary
+*.woff2 binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,6 +3,7 @@
 
 # Explicitly set line endings for text files
 *.ts text eol=lf
+*.mjs text eol=lf
 *.js text eol=lf
 *.json text eol=lf
 *.md text eol=lf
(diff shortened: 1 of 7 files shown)


cowboyd:
Odd. It looks like the tests don't fail on Windows, but they don't pass either.

taras:
It's wierd that looks like it succeeds, but I can reproduce it locally that in pwsh it's returning status code 1. It's defaulting to pwsh, but passes successfully on bash, so I'm just going to switch it to bash.

taras:
Now it's hanging on the main.test.ts in Windows bash.

taras:
The test produced stderr which caused the test to fail in pwsh. In bash, it would hang because bash is capturing command's stderr which disrupted ctrlc-windows function. In @effectionx/process we use ctrlc-windows with a process in detached mode which bypasses this problem. Here, I had to use cmd shell which doesn't disrupt stderr like bash, nor fail the process when stderr is produced.

.gitattributes
```diff
@@ -0,0 +1,20 @@
+# Set default line ending behavior
```
cowboyd, line 1: It's good to have .gitattributes, but seems unrelated to this PR.

taras: I needed to add it for git to correct line breaks for each platform. Otherwise, linebreaks get messed up in linux or windows.

cowboyd (approved):

PR #1042, opened 2025-12-15
🚚 forward port www from v3 -> v4

cowboyd:
## Motivation

We brought in the main website back into the main branch, but we need to forward port it to v4

## Approach

Copy without history the `www/` directory from one to the other, and fix the issues :)


taras (approved):
We can now eliminate www/e4.ts because it's all v4 now. I can do it later if you don't get to it first.

PR #1065, opened 2025-12-27
📝 document equivalents of `Promise.then`

### LINKED ISSUE #944: Add equivalent to `Promise.then` to FAQ

### LINKED ISSUE #946: Add equivalent to shared `Promise` to FAQ

cowboyd:
## Motivation

Resolve https://github.com/thefrontside/effection/issues/944
Resolve https://github.com/thefrontside/effection/issues/946

### DIFF
--- a/docs/faq.mdx
+++ b/docs/faq.mdx
@@ -41,7 +41,104 @@ Object.defineProperty(Promise.prototype, Symbol.iterator, {
 
 ## Why can't I use `call(Promise.resolve(42))` anymore?
 
-[until()](/api/v4/until) function offers a convenient alternative to passing a promise to `call`. In Effection v3, it was possible 
-to `yield* call(Promise.resolve(42))` but we removed this in Effection v4 because we wanted to align call closer 
-to `Function.prototype.call`. You can still use `yield* call(() => Promise.resolve(42))` and 
-`yield* call(async function() {})`.
+In Effection v3, it was possible to `yield* call(Promise.resolve(42))` but we removed this in Effection v4
+because we wanted to align call closer to `Function.prototype.call`. You can still use 
+`yield* call(() => Promise.resolve(42))` and `yield* call(async function() {})`. In v4, we plan to ship 
+a new function called `when` which will take a promise and wrap it in a constructor for convenience. 
+
+## How can I chain operations together like I do with promises?
+
+In JS, it's common in a function that isn't async to chain promises together using `.then()` and `.catch()`
+
+The simplest way is to use the [Chain](https://frontside.com/effection/x/chain/) helper from [effectionx](https://frontside.com/effection/x)
+
+```typescript
+import { Chain } from "@effectionx/chain";
+
+Chain.from(operation)
+  .then(lift((val) => val * 2)),
+  .then(lift((val) => String(val)),
+  .then(lift((val) => val.toUpperCase())),
+```
+
+Note the usage of [`lift()`](../../api/v4/lift) to convert normal synchronous functions into operations.
+
+But composition is so simple, that rolling your own is almost trivial.
+For example, you can make make a "pipeable" `then()` operation like
+so:
+
+```typescript
+function then<A,B>(fn: (value: A) => Operation<B>): (operation: Operation<A>) => Operation<B> {
+  return function*(operation) {
+    return yield* fn(yield* operation);
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/faq.mdx
```diff
@@ -41,7 +41,104 @@ Object.defineProperty(Promise.prototype, Symbol.iterator, {
...
+```typescript
+function then<A,B>(fn: (value: A) => Operation<B>): (operation: Operation<A>) => Operation<B> {
+  return function*(operation) {
+    return yield* fn(yield* operation);
+  }
+}
+```
+
+You can then take an off the shelf pipe function (they're a dime a
+dozen from places like lodash, fp-ts, or even the code you have your
```
jbolda: up*

jbolda (approved):

taras (approved):

PR #1068, opened 2025-12-28
Fix v3/v4 navigation to non-existent page and mobile sidebar in dark mode

### LINKED ISSUE #1058: Guide navigation links don't work if the same page does not exist on other guide.
To reproduce, go to the 'upgrade' page on v4 https://frontside.com/effection/guides/v4/upgrade/

Then, click on the `v3` link. It will try to go to https://frontside.com/effection/guides/v3/upgrade/ which does not exist.

The links should just go to the guide index, not try and match pages which don't exist.

taras:
## Motivation

Fixes #1058 and 
image

## Approach

* when target version doesn't have the same page link to docs index
* Added dark most styles for sidebar
* Updated FAQ to mention until

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -2,45 +2,73 @@
 
 ## Our Pledge
 
-In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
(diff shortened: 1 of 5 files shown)


cowboyd (approved):
Seems like there are a lot of unrelated formatting changes. Did the defaults changed?

frontsidejack:
No, the README.md were never formatted using `deno fmt` because they are excluded from formatting in deno.json. I formatted them because I was edit the files. We should probably remove them from excludes in deno.json. What do you think?

taras:
Oops. I was signed in with frontsidejack in the browser.

cowboyd:
Agree. No need to make exceptions on formatting.

Issue #1029, opened 2025-09-05
Add AI friendly TL;DR to the docs


taras:
This TL;DR was generated (by Claude Code) from the complete Effection documentation (22,242 tokens) with the following requirements:

- Keep under 5,000 tokens for conciseness
- Wrap all code examples in `main` or `run` for proper execution context
- Use resources only when return values need to be tied to parent operation lifecycle
- Always use try/finally blocks around `provide` in resource examples
- Order content in learning sequence (getting started → core concepts → advanced patterns)
- Simplify resource pool example to avoid explicit release management

The original documentation spans 15 MDX files covering installation, TypeScript usage, core concepts, operations, spawning, resources, collections/streams, events, error handling, context, actions, scope management, and process handling.

# Effection TL;DR

**Structured Concurrency and Effects for JavaScript** - A library that makes async JavaScript code more predictable, composable, and cancellable.

### Installation
```bash
# Node.js/Browser
npm install effection

# Deno  
import { main } from "jsr:@effection/effection@3";
```

### Your First Program
```js
import { main, sleep } from 'effection';

await main(function*() {
  console.log('Hello');
  yield* sleep(1000);
  console.log('World!');
});
```

### 1. Operations vs Promises
- **Operations** are stateless recipes that describe what to do
- **Promises** are stateful and execute immediately
- Operations only run when evaluated with `yield*` or `run()`

```js
import { run, until } from 'effection';

// Promise - runs immediately
const promise = fetch('/api');

// Operation - only runs when evaluated
function* fetchData() {
  return yield* until(fetch('/api'));
}

await run(fetchData);
```

### 2. The Rosetta Stone

| Async/Await | Effection |
|-------------|-----------|
| `await` | `yield*` |
| `async function` | `function*` |
| `Promise` | `Operation` |
| `new Promise()` | `action()` |
| `for await` | `for yield* each` |
| `AsyncIterable` | `Stream` |

### 3. Structured Concurrency
- **No operation runs longer than its parent**
- **Every operation exits fully** (guaranteed cleanup)
- Child operations are automatically cancelled when parent completes

```js
import { main, spawn, sleep } from 'effection';

await main(function*() {
  yield* spawn(function*() {
    // This will be cancelled after 5 seconds
    for (let i = 0; i < 100; i++) {
      yield* sleep(1000);
      console.log(i);
    }
  });
  
  yield* sleep(5000); // Parent exits after 5s
});
```

### Entry Points
```js
[...]

PR #1069, opened 2025-12-28
llms.txt and llms-full.txt

### LINKED ISSUE #1030: llms.txt needed
just like something as https://effect.website/docs#docs-for-llms

taras:
## Motivation

Closes https://github.com/thefrontside/effection/issues/1030, https://github.com/thefrontside/effection/issues/1029

We want to make it easy for Effection users to write Effection programs with LLM's assistance. The most important consideration at the moment is the impact that documentation and access to documentation has on LLM context. [LLM context engineering](https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents) is evolving quickly, but there are already some conventions that we can expect to hang around for a while: progressive disclosure and tool friendly access.

The goal of progressive disclosure is to allow the AI agent to incrementally load documentation into its context based on what it needs instead of preloading information unnecessary for completing a specific task. This is accomplished using index pages that give the agent information about where it can find important information. The agent can rely on tools to load additional context. Tools includes abilities to read files from file systems, websites and create and execute scripts.

The agents can use different strategies to load information they need, but making that information easily available will provide a better experience for users by ensuring that agents can generate code quickly and generated code is correct. 

With this in mind, we need to consider 2 use cases:
1. Agent looking up information from the website
2. Agent looking up information from the file system

## Approach

This PR includes 2 files: llms.txt (800 tokens) and llms-full.txt (4000 tokens). llms.txt is a progressive disclosure index that links to markdown files in docs, TypeScript files in lib and references llms-full.txt. These files were generated by Claude Sonnet.

image


www/components/footer.tsx
```diff
@@ -31,6 +31,17 @@ export function Footer(): JSX.Element {
           Interactors <IconExternal />
         </a>
       </section>
+      <section class="flex flex-col gap-y-1">
+        <h1 class="text-sm uppercase font-bold text-blue-primary dark:text-blue-secondary mb-4">
```
jbolda, line 35: Pedantic, but shouldn't there only be one `h1` per page? Seems to be an existing pattern, so appropriate in a follow-up PR.

jbolda (approved):

www/components/footer.tsx
taras: That's a good catch. Updated it to h4

dmg-egg:
@taras What is going on with the LLM guidance in this project?

1. README.md links to `/llms.txt` and `/llms-full.txt` files which don't exist (but did, as as per PR).
2. Seemingly in their place are a pair of near-identical _screenshots_ of [`/www/assets/llms.txt`](https://github.com/thefrontside/effection/blob/v4/www/assets/llms.txt): [`/llms-txt-content.png`](https://github.com/thefrontside/effection/blob/v4/llms-txt-content.png) and [`/llms-txt-page.png`](https://github.com/thefrontside/effection/blob/v4/llms-txt-page.png) from a PR which doesn't document the additions of these images.

Context for me: I was planning to look into wrapping Effection around [Autocannon](https://github.com/mcollina/autocannon/) for a stress testing and flame chart tool I'm working on, so I came here to determine the best way to guide my agents. I can't make enough sense of the situation to be quite confident that this really warrants an Issue.

taras:
lol. Thank you for catching that. That's bizarre. I started cleaning up LLMS.txt file and stopped midway.  I'll fix it this weekend. the AGENTS.md file should be good though. Can you check?

taras:
That's great. Come hangout in our discord. We're happy to answer questions

PR #1073, opened 2026-01-05
Add support for async disposable

### LINKED ISSUE #1023: Make `Task` implement explicitly managed resource API
It should have both `Symbol.dispose` and `Symbol.asyncDispose`

Example: 

```js
async function main() {
  using task = run(function*() {
    try {
      yield* suspend();
    } finally {
      console.log('done');
    }
  });
}
```

joshamaju:
## Motivation

Cleanup effection task at application/effection boundary to avoid hanging operations.

> resolves #1023 

JavaScript now has a native scope based mechanism for cleanup called "Explicit Resource Management"

## Approach
This integrates this spec in two ways:

1. Every Task implements the AsyncDisposable api directly.
2. When calling `createScope()` the scope returned also implements AsyncDisposable interface

There is a bit of hackery that we needed to do with `createScope()` so that it can support both:

```ts
await using scope = createScope();
```
and

```ts
let [scope, destroy] = createScope();
```

### DIFF
--- a/lib/scope-internal.ts
+++ b/lib/scope-internal.ts
@@ -96,7 +96,7 @@ export function createScopeInternal(
   return [scope, destroy];
 }
 
-export interface ScopeInternal extends Scope {
+export interface ScopeInternal extends Scope, AsyncDisposable {
   contexts: Record<string, unknown>;
   ensure(op: () => Operation<void>): () => void;
 }
--- a/lib/scope.ts
+++ b/lib/scope.ts
@@ -35,15 +35,55 @@ export const global = createScopeInternal()[0] as Scope;
  * await destroy(); // prints "done!";
  * ```
  *
+ * `createScope()` also supports explicit resource management
+ * @example
+ * ```js
+ * {
+ *  await using scope = createScope();
+ *
+ *  let delay = scope.run(function*() {
+ *    yield* sleep(1000);
+ *  });
+ *
+ *  scope.run(function*() {
+ *    try {
+ *      yield* suspend();
+ *    } finally {
+ *      console.log('done!');
+ *    }
+ *  });
+ *
+ *  await delay;
+ *  // prints "done!";
+ * }
+ * ```
+ *
  * @param parent scope. If no parent is specified it will derive directly from {@link global}
  * @returns a tuple containing the freshly created scope, along with a function to
  *          destroy it.
  */
 export function createScope(
   parent: Scope = global,
-): [Scope, () => Future<void>] {
+): Scope & AsyncDisposable & [Scope, () => Future<void>] {
   let [scope, destroy] = createScopeInternal(parent);
-  return [scope, () => parent.run(destroy)];
+  let dispose = () => parent.run(destroy);
+
+  let tuple = [scope, dispose];
+
+  Object.defineProperty(scope, Symbol.iterator, {
+    value: tuple[Symbol.iterator].bind(tuple),
+    enumerable: false,
+  });
+
+  Object.defineProperty(scope, Symbol.asyncDispose, {
+    enumerable: false,
+    value: dispose,
+  });
+
+  return scope as unknown as
+    & Scope
+    & AsyncDisposable
+    & [Scope, () => Future<void>];
 }
 
 /**
--- a/lib/task.ts
+++ b/lib/task.ts
@@ -1,3 +1,4 @@
+/// <reference lib="esnext" />
 // deno-lint-ignore-file no-unsafe-finally
 import { DelimiterContext, ErrorContext } from "./delimiter.ts";
 import { createCoroutine } from "./coroutine.ts";
@@ -63,6 +64,10 @@ export function createTask<T>(options: TaskOptions<T>): NewTask<T> {
     [Symbol.toStringTag]: {
       enumerable: false,
       value: "Task",
+    },
+    [Symbol.asyncDispose]: {
+      enumerable: false,
+      value: () => task.halt(),
     },
   }) as Task<T>;
 
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -145,6 +145,8 @@ export interface Task<T> extends Future<T> {
    * Task is complete.
    */
   halt(): Future<void>;
+
+  [Symbol.asyncDispose](): Promise<void>;
 }
 
 /**
--- a/test/run.test.ts
+++ b/test/run.test.ts
@@ -347,4 +347,43 @@ describe("run()", () => {
 
     expect(scope.expect(Children).size).toEqual(0);
   });
+
+  it("can be disposed with 'await using'", async () => {
+    let halted = false;
+
+    {
+      await using _ = run(function* () {
+        try {
+          yield* suspend();
+        } finally {
+          halted = true;
+        }
+      });
+    }
+
+    expect(halted).toEqual(true);
+  });
+
+  it("should throw when 'halt' is invoked after 'await using'", async () => {
+    let halted = false;
+    let error: Error | undefined;
+
+    let task = run(function* () {
+      try {
+        yield* suspend();
+      } finally {
+        halted = true;
+      }
+    });
+
+    {
+      await using _ = task;
+    }
+
+    await task.catch((e) => error = e);
+
+    expect(error).toBeDefined();
+    expect(error?.message).toEqual("halted");
+    expect(halted).toEqual(true);
+  });
 });
--- a/docs/scope.mdx
+++ b/docs/scope.mdx
@@ -280,10 +280,18 @@ The nice thing about scope is that you don't need to worry about it. It's just
 there, ensuring that things get cleaned up as soon as they are no longer needed.
 Sometimes, however, it is necessary to interact with the scope of an operation
 from outside of Effection in normal JavaScript execution. To do this we use the
-[Scope][scope] API. Once we have a reference to a `Scope`, we can use it to run
+[Scope][scope] API.  Once we have a reference to a `Scope`, we can use it to run
 operations as though they were being spawned directly from an operation running
-in that scope. An example of this might be to spawn a new operation for each
-request to an express server.
+in that scope. The best way to use a scope is to capture it from an already running
+operation, but you can also create a new one from scratch.
+
+### Capturing a `Scope` from a running `Operation`
+
+If you are in an operation, the `useScope()` function allows you to
+capture a reference to the current `Scope`. You can then use that
+scope to create tasks that are descendants of that original scope. An
+example of this would be to spawn a new operation for each request to
+an express server.
 
 Express handles the http request, but it does not know anything about
 Effection, and so we have to explicitly connect each request to an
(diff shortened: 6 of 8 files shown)


lib/types.ts
```diff
@@ -52,7 +52,9 @@ export interface Operation<T> {
  * things, if the operation resolves synchronously, it will continue within the
  * same tick of the run loop.
  */
-export interface Future<T> extends Operation<T>, Promise<T> {}
+export interface Future<T> extends Operation<T>, Promise<T> {
+  [Symbol.dispose](): void;
+}
```
cowboyd: Shouldn't this be `Symbol.asyncDispose` on `Task`?

lib/types.ts
```diff
@@ -334,8 +336,9 @@ export interface Scope {
  * Analogous to the built in [`Awaited`](https://www.typescriptlang.org/docs/handbook/utility-types.html#awaitedtype) type.
  * `Yielded<Operation<T>> === T`
  */
-export type Yielded<T extends Operation<unknown>> = T extends
-  Operation<infer TYield> ? TYield
+export type Yielded<T extends Operation<unknown>> = T extends Operation<
+  infer TYield
+> ? TYield
   : never;
```
cowboyd, line 341: This is unrelated, and appears to be an artifact of your IDE?

lib/task.ts
```diff
@@ -74,6 +75,12 @@ export function createTask<T>(options: TaskOptions<T>): NewTask<T> {
       enumerable: false,
       value: "Task",
     },
+    [Symbol.asyncDispose]: {
+      enumerable: false,
+      value: async function () {
+        return await this.halt();
+      },
```
cowboyd: I don't think there is any need to wrap it in an extra function.

```suggestion
      value: () => task.halt(),
```

test/run.test.ts
```diff
@@ -13,9 +13,11 @@ import {
...
 describe("run()", () => {
   it("can run an operation", async () => {
-    await expect(run(function* () {
-      return "hello";
-    })).resolves.toEqual("hello");
+    await expect(
+      run(function* () {
+        return "hello";
+      }),
+    ).resolves.toEqual("hello");
```
cowboyd: This appears to be unrelated and from an IDE, not project settings.

test/run.test.ts
```diff
@@ -318,4 +320,45 @@ describe("run()", () => {
...
+    });
+
+    {
+      await using _ = task;
+    }
+
+    expect(halted).toEqual(true);
+  });
+
+  it("should throw when 'halt' is invoked after 'await using'", async () => {
```
cowboyd, line 367: I'm not sure I understand why this should be true. I'm not saying it shouldn't be true, I just don't quite understand what we're testing here.

lib/types.ts
joshamaju: You're right, it should be on `Task`

lib/types.ts
joshamaju: Can't seem to get it right, even `deno fmt` doesn't put it on the same line.

lib/task.ts
joshamaju: The wrapping is for grabbing a reference to task, or else `this` will point to the outer function.

test/run.test.ts
joshamaju: `deno fmt` also produces the same result.

test/run.test.ts
joshamaju: Check to ensure it aligns with the normal behaviour of manually calling halt more than one time.

lib/task.ts
cowboyd: Ah, I see. It can be done with a single, synchronous function though which will create one less promise.

test/run.test.ts
cowboyd: That's interesting. So on a clean checkout, `deno fmt` will reformat it to this? (it does not do this for me)

test/run.test.ts
cowboyd: Hmmm... halt should actually be idempotent. I think. This might be a bug in 4.0

test/run.test.ts
joshamaju: Yes. Do I need to configure it a certain way?

test/run.test.ts
cowboyd: If I just play with halting directly, it is completely idempotent:

```ts
import { main, sleep, spawn, suspend, until } from "effection";

await main(function*() {

  let task = yield* spawn(suspend);

  // ensure task actually is entered before halting.
  // makes no difference, but just to demonstrate
  yield* sleep(0);

  // halt as an aoperation
  yield* task.halt();
  yield* task.halt();
  yield* task.halt();

  // force halting as a promise
  yield* until(task.halt());
  yield* until(task.halt());
  yield* until(task.halt());



  console.log('done.');
})
```

joshamaju: It only throws on the `Future`. Please see the updated test.

cowboyd: Oh, ok. I see. I got confused. `_` _is_ the task. So after it is halted, via `Symbol.asyncDispose` then the actual task itself should fail. That makes sense. Perhaps it would be clearer written like:

```ts
let task = run(function* () {
  try {
    yield* suspend();
  } finally {
    halted = true;
  }
});

{
  await using _ = task;
}
await task.catch((e) = error = e);

expect(error).toBeDefined();
// ...
```

joshamaju: Exactly what I had in mind

joshamaju:
How do we move this forward?

lib/types.ts
```diff
@@ -145,6 +145,8 @@ export interface Task<T> extends Future<T> {
    * Task is complete.
    */
   halt(): Future<void>;
+
+  [Symbol.dispose](): void;
```
cowboyd: Shouldn't this be `[Symbol.asyncDispose](): Promise<void>`?

Also, as a non-enumerable property, does the explicit resource management even need to be part of the type? A user should never interact with it, expect with `using`

cowboyd (dismissed):
I agree, let's push this over the finish line.

1. Formatting changes: don't worry about it, I'll fix them up.
2. Question about the types
3. We should also add `asyncDispose` to `Scope` (along with a test)

lib/types.ts
joshamaju: "Shouldn't this be [Symbol.asyncDispose](): Promise<void>" - Yes, I'll make that correction.

"as a non-enumerable property, does the explicit resource management even need to be part of the type?" - Yes, without it you get a type error when using `using`.

joshamaju:
Please review the scope feature

lib/types.ts
```diff
@@ -145,6 +145,8 @@ export interface Task<T> extends Future<T> {
    * Task is complete.
    */
   halt(): Future<void>;
+
+  [Symbol.asyncDispose](): Promise<void>;
```
cowboyd, line 149: @joshamaju Do we need to document this? I'm not sure what the best practice is on documenting whether something is an explicitly manageable resource.

cowboyd:
@joshamaju I removed the formatting changes, and moved the scope implementation into ScopeInternal. 

The only remaining issue is documentation. I'm not sure how much is required for `asyncDispose`, and perhaps not at all.

lib/scope-internal.ts
```diff
@@ -55,6 +55,17 @@ export function createScopeInternal(
       destructors.add(op);
       return () => destructors.delete(op);
     },
+  }, {
+    [Symbol.asyncDispose]: {
+      enumerable: false,
+      value: () => {
+        if (parent) {
```
joshamaju: Does this prevent closing the scope when using `using`  inside effection code

cowboyd: I think it prevents closing the `global` scope, which is the only scope that does not have a parent.

cowboyd: I think it prevents closing the `global` scope, which is the only scope that does not have a parent.

taras:
We should add it to resource page to say that they're supported out of the box with this version.

lib/scope-internal.ts
joshamaju: That makes sense

joshamaju:
I've added a section to the upgrade guide documentation

cowboyd:
@taras I'm not sure the resources page isn't really the right place because resources are implicitly cleaned up. Only tasks and scopes implement AsyncDisposable, but not sure where the place to explain that is. The section on typescript maybe? Rosetta Stone?

cowboyd:
@joshamaju What do you think about making the _Array_ returned by `createScope()` actually implement the AsyncDisposable. This would allow us to say:

```ts
using [scope] = createScope();
```

joshamaju:
Is that even possible? I don't think it's possible

joshamaju:
So I just checked. Can only be used like this `using scope = createScope();`, and cannot be async, only `Symbol.dispose`.

cowboyd:
Yeah, it's not possible to use destructing assignment with explicitly managed resources.... another reason I don't like them. I had an AI tell me it was though, and so I got my hopes up. 🤣🤣🤣

I just think it will be awkward to use without being able to do it with a one-liner. Maybe we need a new method?

```ts
using scope = disposableScope();

using scope = createDisposableScope();
```

something like those or another alternative? ☝🏻

joshamaju:
Maybe `createScope` should return a object instead of a tuple?

cowboyd:
I don't think that would work either, would it? We could make the object Iterable for backwards compatibility if it did.

joshamaju:
Yep, it works

joshamaju:
Should it be a separate function i.e `createDisposableScope`?

cowboyd:
So we could define a `scope` property on the tuple?

joshamaju:
How do you mean?

cowboyd:
```ts
export function createScope(
  parent: Scope = global,
): [Scope, () => Future<void>] {
  let [scope, destroy] = createScopeInternal(parent);
  let tuple = [scope, () => parent.run(destroy)];
  let disposable = Object.defineProperty(scope, Symbol.asyncDispose, { value: () => parent.run(destroy) });
  Object.defineProperty(tuple, 'scope', { value: disposable });
}
```

This would then let us say:


```ts
using { scope } = createScope();
```

cowboyd:
Yeah, it's like I thought. Binding expressions are not supported inside using declarations https://github.com/microsoft/TypeScript/issues/55527

It's yet another example of how explicit resource management was railroaded through without concern over how much incongruence with existing patterns it would introduce.

cowboyd:
I think we can still do it, but it means we really have to abuse the runtime and the type system to make it work. I think what we can do is define `Symbol.iterator` _and_ `Symbol.asyncDispose` on the actual returned scope so that it can be used as a tuple and as an explicitly managed resource:

```ts
// 🤣 🤣 
declare function createScope(parent?: Scope): Scope & AsyncDisposable & [Scope, () => Promise<void>];
```

cowboyd:
This is horrendous, but it works:

```ts
export function createScope(
  parent: Scope = global,
): Scope & AsyncDisposable & [Scope, () => Future<void>] {
  let [scope, destroy] = createScopeInternal(parent);
  let dispose = () => parent.run(destroy);
  let tuple = [scope, dispose];
  Object.defineProperty(scope, Symbol.iterator, {
    enumerable: false,
    value: tuple[Symbol.iterator].bind(tuple),
  });
  Object.defineProperty(scope, Symbol.asyncDispose, {
    enumerable: false,
    value: dispose,
  })
  return scope as unknown as Scope & AsyncDisposable & [Scope, () => Future<void>];
}
```

joshamaju:
Is the goal that a user can use it in one of the following ways?

```ts
await using scope = createScope();
```
or like this
```ts
const [scope] = createScope();
```

cowboyd:
@joshamaju yes, that's the idea, although the second way is a no-no. You should always capture the `destroy()` function if you use it the second way.

It is a bit disgusting, but it let's you consume it either way.

joshamaju:
@cowboyd I've added the update.

I'm soo going to abuse `tuple[Symbol.iterator].bind(tuple)` in my own code 😅.

lib/scope.ts
```diff
@@ -7,6 +7,8 @@ import { createScopeInternal } from "./scope-internal.ts";
  */
 export const global = createScopeInternal()[0] as Scope;
 
+type DisposableScope = Scope & AsyncDisposable;
```
cowboyd: Let's not use a separate type so that we don't encourage folks to think about "disposable scopes and not disposable scopes". Specifically, the scope that you get from the tuple should _not_ be typed as disposable. The api should guide folks to _either_ use it as an asynchronous disposable, _or_ as a ref, destroy tuple.

lib/scope.ts
```diff
@@ -35,15 +37,53 @@ export const global = createScopeInternal()[0] as Scope;
...
+  });
+
+  Object.defineProperty(scope, Symbol.asyncDispose, {
+    enumerable: false,
+    value: dispose,
+  });
+
+  return scope as unknown as
+    & DisposableScope
+    & [DisposableScope, () => Future<void>];
```
cowboyd: As mentioned above, the scope inside the tuple should _not_ be disposable (even though it is at runtime) Because you should either use it as a disposable, or as an refe / destructor .

The type of createScope should be:

```ts
createScope(parent?: Scope): Scope & AsyncDisposable & [Scope, () => Future<void>]
```

That way, users are forced to choose how they will consume it.

joshamaju:
I've applied the changes as discussed

cowboyd:
@joshamaju I rebased on v4, squashed all the commits, fixed the formatting and added documentation to the repo as well as the PR. Anything else we should cover?

joshamaju:
I should push updates with the `using` function

cowboyd (approved):
Good to go! 🚀

PR #1075, opened 2026-01-12
✨ Allow peformance optimizations by setting co-routine iterator

cowboyd:
## Motivation

In order to perform iterator hoisting wherein all iterators are consumed within a single yield point, we need a way to modify a co-routine's iterator.

## Approach

This just adds a corresponding setter for the iterator.


cowboyd:
We don't need this since the optimization ended up just re-defining the `iterator` property

PR #1078, opened 2026-01-17
♻️ simplify spawn() by delegating to scope.run

cowboyd:
## Motivation
We're going to end up wrapping unified APIs around most Effection functions so that they can be decorated with middleware. As such, there is no need to have separate code paths for Scope.run() and Scope.spawn() since they end up calling the same thing under the hood.

## Approach
This has `spawn()` delegate to `Scope.run()` so that the apis can be unified and wrapped as one. In other words, wraping `Scope.run()` is the same as wrapping `spawn()`.

### DIFF
--- a/lib/spawn.ts
+++ b/lib/spawn.ts
@@ -1,7 +1,5 @@
-import { Ok } from "./result.ts";
-import type { ScopeInternal } from "./scope-internal.ts";
-import { createTask, type NewTask } from "./task.ts";
-import type { Effect, Operation, Task } from "./types.ts";
+import { useScope } from "./scope.ts";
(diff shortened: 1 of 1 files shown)


taras (approved):
🤯 how is this is even possible?

jbolda (approved):

cowboyd:
@taras I was actually surprised at how simple the change was. Turns out, the spawn operation is just this:

```ts
  let scope = yield* useScope();
  return scope.run(op);
```

taras:
Isn't run also that? How are they not same?

cowboyd:
They both resolve to `createTask()` This makes sure that we take the exact same codepath to get there.

PR #1081, opened 2026-01-19
🐛 fix: destroy task scope when halting to prevent memory leak

### LINKED ISSUE #1080: Effection v4 leaking memory when spawning many tasks
A discord community member shared a [reproduction of this problem](https://gistcdn.githack.com/ancorehraq/4d907ff55ecc5f8935aa7417b9e0f3ba/raw/ca38eb4b51ebdf1384c379b7e2841d088ea473ba/effection-leak.html) and confirmed that this is not a problem in v3.

> The first one is persistent. GC doesn't help, the buffers stick around. It seems to be a Chrome thing. I tried on Firefox, node and bun and they only leak effection state but not the ArrayBuffers.
> Second leak is universal. Allocates 1GB, then GCs down to 560MB or so that stays allocated.

I verified the reproduction on my machine. 

Btw, I tried adding `sleep(0)` between `spawn` and `halt` and after `half` - giving it some breathing room didn't make a difference.

cowboyd:
## Motivation

When a task was halted, the task itself would stop executing but its associated scope remained alive, causing a memory leak.

> resolves #1080 

## Approach

This fix makes the `halt()` operation not do any teardown itself, but instead it delegates to its own scope's `destroy()`. That way, not only is the task execution halted, but every scope in its entire tree is also destroyed. This ensures that all resources are cleaned up on `Task.halt()`.

The key bits are:

1. `halt()` is now just an alias to `destroy()`
2. all task settlement happens in scope destruction, so the task operation becomes

```ts
try {
  yield* top;
} finally {
  yield* destroy();
}
```

Which ensures that on success, failure, or halt, scope destruction always happens.

#### Note

Fixing this  did break one of the `each()` tests which relied on this bugged behavior, so the implementation has been updated to use a simple spawn instead of a resource.

Also, it broke a dubious behavior which is that a child could halt its parent:

```ts
let task: Task<void> = run(function* root() {
  yield* sleep(0);
  yield* spawn(function* child() {
    yield* task.halt();
  });
  yield* suspend();
});
```

I believe that the intuitive behavior here is deadlock, which is what now happens. In other words, the child computation uses scope hacking to capture a reference to its parent, and then depends on that parent's shutdown, which depends on waiting for the child to complete, which requires the child to shutdown, ad infinitum....

I think we can put in a special guard for this at some point and at least throw an exception here when we detect such an obvious deadlock, but I don't think it should hold up fixing an egregious memory leak such as this.

### DIFF
--- a/test/run.test.ts
+++ b/test/run.test.ts
@@ -3,13 +3,18 @@
 import { blowUp, createNumber, describe, expect, it } from "./suite.ts";
 import {
   action,
+  createScope,
   run,
+  type Scope,
   sleep,
   spawn,
   suspend,
   type Task,
   until,
+  useScope,
 } from "../mod.ts";
+import { Children } from "../lib/contexts.ts";
+import assert from "node:assert";
 
 describe("run()", () => {
   it("can run an operation", async () => {
@@ -207,7 +212,10 @@ describe("run()", () => {
     await expect(task).rejects.toMatchObject({ message: "halted" });
   });
 
-  it("can halt itself between yield points", async () => {
+  // TODO: this test is of dubious value. Deadlock might be the right thing here
+  // we can revisit and try to either detect this and deal with it, or maybe raise
+  // an error
+  it.skip("can halt itself between yield points", async () => {
     let task: Task<void> = run(function* root() {
       yield* sleep(0);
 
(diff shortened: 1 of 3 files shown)


test/run.test.ts
```diff
@@ -207,7 +212,7 @@ describe("run()", () => {
     await expect(task).rejects.toMatchObject({ message: "halted" });
   });
 
-  it("can halt itself between yield points", async () => {
+  it.skip("can halt itself between yield points", async () => {
```
jbolda, line 218: If we plan to fix this at some point, is it worth putting a TODO and/or linking to the PR description to describe the deadlock situation. That way we don't forget the context, and other folks could happen to trigger this case then search and find this test.

jbolda (approved):

test/run.test.ts
cowboyd: done!

PR #1083, opened 2026-01-20
Add Node/PNPM monorepo support for effectionx extensions

taras:
# Motivation

The effectionx repository was migrated from Deno to Node/PNPM monorepo structure. The website was hardcoded to read `deno.json` files for package metadata, which no longer works since effectionx now uses `package.json` and `pnpm-workspace.yaml`.

Additionally, effectionx packages are no longer published to JSR, so the JSR API calls were failing and causing inconsistent display on the `/x` pages.

# Approach

### New Monorepo Adapter
Created a new package/workspace abstraction that supports both Deno and Node/PNPM monorepos:

- **lib/workspaces/** - Auto-detects monorepo type by checking for `deno.json` vs `pnpm-workspace.yaml`
- **lib/package/** - Package representation for both Deno and Node projects, reading from respective manifest files
- **lib/registries/** - Registry abstractions for npm (JSR kept for potential future use)

Key implementation details:
- Parse Node `package.json` exports with conditional export support (uses `development` condition for source files)
- Filter hidden packages (e.g., `.internal`) from listings
- Sanitize semver OR ranges (e.g., `^3 || ^4`) for Deno npm: specifiers

### JSR Removal
- Removed `getJSRDetails()` calls - now uses local package manifest data for name/description
- Removed JSR badge from package headers (kept npm badge only)
- Deleted `ScoreCard` component (was entirely JSR-focused)
- Cleaned up unused JSR-specific types


cowboyd:
Why do we need a Deno abstraction?

taras:
We still use demo for effection and I didn't want to completely get rid of the deno integration in case we decide to use deno again. They might fix all of the problems tomorrow and who knows.

www/components/package/header.tsx
```diff
@@ -1,17 +1,22 @@
-import { Package } from "../../lib/package.ts";
+import type { Package } from "../../lib/package/types.ts";
 import { GithubPill } from "./source-link.tsx";
 
 export function* PackageHeader(pkg: Package) {
+  const name = yield* pkg.getName();
+  const version = yield* pkg.getVersion();
+  const scopeName = yield* pkg.getScopeName();
+  const shortName = name.includes("/") ? name.split("/")[1] : name;
```
cowboyd:
```suggestion
  let name = yield* pkg.getName();
  let version = yield* pkg.getVersion();
  let scopeName = yield* pkg.getScopeName();
  let shortName = name.includes("/") ? name.split("/")[1] : name;
```

cowboyd:
> We still use demo for effection and I didn't want to completely get rid of the deno integration in case we decide to use deno again. They might fix all of the problems tomorrow and who knows.

I'm of two minds here. I would love the ability to go back to Deno, but but until then, this is dead code sitting in our repo, right? If we keep the API, and just have one implementation, that should be enough. I'd rather let the api evolve with our needs according to what we need for the website, and not be burdened by having to keep one foot in each world.

www/lib/package/deno.ts
```diff
@@ -0,0 +1,197 @@
...
+    workspaceName,
+    workspacePath,
+    ref,
+    deno: true,
+    node: false,
+    registries,
+
+    // Registry URLs - will use the cached name once loaded
+    get npm() {
+      const name = cachedName ?? `@effectionx/${workspaceName}`;
```
cowboyd:
```suggestion
     let name = cachedName ?? `@effectionx/${workspaceName}`;
```

taras:
Sounds good

cowboyd (approved):
I'm going to go ahead and pre-approve this so that we can move forward. I'll leave whether or not to remove the Deno package api up to you because I can see both ways really. It shouldn't be a blocker

PR #1088, opened 2026-01-22
🚨 lint and format www/

cowboyd:
## Motivation
The www/ webapp was a bit of wild west because linting and formatting were turned off resulting in the style in the web app being all over the place.

## Approach
This turns it back on and fixes and formats the code. There were two files that were causing `deno lint` to crash, so they've been excluded.

- routes/guides-route.tsx
- components/api/api-page.tsx

We can follow up with the deno team to find out why these are crashing later.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,12 +3,16 @@
 
 # Explicitly set line endings for text files
 *.ts text eol=lf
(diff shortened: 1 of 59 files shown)


taras (approved):
👍

PR #1089, opened 2026-01-27
docs: add `@since` tags to exports

### LINKED ISSUE #1039: Mark every API method, in both v3 and v4 as having "since" tag
This makes the api docs useful

alexvyber:
Closes #1039 
Adds `@since` tags to exports

### DIFF
--- a/lib/scoped.ts
+++ b/lib/scoped.ts
@@ -24,6 +24,7 @@ import { createScopeInternal } from "./scope-internal.ts";
  * @param operation - the operation to be encapsulated
  *
  * @returns the scoped operation
+ * @since 3.2
  */
 export function scoped<T>(operation: () => Operation<T>): Operation<T> {
   return {
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -37,7 +37,7 @@ import type { Result } from "./result.ts";
  * ```
  *
  * See [Operations guide](https://frontside.com/effection/docs/operations) for more information.
- *
+ * @since 3.0
  */
 export interface Operation<T> {
   [Symbol.iterator](): Iterator<Effect<unknown>, T, unknown>;
@@ -51,6 +51,7 @@ export interface Operation<T> {
  * can also be evaluated directly within another operation, so among other
  * things, if the operation resolves synchronously, it will continue within the
  * same tick of the run loop.
+ * @since 3.0
  */
 export interface Future<T> extends Operation<T>, Promise<T> {}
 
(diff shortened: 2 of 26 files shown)


lib/scoped.ts
```diff
@@ -24,6 +24,7 @@ import { createScopeInternal } from "./scope-internal.ts";
  * @param operation - the operation to be encapsulated
  *
  * @returns the scoped operation
+ * @since effection-v3.0.0
```
cowboyd: This was released with version 3.2.0 so I think that the algorithm might be not quite right.

lib/do.ts
```diff
@@ -6,6 +6,7 @@ import type { Effect, Operation } from "./types.ts";
  *
  * @param effect - the effect to perform
  * @returns an operation that performs `effect`
+ * @since effection-v4.0.0
```
cowboyd: `Do()` is not part of the public api and should not be reachable via `lib/mod.ts`.

lib/types.ts
```diff
@@ -131,6 +132,7 @@ export interface Future<T> extends Operation<T>, Promise<T> {}
  * @see {@link run}
  * @see {@link spawn}
  * @see {@link Scope.run}
+ * @since effection-v3.0.0
```
cowboyd: I think the Task interface was actually introduce in 2.0

lib/types.ts
```diff
@@ -366,6 +375,7 @@ export interface Coroutine<T = unknown> {
 
 /**
  * @ignore
+ * @since effection-v3.0.0
  */
 export interface Resolve<T> {
```
cowboyd, line 377: We don't need `since` tags for ignored APIs since they are not part of the public api and should not be relied upon.

lib/until.ts
```diff
@@ -11,6 +11,7 @@ import type { Operation } from "./types.ts";
  * @template {T}
  * @param promise
  * @returns {Operation<T>} that succeeds or fails depending on the outcome of `promise`
+ * @since effection-v3.0.0
```
cowboyd: I believe this was introduced in v3.4.0, so again, we may need to revisit the logic.

cowboyd (commented):
Here are a couple of tweaks:

1. Let's only use the major minor version since those are the only meaningful semver points that can introduce or change an api. I.e. instead of `since @effection-v3.0.1`, just `@since 3.1`
2. I've noted it in a couple of places, but it looks as though the version is not correct. In fact, it appears that all of the versions are either 3.0 or 4.0 which means we're losing something somewhere. For example, the `scoped` API was introduced in `v3.2.0`
3. For interface that have methods like `Scope`, we need to include their methods. For example `Scope.run()` was introduced in `3.0`, but `Scope.spawn()` not until `3.5`

taras:
@alexvyber you've done all of the heavy lifting here. I don't want to rush you, but I do want to extend an hand in case you need help driving this over the finish line. Please let me know if there is anything I can do to help.

cowboyd (approved):

PR #1106, opened 2026-02-12
feat: add dynamic /llms.txt route for AI agent discovery

taras:
# Motivation

AI agents have almost zero knowledge (~10-15% confidence) of `@effectionx/*` packages. When tested, agents couldn't recommend packages like `@effectionx/process` for process spawning or `@effectionx/websocket` for WebSocket connections.

The [llms.txt standard](https://llmstxt.org/) provides a machine-readable format for helping AI agents discover and understand project capabilities.

# Approach

Add a dynamic `/llms.txt` route that generates a machine-readable index following the llmstxt.org standard:

- Uses `useWorkspaces("thefrontside/effectionx")` to clone/read the effectionx monorepo
- Enumerates all 23 packages using `all()` for concurrent resolution
- Extracts name and description from each package
- Truncates descriptions to first sentence for agent-friendly consumption
- Returns plain text Response with proper `Content-Type` header

Also updates the `<link rel="alternate">` in HTML head and footer link to point to `/llms.txt` instead of `/assets/llms.txt` (static file kept as fallback for cached URLs).

Pattern follows `blogFeedRoute` — `*handler()` returns `Response`, no `routemap` (keeps it out of sitemap.xml but staticalize will discover it via the `<link>` tag).

### DIFF
--- a/www/components/footer.tsx
+++ b/www/components/footer.tsx
@@ -35,7 +35,7 @@ export function Footer(): JSX.Element {
         <h4 class="text-sm uppercase font-bold text-blue-primary dark:text-blue-secondary mb-4">
           AI Agent Resources
         </h4>
-        <a href="/assets/llms.txt" class="text-gray-800 dark:text-gray-200">
+        <a href="/llms.txt" class="text-gray-800 dark:text-gray-200">
           llms.txt
         </a>
         <a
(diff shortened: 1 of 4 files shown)


cowboyd (commented):
This should be in the sitemap, no?

taras:
@cowboyd done

cowboyd (approved):

PR #1108, opened 2026-02-12
fix: handle signal-killed subprocess exit codes in tests

taras:
## Motivation
Deno 2.6.9 changed node:child_process signal semantics (denoland/deno#32081), and tinyexec now returns undefined exitCode for signal-killed processes.

## Approach
Normalize signal-killed process outputs in test/suite.ts by deriving conventional signal exit codes when exitCode is undefined.


taras:
Superseded by #1109, which is based directly on origin/v4 and includes only the test helper fix.

PR #1115, opened 2026-02-15
Add using helper for JS disposable resources

joshamaju:
# Motivation

Add an Effection using helper so JavaScript Disposable and AsyncDisposable values can be scope-owned without requiring the native using keyword.

# Approach

- Add lib/using.ts with a using operation that binds disposal to scope teardown.
- Validate disposers at runtime and fail fast for non-disposable values.
- Export using from lib/mod.ts.
- Add test/using.test.ts covering sync and async disposal, teardown on error and halt, async disposal completion, and invalid runtime input.

### DIFF
--- a/lib/using.ts
+++ b/lib/using.ts
@@ -1,0 +1,63 @@
+import { call } from "./call.ts";
+import { resource } from "./resource.ts";
+import type { Operation } from "./types.ts";
+
+type DisposableLike = {
+  [Symbol.dispose]?(): void;
+  [Symbol.asyncDispose]?(): PromiseLike<void> | void;
+};
+
+function getDisposer(value: DisposableLike): () => PromiseLike<void> | void {
+  let asyncDispose = value[Symbol.asyncDispose];
+
+  if (typeof asyncDispose === "function") {
+    return asyncDispose.bind(value);
+  }
+
+  let dispose = value[Symbol.dispose];
+
+  if (typeof dispose === "function") {
+    return dispose.bind(value);
+  }
+
+  throw new TypeError(
+    "using() value must implement Symbol.dispose or Symbol.asyncDispose",
+  );
+}
+
+/**
+ * Bind a JavaScript disposable value to the current Effection scope.
+ *
+ * The provided value is yielded immediately, then disposed when the owning
+ * scope exits (on return, error, or halt).
+ *
+ * @example
+ * ```ts
+ * import { run, using } from "effection";
+ *
+ * class Connection {
+ *   opened = true;
+ *   [Symbol.dispose]() {
+ *     this.opened = false;
+ *   }
+ * }
+ *
+ * await run(function* () {
+ *   let connection = yield* using(new Connection());
+ *   connection.opened; // true while in scope
+ * });
+ * ```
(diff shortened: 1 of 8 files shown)


lib/using.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  if (typeof dispose === "function") {
+    return dispose.bind(value);
+  }
+
+  throw new TypeError(
+    "using() value must implement Symbol.dispose or Symbol.asyncDispose",
+  );
+}
+
```
cowboyd: Let's just inline all of this into the `finally`. I don't think that the getDisposer adds any clarity since it isn't re-used.

test/using.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  isDisposed = false;
+  disposeStarted = false;
+
+  async [Symbol.asyncDispose]() {
+    this.disposeStarted = true;
+    await new Promise<void>((resolve) => setTimeout(resolve, 20));
+    this.isDisposed = true;
+  }
+}
+
```
cowboyd: Let's move these to the bottom of the file. That way the test is the most salient aspect of the test file.

lib/using.ts
```diff
@@ -51,13 +28,23 @@ function getDisposer(value: DisposableLike): () => PromiseLike<void> | void {
...
+      "using() value must implement Symbol.dispose or Symbol.asyncDispose",
+    );
+  }
 
   return yield* resource<T>(function* (provide) {
     try {
       yield* provide(value);
     } finally {
-      yield* call(() => disposer());
+      yield* call(() => disposer.bind(value)());
```
cowboyd: let's just use a `dispose.call(value)` here. Simpler, and creates one less wrapper function

lib/using.ts
```diff
@@ -51,13 +28,23 @@ function getDisposer(value: DisposableLike): () => PromiseLike<void> | void {
 export function* using<T extends Disposable | AsyncDisposable>(
   value: T,
 ): Operation<T> {
-  let disposer = getDisposer(value);
+  let disposer = Symbol.asyncDispose in value
```
cowboyd: let's call this `dispose` since it is just a function

cowboyd (commented):
Looks good. We don't have to do this with this PR, but we need a place where we are going to explain this. Maybe in the resources section of the guide?

cowboyd (approved):

joshamaju:
@cowboyd Please check the tests

taras:
@joshamaju can you rebase and verify the signatures

joshamaju:
Done

cowboyd:
@joshamaju I'm still seeing some un-verified commits as well as a lot of churn.

joshamaju:
> @joshamaju I'm still seeing some un-verified commits as well as a lot of churn.

Right, I'll look into that

cowboyd:
Looks almost ready, but there is still a change to a vector graphic in one of the blog posts

joshamaju:
Is that for me to fix? I only formatted the svgs for the CI to pass.

cowboyd (approved):

PR #1116, opened 2026-02-16
feat: export createApi from main module

taras:
## Summary

Export the `createApi` function from the main effection module so users can create APIs with middleware support without importing from `experimental.ts`.

## Changes

- Add comprehensive JSDoc documentation with examples for `createApi`
- Export `createApi` from `lib/mod.ts`
- Update test to import `createApi` from main module instead of experimental

## Why

The `createApi` function enables:
- **Dependency injection**: Swap implementations based on context
- **Test mocking**: Replace real operations with test doubles
- **Instrumentation**: Add logging, metrics, or tracing middleware

This is now stable enough to be part of the public API.

## Usage

```ts
import { createApi, run } from "effection";

const myApi = createApi("myApi", {
  *fetch(url: string) {
    return yield* until(globalThis.fetch(url));
  }
});

// Use the API
await run(function*() {
  let response = yield* myApi.operations.fetch("/api/data");
});

// With middleware for testing
await run(function*() {
  yield* myApi.around({
    *fetch() {
      return { ok: true, json: () => ({ mocked: true }) };
    }
  });
  // All fetch calls now return mocked data
});
```

## Notes

Once this PR is merged and the preview package is published, we can update the `@effectionx/*` packages to use `createApi`.

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -1,8 +1,73 @@
-// deno-lint-ignore-file ban-types
-import type { Api, Context, Operation, Scope } from "./types.ts";
 import { createApiInternal } from "./api-internal.ts";
 import type { ScopeInternal } from "./scope-internal.ts";
+// deno-lint-ignore-file ban-types
+import type { Api, Context, Operation, Scope } from "./types.ts";
 
+/**
+ * Create an API object that provides middleware support for a set of operations.
+ *
+ * APIs enable context-sensitive functionality such as dependency injection,
+ * test mocking, and instrumentation by allowing middleware to be applied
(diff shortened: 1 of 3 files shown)


cowboyd:
I don't think we're there yet. We haven't had any releases with it exported from main, and while there are less questions lingering, there are still some. 

the biggest questions to resolve are:

- do we need to change the api anymore?
- how will it effect performance?

At the very least, we should have perf regression testing in place before we move this from experimental to 100% supported.

taras:
Oh sorry, is it exported from experimental?

🤦 that makes more sense

PR #1120, opened 2026-02-19
feat(www): add OG image generation for blog posts

taras:
# Motivation

Social media platforms (Twitter, LinkedIn, Facebook, etc.) don't support SVG images in Open Graph `og:image` meta tags. Our blog posts use SVG featured images that render beautifully on the blog but appear broken or missing when shared on social media.

This PR adds build-time PNG generation from SVG featured images so social previews work correctly.

# Approach

- **PNG generation script**: Uses Resvg WASM to render SVGs to 1200×630 PNGs (standard OG dimensions)
- **Effection structured concurrency**: Uses bounded parallelism (4 concurrent) for efficient batch processing
- **Animation stripping**: Removes CSS animations from animated SVGs to show "final state" where all content is visible
- **Light mode forcing**: Uses cascade override injection (`!important` rules at end of SVG) to ensure consistent light mode colors regardless of system theme
- **TTF fonts committed**: Inter and JetBrains Mono (OFL licensed) since Resvg WASM doesn't support WOFF2

## Build & Deploy Flow

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                            GitHub Actions CI                                 │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│  1. Generate OG Images                                                      │
│     ┌──────────────────┐                                                    │
│     │ deno task        │     ┌─────────┐      ┌─────────┐                   │
│     │ generate-og-     │────▶│  SVG    │─────▶│  PNG    │                   │
│     │ images           │     │ (blog)  │      │ (1200×  │                   │
│     └──────────────────┘     └─────────┘      │  630)   │                   │
│            │                                   └─────────┘                   │
│            │ Resvg WASM                             │                        │
│            │ • Strip animations                     │                        │
│            │ • Force light mode                     │                        │
│            │ • Embed TTF fonts                      │                        │
│            ▼                                        ▼                        │
│  2. Start Dev Server                                                        │
│     ┌──────────────────┐                                                    │
│     │ deno task dev    │  serves both SVG and PNG via assetsRoute("blog")   │
│     │ localhost:8000   │                                                    │
│     └──────────────────┘                                                    │
│            │                                                                │
│            ▼                                                                │
│  3. Staticalize                                                             │
│     ┌──────────────────┐     ┌─────────────────────────────────────────┐    │
│     │ staticalize      │────▶│ Crawls sitemap.xml                      │    │
│     │ --site localhost │     │ For each page:                          │    │
│     │ --output built/  │     │   • Downloads HTML                      │    │
│     │ --base effection │     │   • Parses [content] attributes         │    │
│     │   .deno.dev      │     │   • Finds og:image meta tag URL         │    │
│     └──────────────────┘     │   • Downloads referenced PNG ◀────────┐ │    │
│            │                 │   • Rewrites URL to production base   │ │    │
│            │                 └───────────────────────────────────────┼─┘    │
│            │                                                         │      │
│            │                 <meta property="og:image"               │      │
│            │                   content="http://localhost:8000/       │      │
│            │                     blog/.../image.png">────────────────┘      │
│            ▼                                                                │
│  4. Deploy to Deno Deploy                                                   │
│     ┌──────────────────┐                                                    │
│     │ built/           │  Contains HTML + PNG files                         │
│     │ ├── blog/        │  og:image URLs rewritten to                        │
│     │ │   └── post/    │  https://effection.deno.dev/blog/.../image.png     │
│     │ │       └── .png │                                                    │
│     │ └── ...          │                                                    │
│     └──────────────────┘                                                    │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

## Result

When a blog post is shared on social media, the platform fetches the `og:image` URL and displays the PNG thumbnail instead of showing a broken image or generic fallback.


cowboyd (commented):
I'd try lazy generation first. It will be a less invasive change, and it means the website will work the same in local context as everywhere (no need to have a separate task at build time)

It also means that if generation isn't working, it will be caught during dev time before CI even runs

cowboyd:
If there is principle here to extract, it is that what makes staticalize work is that the there is no build, only capture. The server _is_ the build.

www/plugins/og-image.ts
```diff
@@ -0,0 +1,202 @@
...
+  options: OgImageOptions,
+): Operation<RevolutionPlugin> {
+  let env = yield* initializeRenderer(options.fontsDir);
+  let cache = yield* until(caches.open("og-images"));
+
+  return {
+    http(request, next) {
+      let url = new URL(request.url);
+
+      if (url.pathname.startsWith("/blog/") && url.pathname.endsWith(".png")) {
```
taras: this seems oddly specific. Would it be better to detect if a file with svg name exists instead?

www/plugins/og-image.ts
```diff
@@ -0,0 +1,202 @@
...
+  "svg-label-heading": "#0b2a5b",
+};
+
+function transformSvg(svg: string): string {
+  svg = replaceFontFamilies(svg);
+  svg = fixCssFills(svg);
+  svg = stripAnimations(svg);
+  svg = forceLightMode(svg);
+  return svg;
+}
```
taras: should we use a pipe here?

www/plugins/og-image.ts
```diff
@@ -0,0 +1,202 @@
...
+  svg = svg.replace(/animation-[a-z-]+:\s*[^;]+;/g, "");
+
+  let overrideCSS = `
+    [class*="svg-anim-"] { opacity: 1 !important; }
+    .svg-cursor { opacity: 0 !important; }
+  `;
+  return svg.replace(/<\/style>/i, `${overrideCSS}</style>`);
+}
+
+function forceLightMode(svg: string): string {
```
taras: Add a comment about what this function does

www/plugins/og-image.ts
```diff
@@ -0,0 +1,217 @@
...
+        "Content-Type": "image/png",
+        "Cache-Control": "public, max-age=604800",
+      },
+    });
+
+    yield* until(cache.put(request, response.clone()));
+    return response;
+  } catch (error) {
+    console.error(`OG image generation failed for ${pathname}:`, error);
+    return yield* serveFallback(request, options.basedir);
```
cowboyd, line 107: I'm not sure if this should be recoverable. It means we _found_ an svg that should be rendered off of the png but that it failed, which means that there is a bug in our site.

www/routes/blog-post-route.tsx
```diff
@@ -44,6 +44,7 @@ export function blogPostRoute({
       let AppHtml = yield* useAppHtml({
         title: `${post.title} | Blog | Effection`,
         description: post.description,
+        ogImage: post.ogImage ? `/blog/${id}/${post.ogImage}` : undefined,
```
cowboyd, line 47: this is where the fallback should go, not a const. This is detecting that there is actually no image associated with the blog post.

www/resources/blog.ts
```diff
@@ -130,6 +133,17 @@ function* loadBlog(): Operation<Blog> {
...
+      // Compute OG image path: if SVG exists, the PNG will be generated on-demand
+      // Note: blog-post-route.tsx prepends /blog/{id}/ to this value
+      let ogImage: string | undefined;
+      if (frontmatter.image?.endsWith(".svg")) {
+        let svgFullPath = `${directory}${id}/${frontmatter.image}`;
+        if (existsSync(svgFullPath)) {
+          ogImage = frontmatter.image.replace(/\.svg$/, ".png");
+        }
+      }
+
```
cowboyd, line 146: This bleeds the plugin logic into the blog post, while at the same time duplicating it. not good.

The plugin should transform the html and do the check there. You could probably generate it at that point to, so the plugin is not an *http() plugin, but an *html().

cowboyd (changes requested):

cowboyd:
This has been implemented.

PR #1121, opened 2026-02-19
🐛 remove banned triple slash

cowboyd:
# Motivation

jsr.io does not allow triple slashes. 

# Approach

This remove it since it can be set by the consuming library. This also adds validation checks after tests have successfully run to see if publication would succeed on both npm and jsr.

image

This also adds these same pre-flight checks to the actual publication workflow. It ensures that _both_ JSR _and_ NPM will work before attempting _either_. That way we don't end up in a situation where we have the package on one registry and not another.

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -40,6 +40,45 @@ jobs:
         if: runner.os != 'Windows'
         run: deno task test
 
+  jsr:
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: setup deno
+        uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+
+      - name: dry run publish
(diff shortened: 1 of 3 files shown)


taras (approved):

.github/workflows/verify.yaml
```diff
@@ -40,6 +40,45 @@ jobs:
...
+        with:
+          deno-version: v2.x
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 24
+
+      - name: Build NPM
+        run: deno task build:npm 4.0.0-verify.0
```
taras, line 76: Why is there a hardcoded version here?

taras, line 76: Oh, i see because you need a version number. It might be worth adding a comment here to explain.

cowboyd: Yeah, this is just a total dummy number. It could be 12.1000.675-verify.0

PR #1124, opened 2026-02-19
🤖 Add gitmoji directive to AGENTS.md

cowboyd:
## Motivation

gitmoji is faster to parse with the eye, less stodgy than standard commits, but equally (if not more) expressive.

## Approach

Add a "Commit and PR conventions" section to AGENTS.md to teach agents how to use gitmoji


taras (approved):
👍

PR #1127, opened 2026-02-25
✨ Export reducer internals via experimental and label infrastructure effects

taras:
# Motivation

External packages (like `@effectionx/durable`) need to build custom reducers that intercept effect processing — for example, to record effect resolutions to a durable stream and replay them on resume. Currently, the types and contexts required to build a custom reducer (`ReducerContext`, `Instruction`, `InstructionQueue`, `DelimiterContext`) are internal to Effection and not accessible through any public entrypoint.

Additionally, custom reducers need to distinguish **infrastructure effects** (like `useCoroutine()`, `useScope()`, scope destruction) from **user-facing effects** (like `sleep()`, `call()`, `action()`). Infrastructure effects that use `withResolvers()` currently have no description label, making them indistinguishable from user code.

This PR makes durable execution (and other custom reducer patterns) **completely additive** to Effection — no fork required.

# Approach

**1. Export reducer internals via `experimental.ts`** (7 lines added)

Re-export `Reducer`, `ReducerContext`, `InstructionQueue`, and the `Instruction` type from `lib/reducer.ts`, plus `DelimiterContext` from `lib/delimiter.ts`, through the existing `experimental` entry point. This follows the established pattern — `experimental.ts` already exports `api` for the same reason (enabling external middleware).

**2. Add `export` to `Instruction` and `InstructionQueue`** (2 lines changed in `lib/reducer.ts`)

These were module-private. Making them `export` allows the re-export in `experimental.ts`.

**3. Label infrastructure `withResolvers()` calls** (5 files, 1 line each)

Add description strings to `withResolvers()` calls in core infrastructure:
- `lib/callcc.ts`: `"await callcc"`
- `lib/delimiter.ts`: `"await delimiter"`
- `lib/each.ts`: `"await each done"`, `"await each context"`
- `lib/future.ts`: `"await future"`
- `lib/scope-internal.ts`: `"await destruction"`

The `description` parameter already exists on `withResolvers()` and is already used elsewhere. These labels are purely additive — they make infrastructure effects identifiable by description, which is useful for debugging, inspection, and custom reducers.

**Zero behavior changes.** All 121 test steps across 15 core test suites pass unchanged.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -74,7 +74,7 @@ jobs:
           path: ./build/npm
 
       - name: Publish NPM
-        run: npm publish --access=public --tag=latest
+        run: npm publish --access=public --tag=next
         working-directory: ./build/npm
 
   publish-jsr:
--- a/deno.json
+++ b/deno.json
@@ -2,7 +2,7 @@
   "name": "@effection/effection",
   "exports": "./mod.ts",
   "license": "ISC",
-  "publish": { "include": ["lib", "mod.ts", "README.md"] },
+  "publish": { "include": ["lib", "mod.ts", "experimental.ts", "README.md"] },
   "lock": false,
   "tasks": {
     "test": "deno test --allow-run --allow-read --allow-env --allow-ffi",
@@ -28,6 +28,10 @@
     "ctrlc-windows": "npm:ctrlc-windows@2.2.0",
     "tinyexec": "npm:tinyexec@1.0.1",
     "https://deno.land/x/path_to_regexp@v6.2.1/index.ts": "npm:path-to-regexp@8.2.0"
+  },
+  "exports": {
+    ".": "./mod.ts",
+    "./experimental": "./experimental.ts"
   },
   "nodeModulesDir": "auto",
   "workspace": [
--- a/deno.json
+++ b/deno.json
@@ -12,7 +12,7 @@
     "build:jsr": "deno run -A tasks/build-jsr.ts"
   },
   "lint": {
-    "rules": { "exclude": ["prefer-const", "require-yield"] },
+    "rules": { "exclude": ["prefer-const", "require-yield", "ban-types"] },
     "exclude": ["build", "docs", "tasks"],
     "plugins": ["lint/prefer-let.ts"]
   },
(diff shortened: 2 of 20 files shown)


cowboyd:
This should be solved with and API/middleware.

taras:
Can you say more? Do you expect this to already be solvable with middleware? or is something missing that would make it possible with middleware?

cowboyd:
Extensions to the core should always be done with an `Api` like `ScopeApi` or `MainApi`. Maybe something like `ReducerApi`?

taras:
@cowboyd we already have a reducer context and the reducer is an interface so that makes sense. we'll need to see what the performance implications are.

jackyzha0:
hey! just wanted to express interest in this landing :) +1 this seems like a useful addition and it's not something we can easily do via patching as things break transitively

taras:
@jackyzha0 I haven't pushed this forward because I was able to do everything I wanted to do without it. 

I point to some of my solutions if it aligns with your use case. Can you tell me how you want to use this?

dcolascione:
I've also been working on runtime tricky that needs access to the Reducer object. This work would be quite useful and would save us from having to commit import crimes. I mean, there are all sorts of workarounds for not having this export, ranging from ugly to Lovecraftian, but since this is a useful extension point and pretty (IMHO) feasible to support with forward compatibility, the best option is to land it after all.

jackyzha0:
@taras i'd be curious what you came up with! we basically needed access to Priority, ReducerContext, and ErrorContext, and the Reducer class

the context ones we were able to fish out by taking advantage of the stringly-keyed context slots but got a bit stick on Reducer and ended up doing some import patching hackery with `createRequire` that makes me sad

taras:
I was using this branch for https://github.com/thefrontside/effectionx/pull/179 but I was able to simplify my implementation not to need these. However, we do intend to expose some of the Effection internals via middleware in [4.1](https://github.com/thefrontside/effection/blob/v4-1-alpha/lib/api.ts#L24). 

We can do a write up about what's coming to see if it'll find both of your use cases.

cowboyd:
@jackyzha0 @dcolascione +1 on hearing your use-cases. We definitely are all about exposing the internals of Effection. We just want to make sure we do it in the right way since it means moving more things into the public API, and we wanted to craft it in just the right shape.

cowboyd:
We should be using middleware from the upcoming release for this.

PR #1133, opened 2026-03-01
📝 Add Library Integration guide for Promise-based API pattern

taras:
# Motivation

A community member asked in Discord: "Do you have any recommendations for using Effection inside a library, while exposing a more traditional API to the clients?"

This is a common use case that deserves dedicated documentation. The pattern evolved through discussion and is now captured in a [gist](https://gist.github.com/taras/a8e3861d34d6005781069fdfb16de475), but should live in the official docs.

# Approach

Add a new "Library Integration" guide to the Advanced section covering the pattern for using Effection internally while exposing a Promise-based API:

- **`createScope()`** as the integration point
- **`withResolvers()`** to synchronize initialization and prevent race conditions
- **`resource()`** for lifecycle management
- **`suspend()`** to keep the scope alive between calls
- **`scope.run()`** as the bridge to Promises
- **`Symbol.asyncDispose`** for `await using` support

Changes:
- Create `docs/library-integration.mdx` with the full pattern and connection pool example
- Add to "Advanced" section in `docs/structure.json`
- Add cross-reference from `docs/scope.mdx` (in the "Creating a brand new Scope" section)
- Add FAQ entry in `docs/faq.mdx` pointing to the guide

The guide includes gotchas identified during review:
- Guarding against post-shutdown calls (track a `closed` flag)
- Awaiting `destroy()` for cleanup completion
- In-flight work behavior during shutdown
- Startup failure propagation

### DIFF
--- a/docs/library-integration.mdx
+++ b/docs/library-integration.mdx
@@ -1,0 +1,302 @@
+You build a library with a Promise API. Your users call `await client.query()`
+and move on. Internally, though, you need stronger lifecycle guarantees than
+promises can give you. You need requests, sockets, subscriptions, and cleanup
+tied to one owner. You need work to stop when the library shuts down.
+
+This is where Effection fits behind a Promise API.
+
+## The Integration Boundary
+
+The core of the pattern is a scope that lives for the lifetime of your library.
+You create it once, initialize your stateful resources inside it, and keep it
+alive until shutdown. `withResolvers()` gives you a readiness gate that
+prevents race conditions between "pool is ready" and "first query arrives."
+
+```typescript
+import { createScope, suspend, withResolvers } from "effection";
+
+const [scope, destroy] = createScope();
+const { operation: poolReady, resolve, reject } = withResolvers<PoolState>();
+
+let closed = false;
+
+scope.run(function* () {
+  try {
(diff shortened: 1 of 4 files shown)


docs/library-integration.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+Expose plain methods that call `scope.run(() => operation)`. This returns a
+promise, so clients stay in normal `async/await`, while your internal work
+stays under structured concurrency.
+
+```typescript
+function* executeQuery(sql: string): Operation<QueryResult> {
+  if (closed) {
+    throw new Error("Pool is shut down");
+  }
```
cowboyd, line 78: Code like this should never appear inside an `Operation` since this is the point of structured concurrency. The lifetime of code does not need to be track apart from its own execution. If this check is necessary, it should happen at the promise layer.

cowboyd:
I had another read. I think it needs a tighter narrative arc. I wasn't able to follow the individual pieces, so maybe start with a:

this is the high level shape. You have the Library artifact lifecycle you init it, you use it, you shut it down. You can then show this lifecycle (which s currently at the end), and _then_ show how you implement your initializing and shut down, and then how you handle each individual operation.

PR #1135, opened 2026-03-02
🤖 Require braces on if statements

cowboyd:
## Motivation

Bare/braceless `if` statements reduce readability and are prone to bugs when
additional statements are added later. The [Heartbleed vulnerability](https://dwheeler.com/essays/heartbleed.html#goto-fail) in Apple's SSL implementation (goto fail) is a famous example of how a missing brace led to a critical security flaw.

## Approach

Added a "Code style" section to `AGENTS.md` with the rule.


taras (approved):
👍

PR #1140, opened 2026-03-16
✍️ Reframe strict structured concurrency around lifetime correctness

taras:
# Motivation

The strict structured concurrency post already introduced foreground and background work, but it still read primarily as a convenience refinement. This follow-up makes the stronger claim explicit: when incidental background work can keep a scope alive after the meaningful computation is complete, the lifetime model is semantically wrong and every framework built on top of it inherits an incorrectness tax.

# Approach

Rewrite the article to lead with the correctness argument, introduce foreground vs background earlier, and recast the spinner/timeout example as a semantic and operational failure mode rather than cleanup boilerplate. Preserve the existing points about aggressive teardown, classic structured concurrency, and orderly cleanup while tightening the argument around lifetime correctness, hung scopes, and the operational cost of letting support work outlive the computation it was supposed to support.


cowboyd:
This feedback has been integrated.

PR #1141, opened 2026-03-18
fill out JSDocs with examples

### LINKED ISSUE #1139: `race()` docs out of date
The `fetch` export no longer exists.

See the JSDocs on `race`.
https://github.com/thefrontside/effection/blob/effection-v4.0.2/lib/race.ts#L18-L25

jbolda:
# Motivation

Filling out the JSDocs a bit more. These show up both on the website, but also in IDEs.

Closes #1139.

# Approach

Fix a few broken bits, and add more examples.

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -81,7 +81,7 @@ export const api: Apis = {
     },
     *destroy() {},
     set(scope, context, value) {
-      return ((scope as ScopeInternal).contexts[context.name] = value);
+      return (scope as ScopeInternal).contexts[context.name] = value;
     },
     delete(scope, context) {
       return delete (scope as ScopeInternal).contexts[context.name];
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -3,6 +3,33 @@ import type { Api, Context, Operation, Scope } from "./types.ts";
 import { createApiInternal } from "./api-internal.ts";
 import type { ScopeInternal } from "./scope-internal.ts";
 
+/**
+ * Create a new {@link Api}. This is the constructor behind
+ * middleware decoration used through core such as with {@link Scope#around}.
+ * One may implement an API for any operation to
+ *
+ * @example
+ * ```ts
+ * import { createApi, type Operation } from "effection/experimental";
+ *
+ * interface DatabaseApi {
+ *   query(sql: string): Operation<{ id: number; title: string }[]>;
+ * }
+ *
+ * // pass in the generic type or allow it to infer
+ * let Database = createApi<DatabaseApi>("database", {
+ *   *query(sql) {
+ *     console.log("running", sql);
+ *     return [];
+ *   },
+ * });
+ * ```
+ *
+ * @param name - the API identifier used in error and debug messages
+ * @param core - the core function/value implementations for this API
+ * @returns an {@link Api} that can be invoked and decorated per-scope
+ * @since 4.1
+ */
 export function createApi<A extends {}>(name: string, core: A): Api<A> {
   return createApiInternal(name, core);
 }
@@ -23,6 +50,30 @@ interface Apis {
   Main: Api<MainApi>;
 }
 
+/**
+ * Built-in internal APIs used by Effection's runtime and host integration.
+ * Advanced integrations can decorate these APIs in a scope.
+ *
+ * @example
+ * ```ts
+ * import { run, useScope } from "effection";
+ * import { api } from "effection/experimental";
+ *
+ * await run(function* () {
+ *   let scope = yield* useScope();
+ *
+ *   // Observe every scope creation in this scope subtree
+ *   scope.around(api.Scope, {
+ *     create(args, next) {
+ *       console.log("creating scope");
+ *       return next(...args);
+ *     },
+ *   });
+ * });
+ * ```
+ *
+ * @since 4.1
+ */
 export const api: Apis = {
   Scope: createApi<ScopeApi>("Scope", {
     create() {
@@ -30,7 +81,7 @@ export const api: Apis = {
     },
     *destroy() {},
     set(scope, context, value) {
-      return (scope as ScopeInternal).contexts[context.name] = value;
+      return ((scope as ScopeInternal).contexts[context.name] = value);
     },
     delete(scope, context) {
       return delete (scope as ScopeInternal).contexts[context.name];
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -6,7 +6,7 @@ import type { ScopeInternal } from "./scope-internal.ts";
 /**
  * Create a new {@link Api}. This is the constructor behind
  * middleware decoration used through core such as with {@link Scope#around}.
- * One may implement an API for any operation to
+ * One may implement an API around any operation or value and then decorate it per-scope.
  *
  * @example
  * ```ts
--- a/lib/context.ts
+++ b/lib/context.ts
@@ -5,6 +5,41 @@ import { Do } from "./do.ts";
 /**
  * Create a new {@link Context}
  *
+ * @example
+ * ```ts
+ * import { createContext, main } from "effection";
+ *
+ * let YourContext = createContext<string>("context-id");
+ *
+ * await main(function* () {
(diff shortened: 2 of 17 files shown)


lib/api.ts
```diff
@@ -3,6 +3,33 @@ import type { Api, Context, Operation, Scope } from "./types.ts";
 import { createApiInternal } from "./api-internal.ts";
 import type { ScopeInternal } from "./scope-internal.ts";
 
+/**
+ * Create a new {@link Api}. This is the constructor behind
+ * middleware decoration used through core such as with {@link Scope#around}.
+ * One may implement an API for any operation to
```
taras: This sentence currently reads as truncated, so the constructor's purpose is less clear than the example below it.

```suggestion
 * One may implement an API around any operation or value and then decorate it per-scope.
```

lib/context.ts
```diff
@@ -5,6 +5,41 @@ import { Do } from "./do.ts";
...
+ *   // where the arbitrary operations need access to the context value.
+ *   for (let operation of operations) {
+ *     yield* scope.run(function*() {
+ *       const value = yield* YourContext.expect(); // "abc-123"
+ *       // or set a different value for children of this scope
+ *       YourContext.set("def-456");
+ *     });
+ *   }
+ * });
+ * ```
```
taras: This second example is hard to follow as written because it uses `useScope()` without importing it, references `operations` without defining it, and calls `YourContext.set()` without `yield*`. I think a smaller, self-contained example would make the distinction between `scope.set()` and `Context.expect()` much clearer.

```suggestion
 * @example
 * ```ts
 * import { createContext, main, useScope } from "effection";
 *
 * let YourContext = createContext<string>("context-id");
 *
 * await main(function* () {
 *   let scope = yield* useScope();
 *   scope.set(YourContext, "abc-123");
 *
 *   console.log(scope.get(YourContext)); // "abc-123"
 *   console.log(yield* YourContext.expect()); // "abc-123"
 *
 *   yield* YourContext.set("def-456");
 *   console.log(yield* YourContext.expect()); // "def-456"
 * });
 * ```
```

lib/types.ts
```diff
@@ -356,6 +391,34 @@ export interface Scope {
  * sensitivity such as dependency injection, test mocking, and
  * instrumentation.
  *
+ * @example
+ * ```ts
+ * import { createApi, type Operation } from "effection";
```
taras: This example imports `createApi` from the main entrypoint, but in this PR it is exported from `effection/experimental`. Adjusting the import will keep the example consistent with the actual public surface.

```suggestion
 * import { type Operation } from "effection";
 * import { createApi } from "effection/experimental";
```

lib/types.ts
```diff
@@ -398,9 +461,24 @@ export interface Api<A> {
...
+ *
+ * function* DbMiddleware<[string], Operation<{ id: number; title: string }[]>>(args, next) {
+ *   let [sql] = args;
+ *   let start = Date.now();
+ *   try {
+ *     return yield* next(...args);
+ *   } finally {
+ *     console.log("query ms", Date.now() - start, sql);
+ *   }
+ * };
```
taras, line 480: This example currently isn't valid TypeScript, so it's likely to confuse readers trying to understand the `Middleware` shape. Suggest expressing it as a typed value instead.

```suggestion
 * import type { Middleware, Operation } from "effection";
 *
 * let dbMiddleware: Middleware<
 *   [string],
 *   Operation<{ id: number; title: string }[]>
 * > = function* (args, next) {
 *   let [sql] = args;
 *   let start = Date.now();
 *   try {
 *     return yield* next(...args);
 *   } finally {
 *     console.log("query ms", Date.now() - start, sql);
 *   }
 * };
```

lib/resource.ts
```diff
@@ -79,12 +79,29 @@ export function resource<T>(
...
+ *
+ * @example
+ * ```ts
+ * function useValue() {
+ *   return resource(function* (provide: Provide<number>) {
+ *     yield* provide(42);
+ *   });
+ * }
+ *
+ * console.log(yield* useValue); // 42
```
taras: This example forgets to call `useValue()`, which makes the consumption pattern look different from how `resource()` is actually used.

```suggestion
 * console.log(yield* useValue()); // 42
```

lib/types.ts
jbolda: Ah, I converted it from the `let` to a `function*` and forgot to adjust the type. I think we prefer `function*` syntax though?

lib/api.ts
```diff
@@ -23,14 +50,38 @@ interface Apis {
...
+ */
 export const api: Apis = {
   Scope: createApi<ScopeApi>("Scope", {
     create() {
       throw new TypeError(`no handler for Scope.create()`);
     },
     *destroy() {},
     set(scope, context, value) {
-      return (scope as ScopeInternal).contexts[context.name] = value;
+      return ((scope as ScopeInternal).contexts[context.name] = value);
```
cowboyd: what are the extra parens for?

lib/context.ts
```diff
@@ -5,6 +5,22 @@ import { Do } from "./do.ts";
 /**
  * Create a new {@link Context}
  *
+ * @example
+ * ```ts
+ * import { createContext, main } from "effection";
+ *
+ * let YourContext = createContext<string>("context-id");
```
cowboyd: Let's use `const` here since this is a true constant

lib/race.ts
```diff
@@ -16,12 +16,21 @@ import { Err, Ok, type Result } from "./result.ts";
...
+ *     (function* () {
+ *       yield* sleep(100);
+ *       // this is halted as it loses the race
+ *       return "slow";
+ *     })(),
+ *     (function* () {
+ *       yield* sleep(10);
+ *       return "fast";
+ *     })(),
+ *   ]);
```
cowboyd: instead of IIFE's let's use call()

lib/resource.ts
```diff
@@ -79,12 +79,29 @@ export function resource<T>(
...
+ * }
+ *
+ * console.log(yield* useValue()); // 42
+ *
+ * // alternatively, the preferred pattern is
+ * function useValue() {
+ *   return resource<number>(function* (provide) {
+ *     yield* provide(42);
+ *   });
+ * }
```
cowboyd, line 104: This example doesn't provide much.  A resource is about acquisition and release and so any example should demonstrate those two things. A connection pool maybe?

lib/scope.ts
```diff
@@ -4,6 +4,15 @@ import { createScopeInternal } from "./scope-internal.ts";
...
  * The root of all Effection Scopes.
+ *
+ * @example
+ * ```ts
+ * import { global, sleep } from "effection";
+ *
+ * await global.run(function* () {
+ *   yield* sleep(10);
+ * });
+ * ```
```
cowboyd: Do we need an example here? I don't think this adds anything beyond what is already in Scope

cowboyd (commented):
There are some where I'm not so sure we need examples. Or if we do have them, they should be very, very concise.

lib/scope.ts
jbolda: Agreed, must have accidentally staged when I meant to remove.

lib/api.ts
jbolda: Oops, prettier auto-format must have snuck in.

lib/resource.ts
jbolda: I was trying not to complicate it too much, but I see the point. The main intent was for folks that land on this API page trying to understand what the value of `provide` is.
https://frontside.com/effection/api/v4/Provide/

Is it valid to have an expectation that they arrive there having already looked at the value of using a resource? Maybe just a simple comment or something saying as such?

cowboyd: I think they will most likely arrive by hovering on a parameter `resource()` so could definitely be like "what is this?"

jbolda: I updated it. What do you think of this?

jbolda: @cowboyd updated per our conversation.

cowboyd (approved):

PR #1142, opened 2026-03-22
Fix /x/worker/ page broken by generic type params in MDX

taras:
## Summary
- The `/x/worker/` page was failing with "Failed to load worker due to error" because MDX interpreted generic type references like `Result<T>` in JSDoc comments as JSX tags
- Added a regex in `useMarkdown` to escape uppercase-starting angle bracket patterns (e.g. `<T>`, `<TSend, TRecv>`) to HTML entities before MDX processing
- Only affects TypeScript-style generics (uppercase start), not actual HTML tags (lowercase)

## Test plan
- [x] Verified `/x/worker/` renders correctly with README + API reference (155KB vs 13KB error page)
- [x] Verified `/x/fetch/`, `/x/websocket/`, `/x/process/` still render correctly
- [x] No server errors in console

### DIFF
--- a/www/hooks/use-markdown.tsx
+++ b/www/hooks/use-markdown.tsx
@@ -49,6 +49,14 @@ export function* useMarkdown(
   );
   let sanitized = yield* sanitize(markdown);
 
+  // Escape generic type parameters like <T>, <TSend, TRecv> that MDX
+  // would interpret as JSX tags. Only matches uppercase-starting identifiers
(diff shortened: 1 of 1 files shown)


jbolda (approved):

cowboyd (commented):
Looks good to me. One thing though: Can you ask Claude not to market itself by adding itself to commits and PRs?

PR #1149, opened 2026-04-09
✨ expose Result<T> as public API

rauhryan:
## Motivation

`Result<T>`, `Ok()`, and `Err()` are already exported from the public module surface, but they are still marked `@ignore` in the API docs.

This PR makes those APIs explicitly public so follow-on work can reference `Result<T>` without relying on undocumented types.

## Approach

- remove `@ignore` from `Result<T>`, `Ok()`, and `Err()`
- add public API documentation and examples for each
- leave `unbox()` internal

This is the base PR for the stacked `allSettled()` work.

### DIFF
--- a/lib/action.ts
+++ b/lib/action.ts
@@ -63,7 +63,7 @@ export function action<T>(executor: Executor<T>, desc?: string): Operation<T> {
               discard();
               discarded(Ok());
             } catch (error) {
-              discarded(Err(error as Error));
+              discarded(Err(error));
             }
           };
         },
--- a/lib/box.ts
+++ b/lib/box.ts
@@ -5,6 +5,6 @@ export function* box<T>(op: () => Operation<T>): Operation<Result<T>> {
   try {
     return Ok(yield* op());
   } catch (error) {
-    return Err(error as Error);
+    return Err(error);
   }
 }
--- a/lib/delimiter.ts
+++ b/lib/delimiter.ts
@@ -86,7 +86,7 @@ export class Delimiter<T>
       }
     } catch (error) {
       this.computed = true;
-      this.outcome = Just(Err(error as Error));
+      this.outcome = Just(Err(error));
     } finally {
       this.finalized = true;
       this.outcome = this.outcome ?? Nothing();
--- a/lib/race.ts
+++ b/lib/race.ts
@@ -45,7 +45,7 @@ export function* race<T extends Operation<unknown>>(
             let value = yield* operation;
             winner.resolve(Ok(value as Yielded<T>));
           } catch (error) {
-            winner.resolve(Err(error as Error));
+            winner.resolve(Err(error));
           }
         }),
       );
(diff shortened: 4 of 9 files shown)


docs/spawn.mdx
```diff
@@ -163,6 +163,20 @@ main(function *() {
 });
 ```
 
+If you need to wait for all operations to settle regardless of whether they
+succeed or fail, use [`allSettled()`][allSettled] instead. Unlike `all()`,
+`allSettled()` never short-circuits on error — each result is represented as
+`{ status: "fulfilled", value }` or `{ status: "rejected", reason }`, just
+like `Promise.allSettled()`.
```
cowboyd: Let's make this a link to MDN

test/all-settled.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        }),
+      ]);
+    });
+
+    await sleep(0);
+    await task.halt();
+
+    expect(teardown).toEqual(true);
+  });
+
```
cowboyd: This test is redundant, so let's leave it out.

lib/all-settled.ts
```diff
@@ -0,0 +1,62 @@
...
+ *  // google => { status: "fulfilled", value: Response }
+ *  // notASite => { status: "rejected", reason: Error }
+ * });
+ * ```
+ *
+ * @param ops a list of operations to wait for
+ * @returns the list of settled results, in the order they were given
+ * @since 4.1
+ */
+export function* allSettled<T extends readonly Operation<unknown>[] | []>(
```
cowboyd, line 35: I'd make this out of `all()` 

```ts
export function allSettled(ops) {
  return all(ops.map(function*(op) {
    try {
      return { status: "fulfilled", value: yield* op };
    } catch (reason) {
      return { status: "rejected", reason } as const;
    }
  }));
```

cowboyd (commented):
I know I said that we should be perfectly aligned with the JavaScript, but now I'm having second thoughts. The reason is because "fulfilled" and "rejected" are very explicitly Promise nomenclature, and operations are not promises.

Also, we have a Result interface which we use everywhere.

thoughts?

joshamaju:
> I know I said that we should be perfectly aligned with the JavaScript, but now I'm having second thoughts. The reason is because "fulfilled" and "rejected" are very explicitly Promise nomenclature, and operations are not promises.
> 
> Also, we have a Result interface which we use everywhere.
> 
> thoughts?

If the goal is to reduce friction for newcomers, aligning with familiar JavaScript terminology like "fulfilled" and "rejected" makes sense. Even though operations aren’t Promises, these terms are widely understood and lower the learning curve.

The alternative would require introducing additional abstractions—such as Effection’s box utility or similar constructs from other libraries to achieve the same behavior.

rauhryan:
Split this into stacked PRs as requested. This PR is now the base PR for exposing `Result<T>`, `Ok()`, and `Err()` as public API. The stacked `allSettled()` implementation lives here: https://github.com/rauhryan/effection/pull/1

cowboyd (approved):

PR #1155, opened 2026-04-16
🐛 Ensure Task.halt() does not resolve before resource cleanup finishes

### LINKED ISSUE #1153: Task.halt() can resolve before resource cleanup finishes
## Summary

`Task.halt()` can resolve while async `resource()` cleanup is still blocked. That makes `halt()` report that shutdown is complete before the resource finalizer has actually finished.

This is observable with only public Effection APIs.

## Reproduction

Tested with `effection@4.0.2`. I also checked `effection-v4.1.0-alpha.7`; the relevant `resource()`, `Task.halt()`, scope destruction, and `Delimiter.exit()` paths appear to have the same cancellation/cleanup shape.

```ts
import { resource, run, until } from "effection";

async function main(): Promise<void> {
  const events: string[] = [];
  const cleanupEntered = Promise.withResolvers<void>();
  const releaseCleanup = Promise.withResolvers<void>();
  const cleanupFinished = Promise.withResolvers<void>();

  function useResource() {
    return resource(function* resourceLifecycle(provide) {
      try {
        yield* provide("resource");
      } finally {
        events.push("cleanup:entered");
        cleanupEntered.resolve();

        yield* until(releaseCleanup.promise);

        events.push("cleanup:after-block");
        cleanupFinished.resolve();
      }
    });
  }

  const task = run(function* parentTask() {
    yield* useResource();
    events.push("main:return");
  });

  await cleanupEntered.promise;
  events.push("outside:calling-halt");

  await task.halt().then(
    () => events.push("halt:resolved"),
    (error: unknown) =>
      events.push(
        `halt:rejected:${
          error instanceof Error ? error.message : String(error)
        }`,
      ),
  );

  events.push("outside:releasing-cleanup");
  releaseCleanup.resolve();
  await cleanupFinished.promise;
  events.push("cleanup:finished-observed");

  await task.then(
    () => events.push("task:resolved"),
    (error: unknown) =>
      events.push(
        `task:rejected:${
          error instanceof Error ? error.message : String(error)
        }`,
      ),
  );

  console.log(events.join("\n"));
}

await main();
```

## Actual output

```text
main:return
cleanup:entered
outside:calling-halt
halt:resolved
outside:releasing-cleanup
cleanup:after-block
cleanup:finished-observed
task:rejected:halted
```

`halt:resolved` happens before `outside:releasing-cleanup` and before `cleanup:after-block`.

## Expected behavior

`task.halt()` should not resolve until all shutdown associated with the task is complete, including async resource cleanup that is already in progress.

Expected ordering should put `halt:resolved` after `cleanup:after-block` / `cleanup:finished-observed`.

## Why this matters

The `Task.halt()` docs say it returns a future that resolves only when all shutdown associated with the task is complete. Code that relies on `await task.halt()` as a join point can proceed while resource finalizers are still running.

That can produce leaked work, use-after-close races, and incorrect teardown ordering.

## Notes

This looks related to the interaction between task scope destruction, `encapsulate(... finally { yield* group.halt(); })`, and `Delimiter.exit()` forcing coroutine completion via `.return()`. A second halt can cut across the in-progress cleanup join, making the parent shutdown path resolve even though a resource finalizer is still suspended and will continue later.

cowboyd:
## Motivation

`Task.halt()` can resolve while async resource cleanup is still running. Code that uses `await task.halt()` as a join point can proceed while resource finalizers are still suspended, producing leaked work and use-after-close races.

Fixes #1153

## Approach

Add a `settling` flag to `Delimiter` that `encapsulate` sets when entering its finally block (before `group.halt()`). While settling, `exit()` skips the `.return()` call since the operation is already winding down naturally. `close()` now also waits for the delimiter future when the outcome is set but not yet finalized.

The flag is reset after `group.halt()` completes so that inner `encapsulate` calls (e.g. `callcc`) don't permanently poison the parent delimiter.

This is a spike — the `settling` flag adds a new state bit to `Delimiter` and a coupling between `encapsulate` and `Delimiter` via `DelimiterContext`. A future refactor could internalize this into the Delimiter itself.

### TODOs and Open Questions

- [ ] Evaluate whether `settling` should live inside the Delimiter's own `[Symbol.iterator]` rather than being set externally by `encapsulate`

### DIFF
--- a/lib/delimiter.ts
+++ b/lib/delimiter.ts
@@ -12,6 +12,7 @@ export class Delimiter<T>
   finalized = false;
   future = withResolvers<Maybe<Result<T>>>();
   computed = false;
+  settling = false;
   routine?: Coroutine;
   outcome?: Maybe<Result<T>>;
 
@@ -33,6 +34,10 @@ export class Delimiter<T>
     this.exit(Nothing());
   }
 
+  settle(): void {
+    this.settling = true;
+  }
+
   *close(): Operation<void> {
     let done = this.future.operation;
     let interrupted = !this.computed;
@@ -45,6 +50,8 @@ export class Delimiter<T>
     };
     if (!this.outcome) {
       this.interrupt();
+      yield* this.close();
+    } else if (!this.finalized) {
       yield* this.close();
     } else {
       if (interrupted && this.outcome.exists && !this.outcome.value.ok) {
@@ -65,7 +72,7 @@ export class Delimiter<T>
     if (!this.routine) {
       this.finalized = true;
       this.future.resolve(this.outcome);
-    } else {
+    } else if (!this.settling) {
       this.routine.return(Ok(this.outcome));
     }
   }
--- a/lib/task-group.ts
+++ b/lib/task-group.ts
@@ -1,6 +1,8 @@
 import { createContext } from "./context.ts";
 import { box } from "./box.ts";
+import { DelimiterContext } from "./delimiter.ts";
 import { Ok, unbox } from "./result.ts";
+import { useScope } from "./scope.ts";
 import type { Operation, Task } from "./types.ts";
 
 export class TaskGroup {
@@ -40,7 +42,14 @@ export function encapsulate<T>(operation: () => Operation<T>): Operation<T> {
     try {
       return yield* operation();
     } finally {
-      yield* group.halt();
+      let scope = yield* useScope();
+      let delimiter = scope.expect(DelimiterContext);
+      delimiter.settle();
+      try {
+        yield* group.halt();
+      } finally {
+        delimiter.settling = false;
+      }
     }
   });
 }
--- a/test/run.test.ts
+++ b/test/run.test.ts
@@ -4,6 +4,7 @@ import { Children } from "../lib/contexts.ts";
 import {
   action,
   createScope,
+  resource,
   run,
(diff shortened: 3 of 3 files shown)


cowboyd:
This PR fixes `task.halt()` not waiting for resource cleanup, but it introduces a regression: `race()` hangs when one arm contains an operation that uses `resource()`, even when that arm completes normally.

Minimal repro (only uses `race`, `resource`, `withResolvers`):

```ts
import { race, resource, withResolvers } from "effection";

yield* race([
  withResolvers<void>().operation,  // never resolves
  (function*() {
    yield* resource<number>(function*(provide) {
      yield* provide(42);
    });
  })(),
]);

console.log("race returned");  // never prints with this PR applied
```

On alpha.7 without the patch this returns immediately. With the patch applied it hangs indefinitely.

**Likely cause:** the new branch added to `Delimiter.close()`:

```js
else if (!this.finalized) {
  yield* this.close();
}
```

When `race` halts the winning candidate task during its post-winner shutdown pass, the candidate's outcome exists (its generator returned normally) but the delimiter isn't finalized yet because the resource is still live in scope. The new branch awaits `this.future`, but nothing drives that future to resolve — `exit()` is now gated by `!this.settling`, which together with the awaiting close produces a deadlock.

Issue #898, opened 2024-02-28
add `allSettled()` operation congruent with `Promise.allSettled()`


cowboyd:
> originally proposed by @bdougherty 

Effection sits parallel to the JavaScript concurrency ecosystem, but with structural concurrency guarantees in place. Our mantra is "if you know JS, then you know Effection"

While we have [`all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) and [`race()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race), we are missing the equivalent of [`allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)

# TODO
- #1150 as a separate PR from #1149

PR #1158, opened 2026-04-23
✨ add allSettled() operation congruent with Promise.allSettled()

rauhryan:
## Motivation

Addresses #898 — Effection has `all()` and `race()` matching `Promise.all()` and `Promise.race()`, but lacks an equivalent of `Promise.allSettled()`.

## Approach

Adds a new `allSettled()` operation that runs all given operations concurrently and waits for every one to settle. Unlike `all()`, it never short-circuits on error — each result is represented as Effection's `Result<T>` shape (`{ ok: true, value }` or `{ ok: false, error }`).

Key decisions:
- Uses Effection's `Result<T>` rather than `PromiseSettledResult<T>` so settled outcomes compose consistently with the rest of the library
- Implements `allSettled()` by composing existing primitives: `all()` over `box()`-wrapped operations
- Sibling operations are not halted when one fails; all members run to completion and their outcomes are preserved
- Documents `Result`, `Ok`, and `Err` as public API since `allSettled()` now exposes `Result<T>` directly

### Files changed
- `lib/all-settled.ts` — `allSettled()` implementation and `AllSettled` mapped type using `Result<T>`
- `lib/result.ts` — public API docs for `Result`, `Ok`, and `Err`
- `test/all-settled.test.ts` — coverage for success, failure, teardown, and type inference using `Result<T>`
- `docs/async-rosetta-stone.mdx` and `docs/spawn.mdx` — updated prose for the `Result<T>` shape and linked result helpers
- `AGENTS.md` — updated guidance for `allSettled()` semantics

### Test coverage
- Empty list resolves to `[]`
- All fulfilled → array of `{ ok: true, value }`
- All rejected → array of `{ ok: false, error }` without throwing
- Mixed outcomes preserve input order
- Does not reject when one operation rejects asynchronously
- Does not halt sibling operations when one fails
- Runs teardown for all operations after completing
- Type inference matches the `Result<T>`-based tuple shape

### DIFF
--- a/lib/all-settled.ts
+++ b/lib/all-settled.ts
@@ -16,12 +16,12 @@ import type { Operation, Yielded } from "./types.ts";
  * ### Example
  *
  * ``` javascript
- * import { allSettled, expect, main } from 'effection';
+ * import { allSettled, main, until } from 'effection';
  *
  * await main(function*() {
  *  let [google, notASite] = yield* allSettled([
- *    expect(fetch('http://google.com')),
- *    expect(fetch('http://nope.example')),
+ *    until(fetch('http://google.com')),
+ *    until(fetch('http://nope.example')),
  *   ]);
  *  // google => { ok: true, value: Response }
  *  // notASite => { ok: false, error: Error }
@@ -45,7 +45,7 @@ export function* allSettled<T extends readonly Operation<unknown>[] | []>(
 
 /**
  * This type allows you to infer heterogenous operation types.
- * e.g. `allSettled([sleep(0), expect(fetch("https://google.com"))])`
+ * e.g. `allSettled([sleep(0), until(fetch("https://google.com"))])`
  * will have a type of `Operation<[Result<void>, Result<Request>]>`
  */
 type AllSettled<T extends readonly Operation<unknown>[] | []> = {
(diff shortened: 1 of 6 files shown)


lib/all-settled.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * ### Example
+ *
+ * ``` javascript
+ * import { allSettled, expect, main } from 'effection';
+ *
+ * await main(function*() {
+ *  let [google, notASite] = yield* allSettled([
+ *    expect(fetch('http://google.com')),
+ *    expect(fetch('http://nope.example')),
```
cowboyd: is `expect()` supposed to be there?

test/all-settled.test.ts
```diff
@@ -0,0 +1,174 @@
...
+          }
+        }),
+        asyncReject(10, "bar"),
+      ]);
+    });
+
+    expect(teardown).toEqual(true);
+  });
+
+  it("runs teardown for all operations after allSettled completes", async () => {
```
cowboyd: teardown should happen _before_ allSettled completes, no?

cowboyd (approved):
This looks great! I'm a little suspicious of the benchmark since it does not exercise any code in this actual benchmark path.

@rauhryan Might be worth installing the godspeed MCP and doing some investigation before you merge just to make sure.

Issue #1159, opened 2026-04-25
Promise-style task.halt() can deadlock later yield* task observation


dcolascione:
## Summary

When a task is halted through the Promise facade (`task.halt().catch(...)`) from inside an Effection operation, a later generator-level observation of the same task (`yield* task`) can hang indefinitely.

The equivalent operation-level halt (`yield* task.halt()`) completes and `yield* task` then reports `TaskHaltedError` as expected.

This repro uses the published upstream package, not a local shim. `npx` installed `effection@4.0.2` for the command below.

## Reproduction

```sh
npx -p effection@latest -p tsx tsx --eval 'import { run, suspend, withResolvers, useScope } from "effection";
async function main(){
const result = await Promise.race([
run(function*(){
  const events:string[] = [];
  const ready = withResolvers<void>("ready");
  const scope = yield* useScope();
  const task = scope.run(function*(){ events.push("child:start"); ready.resolve(); yield* suspend(); });
  yield* ready.operation;
  events.push("parent:halt-request");
  void task.halt().catch(() => {});
  try { yield* task; events.push("parent:task ok"); } catch (error) { events.push(`parent:task:${error instanceof Error ? error.message : String(error)}`); }
  return { events };
}),
new Promise((resolve) => setTimeout(() => resolve({ timeout: true }), 1000)),
]);
console.log(JSON.stringify(result));
}
main().catch((error) => { console.error("main error", error); process.exitCode = 1; });'
```

Actual output:

```json
{"timeout":true}
```

## Control Case

If the halt is driven through the Effection operation interface instead of the Promise interface, the task is observable afterward:

```sh
npx -p effection@latest -p tsx tsx --eval 'import { run, suspend, withResolvers, useScope } from "effection";
async function main(){
const result = await Promise.race([
run(function*(){
  const events:string[] = [];
  const ready = withResolvers<void>("ready");
  const scope = yield* useScope();
  const task = scope.run(function*(){ events.push("child:start"); ready.resolve(); yield* suspend(); });
  yield* ready.operation;
  events.push("parent:halt-request");
  yield* task.halt();
  events.push("parent:halt-done");
  try { yield* task; events.push("parent:task ok"); } catch (error) { events.push(`parent:task:${error instanceof Error ? error.message : String(error)}`); }
  return { events };
}),
new Promise((resolve) => setTimeout(() => resolve({ timeout: true }), 1000)),
]);
console.log(JSON.stringify(result));
}
[...]

Issue #1154, opened 2026-04-14
Interruption can interrupt `yield*` statements in `finally` blocks


dcolascione:
# Problem one: split cleanup

Consider this idiomatic code:

```
function* myWork() {
  try {
    yield* doThing();
  } finally {
    yield* cleanup1();
    yield* cleanup2();
  }
}
```

Having completed async call `doThing()`, our task is blocked waiting on `cleanup1()`. Our task gets cancelled. `cleanup2` never runs. If we were expecting `cleanup2` to reclaim a resource, we now have leak a resource. Sure, we could nest `finally` blocks, putting at most one `yield*` in each. That's not idiomatic code. Nor is eschewing `finally`: `finally` is not only built into the language, but recommended by Effection for general-purpose cleanup.

# Problem 2: cancelled task continue unexpectedly, producing unsound execution

JavaScript forwards `.return()` on generator objects to the innermost generator; it provides it no way to propagate `ReturnCompletion` (which is supposed to unwind a `yield*` chain to top level) if a generator in the `yield*` chain contains `yield*`s in its `finally` block. For example,

```
import { run, spawn, suspend, sleep, type Operation } from "effection";

function* inner(): Operation<number> {
  try {
    yield* suspend();
    return 123;
  } finally {
    // This single yield is enough to break the .return() unwind chain.
    yield* sleep(0);
  }
}

await run(function* (): Operation<void> {
  const task = yield* spawn(function* (): Operation<void> {
    const value: number = yield* inner();

    console.log("runtime value:", value);
    console.log(value.toFixed(1)); // type-checks, but crashes at runtime
  });

  yield* sleep(0);
  yield* task.halt();
});
```

Control reaches `console.log` because Effection implements cancellation as a single `.return()` on the coroutine iterator; JavaScript forwards `.return()` to the innermost iterator, which forces `suspend()` to return, which results in `sleep(0)` running in the `finally` block, which, as a side effect of yielding, breaks the `ReturnCompletion` chain that would propagate the `.return()`-instigated unwind up the generator stack. Thus, `inner` returns some value normally, which the spawned task tries to process as if it were an intended part of program flow — `value` is not a number, so `value.toFixed(1)` crashes at runtime despite the program being correctly typed. The program is unsound.

[...]

PR #1164, opened 2026-05-07
♻️ Invert reducer ↔ delimiter control

cowboyd:
# Motivation

Issues #1153, #1154, and #1159 share a single root cause: cancellation
in v4 is one-shot.

The proposed fix is to have the co-routine, after cancellation, continually return, and return, and return with each yield, and then protect "critical" sections with a shield that makes normal execution for secments of the effect stream.

This PR is an attempt to enable these refactors by allowing the conetxt control exactly how the co-routine is advanced at every stem.

# Approach

Invert control from the co-routine to the delimiter.


taras (approved):
rubber

jbolda (approved):

PR #1165, opened 2026-05-07
🐛 wrap non-Error causes passed to Err()

rauhryan:
# Motivation

`Err()` currently accepts `Error`, but JavaScript `catch` blocks and promise rejections can produce `unknown` thrown values. That makes it awkward to pass caught values directly to `Err()` and can leak non-`Error` throw values through APIs that model failures as `Error`.

This splits the `Err()` behavior change out so it can target the same base branch as #1149.

# Approach

- change `Err()` to accept `unknown`
- preserve existing `Error` instances as-is
- wrap non-`Error` causes in `ThrowValueError(String(cause), { cause })`
- remove internal `as Error` casts where caught values are passed to `Err()`
- add regression coverage for `Err()` and `until(Promise.reject(...))`

### DIFF
--- a/lib/action.ts
+++ b/lib/action.ts
@@ -63,7 +63,7 @@ export function action<T>(executor: Executor<T>, desc?: string): Operation<T> {
               discard();
               discarded(Ok());
             } catch (error) {
-              discarded(Err(error as Error));
+              discarded(Err(error));
             }
           };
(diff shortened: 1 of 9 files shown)


cowboyd:
This has been superseded by #1149

Issue #1160, opened 2026-04-26
`yield* task.halt()` from the same task is a self-join


dcolascione:
Upstream `test/run.test.ts` currently has these tests:

- `can halt itself`
- `can halt itself between yield points`

Both call `yield* task.halt()` from inside the task being halted.

That seems like a broken API pattern. `halt()` requests cancellation and waits for the task's teardown to finish. If the task itself is yielding on that drain, teardown cannot finish until the current generator unwinds. That's a self-join.

Minimal shape:

```ts
let task: Task<void> = run(function* () {
  yield* sleep(0);
  yield* task.halt();
});

await task;
```

I think self-cancel needs a request-only operation instead, e.g. `task.requestHalt()` / `task.requestCancel()`, with `task.halt()` kept as the external request-and-drain API.

If `yield* task.halt()` from inside the same task is intentionally unsupported, it would be useful for Effection to document that or reject it with a clear error instead of treating it as normal cancellation behavior.

cowboyd:
> That seems like a broken API pattern.

I agree. It has been supported since v1, but not strongly. In reality, it has carried forward into version 4 via inertia more than anything else. IIRC, the original rationale was that because it was "possible", then it should be "doable." 

Looking at it now, however, it feels like this should be a hard exception. I have personally done this precisely zero times in almost ten years, and I would look very critically on any code that came across my desk doing so.

Thank you for taking to report all these things. They very helpful, and I wish only that I could fix them more quickly.

cowboyd:
I think step one is disallowing self-join entirely. https://github.com/thefrontside/effection/pull/1162

For cases where a task wishes to request its own cancellation, that is certainly doable within user-land. We could consider moving it into core at a later point.

cowboyd:
Fixed with https://github.com/thefrontside/effection/pull/1168

PR #1168, opened 2026-05-30
♻️ Unify task / coroutine lifecycles

### LINKED ISSUE #1159: Promise-style task.halt() can deadlock later yield* task observation

cowboyd:
## Motivation

> fixes #1159 #1160 #1153 #1128 
> closes #1167 #1155 #1162
> sets up #1154 

There were correctness issues in v4 that arose when halting. They all stemmed from the fact that different parts of the codebase had different ideas about when a task was actually "done" — especially when that task was halted. This led to multiple races around the question.

## Approach

This rectifies the situation in the following ways:

1. Every coroutine has a single definite future that can only be resolved when the coroutine is completely exhausted.
2. The task future is derived entirely from the coroutine's future.
3. The task's halt() future is also derived entirely from the coroutine's future.

For tasks, error / success / halt all flow through the same code path: the complete conclusion of the coroutine's body. The only way to see _any_ of these values is for the reducer to ask the coroutine to produce { done: true, value: T }. Previously, we were just dropping this on the floor and resolving a future in a finally {} block of the body.

This entailed making the coroutine layer smarter than before, but the result is a simpler architecture. Rather than being managed by a host of reducers, delimiters, and error boundaries, the coroutine is now the primary organ of evaluation. Coroutines themselves — not instructions — are added to and removed from the reducer queue, so there is no after-the-fact validation of instructions to drop stale ones. A coroutine that changes its internal state between ticks is still fresh when it is dequeued. This eliminated the need for tickets.

We were also able to make trap() the single abstraction for error containment. There is no longer a separate error boundary at the task level.


lib/task.ts
```diff
@@ -1,149 +1,221 @@
...
-  try {
-    yield* delimiter;
-  } finally {
-    scope.set(ErrorContext, original.error);
-    scope.set(DelimiterContext, original.delimiter);
-    let outcome = delimiter.outcome!;
-    return (yield {
-      description: "trap return",
-      enter(resolve) {
+class TaskInternal<T> implements TaskInterface<T> {
```
jbolda: Is the internal version to avoid `_promise` in the public API or is there someone more?

cowboyd: @jbolda precisely. There are several reasons:

1. hide private fiels
2. "bind" public methods: `{ let halt = task.halt; yield* halt() }` // <- works
3. make debug output nice: `Task {}` instead of Task { _promise: ..., control:  { routine:  {.... } }}`

Basically make it a very friendly commitment to the public api that lets us have a very free hand about how we change everything inside.

lib/task-group.ts
```diff
@@ -2,24 +2,19 @@ import { createContext } from "./context.ts";
...
-  *halt(): Operation<void> {
+  *halt() {
+    let set = this.tasks;
     let total = Ok();
-    while (this.tasks.size > 0) {
-      let tasks = [...this.tasks].reverse();
-      this.tasks.clear();
-      for (let task of tasks) {
+    while (set.size > 0) {
+      let tasks = [...set];
```
jbolda, line 14: Not `.reverse()`ing anymore? Do we add them in reverse order already?

test/run.test.ts
```diff
@@ -386,4 +414,40 @@ describe("run()", () => {
     expect(error?.message).toEqual("halted");
     expect(halted).toEqual(true);
   });
+
+  // E1: parent body returns; destroy iterates children sequentially.
+  // While awaiting one child's slow halt, another sibling errors in the
+  // background. That error is captured via raise → forcedOutcome → task
+  // settles with the background error.
+  // TODO: we may need another test case to make sure that all siblings are torn down.
```
jbolda: Planned for future PR?

jbolda (approved):

lib/task-group.ts
cowboyd: It is still reverse, but the way it's accomplished is different. if you look at the iteration below, it counts down to zero.

```ts
for (let i = tasks.length - 1; i >= 0; i--) {
```

The biggest refactor here is that we capture the current set of tasks in the set, then we halt them in reverse order, and we loop around again while `set.size > 0` so that in case something happened that added more tasks while we were halting, we halt those as well.

PR #1169, opened 2026-06-02
✨ CodSpeed benchmarks

taras:
# Motivation

We want to start measuring performance and tracking it to prevent regressions before a PR is merged. We have a bunch of scenarios that we created over the last few years, but they're not tracked anywhere. [Codspeed](https://codspeed.io) was recently brought to our attention. This adds Codspeed integration via tinybench.

# Approach

## Not using Deno

Unfortunately, it's not possible to use Deno with Codspeed because Codspeed relies on V8 APIs that are not exposed by Deno. Codspeed has a napi module, but that module reaches into V8 symbols not exported by Deno.

## Deno to Node integration

To use it with Node, we generate a bundle then install that bundle with npm which gives us the ability to run the Node runner with tinybench.

## Single entry point

There is a file called `codspeed-bench.ts` in `bench-node` directory that imports all scenarios and executes them.

## CI Integration

In CI, we capture the results and send them to Codspeed

## Run locally


cowboyd:
LGTM, but rather than have two sets of benchmarks in `bench/` and `bench-node/`, let's just make our benchmarks be the benchmarks. I think it's safe to get rid of the old ones and just make `bench-node` -> the canonical `bench/`

cowboyd (approved):
⚡

PR #1184, opened 2026-06-10
⚡️Use prototyped class for ScopeInternal

cowboyd:
# Motivation

The scope implementation was creating a unique prototype for every single scope instance which caused depotimization of every single scope method.

# Approach
This extracts all the core methods into a single prototype based class that shares a single implementation among all instances, which allows V8 (and presumably other runtimes) to inline the implementation.

### DIFF
--- a/lib/scope-internal.ts
+++ b/lib/scope-internal.ts
@@ -8,67 +8,102 @@ import type { Context, Operation, Scope, Task } from "./types.ts";
 export function createScopeInternal(
   parent?: Scope,
 ): [ScopeInternal, () => Operation<void>] {
-  let destructors = new Set<() => Operation<void>>();
-  let destruction = createFuture<void>();
+  let impl = new ScopeImpl(parent as ScopeInternal);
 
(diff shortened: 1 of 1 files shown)


cowboyd:
### Pre-merge check: verify Symbol-keyed augmentations still work

The wrapper change moves `scope` from `Object.create({...closures...})` to `Object.create(Scope, {own-bound-methods})`. `Symbol.toStringTag` lives on the shared `Scope` prototype (good), but two Symbol-keyed properties are added *after* construction by `createScope`:

- `Symbol.iterator` — `lib/scope.ts:75-78`, used for tuple destructuring `let [scope, dispose] = createScope()`
- `Symbol.asyncDispose` — `lib/scope.ts:80-83`, used by `await using scope = createScope()`

Both rely on `Object.defineProperty(scope, ...)` succeeding on the returned scope object. Should still work since the wrapper is a plain object, but worth running:

1. A test that destructures `createScope()` as a tuple
2. A test that uses `await using` semantics

If both pass on this branch, this comment can be resolved.

PR #1186, opened 2026-06-10
🐛 make sure to propagate unwind outside scoped

### LINKED ISSUE #1185: scoped() + race() prevents halt propagation in callcc resolution path
## Description

When `callcc` is resolved externally (e.g., via `scope.run(() => resolve(...))` as used by `main()` for SIGINT handling), the spawned body should be halted and its generator should unwind. This works correctly in most cases, but when the body contains a `for` loop with `scoped()` wrapping operations that use `race()`, halt does not propagate. The loop runs to completion instead of stopping.

This is the pattern used by `main()` for SIGINT: the SIGINT handler calls `scope.run(() => resolve({ status: 130 }))`, which resolves the `callcc` and should tear down the body via strict structured concurrency.

## Reproduction

```typescript
import { callcc, race, run, scoped, sleep, useScope } from "effection";

let events: string[] = [];
let triggerShutdown;

let task = run(function* () {
  return yield* callcc(function* (resolve) {
    let scope = yield* useScope();

    triggerShutdown = () => {
      scope.run(() => resolve({ status: 130 }));
    };

    for (let round of [1, 2, 3]) {
      events.push(`round-${round}-start`);

      yield* scoped(function* () {
        events.push(`round-${round}-resource`);

        try {
          for (let step of [1, 2, 3]) {
            let result = yield* race([
              function* () {
                yield* sleep(5000);
                return "timeout";
              }(),
              function* () {
                yield* sleep(50);
                return `step-${step}`;
              }(),
            ]);
            events.push(`step-${step}-result: ${result}`);
          }
        } finally {
          events.push(`round-${round}-teardown`);
        }
      });

      events.push(`round-${round}-complete`);
    }
  });
});

// Let round 1 start
await new Promise((r) => setTimeout(r, 10));

// Simulate SIGINT
triggerShutdown();

try { await task; } catch {}

console.log(events);
```

**Expected:** `round-1-start`, `round-1-resource`, `round-1-teardown`. The loop stops. Round 2 never starts.

**Actual:** All 3 rounds run to completion with all steps producing results. The halt has no effect on the loop.

## Narrowing

| Pattern | Halt works? |
|---|---|
| `callcc` + `sleep` | Yes |
| `callcc` + `race` (no `scoped`) | Yes |
| `callcc` + `scoped` (no `race`) | Yes |
| `callcc` + `scoped` + `race` | **No** |

The combination of `scoped()` and `race()` in the `callcc` halt path is the trigger.

## Impact

This breaks SIGINT handling in any application that uses `main()` with `scoped()` round boundaries and `race()`-based convergence (e.g., timeouts). The `main()` SIGINT handler uses exactly this `callcc` + `scope.run(resolve)` pattern.

cowboyd:
# Motivation

Fix #1185.

 If a routine was halted while suspended inside scoped(), code after yield* scoped(...) would still run.

iter.return() yielded the destroy() effects in scoped's finally, but the resume came back as iter.next() which takes the frame out of return-mode. trap() re-arms via trap.exit(); scoped() had no equivalent. 

Fix: inline the Trap, run destroy() in critical() so cleanup completes, then yield trap.exit() to re-arm unwind based on the inner trap's outcome.

### DIFF
--- a/lib/scoped.ts
+++ b/lib/scoped.ts
@@ -1,7 +1,9 @@
 import type { Operation } from "./types.ts";
-import { trap } from "./trap.ts";
-import { useCoroutine } from "./coroutine.ts";
+import { Trap, trap } from "./trap.ts";
+import { critical, useCoroutine } from "./coroutine.ts";
 import { createScopeInternal } from "./scope-internal.ts";
+import { Just } from "./maybe.ts";
+import { Err, Ok } from "./result.ts";
 
 /**
  * Encapsulate an operation so that no effects will persist outside of
@@ -32,12 +34,17 @@ export function scoped<T>(operation: () => Operation<T>): Operation<T> {
       let routine = yield* useCoroutine();
       let original = routine.scope;
       let [scope, destroy] = createScopeInternal(original);
+      let t = new Trap<T>(routine);
       try {
         routine.scope = scope;
-        return yield* trap(operation);
+        t.outcome = Just(Ok(yield* trap(operation)));
+      } catch (error) {
+        t.outcome = Just(Err(error));
       } finally {
         routine.scope = original;
-        yield* destroy();
+        yield* critical(destroy);
+        // deno-lint-ignore no-unsafe-finally
+        return (yield t.exit()) as T;
       }
     },
   };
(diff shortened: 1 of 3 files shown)


taras (approved):
Toight

Issue #1188, opened 2026-06-12
External resolve in action loop advances one extra iteration before halt


tarasvalstro:
## Summary
This appears to be in the same family as #1185, but broader than `scoped + race`.

When shutdown is triggered externally via `scope.run(() => resolve(...))` from inside an `action`-orchestrated loop, execution advances into the next loop iteration before teardown settles.

## Expected
After shutdown trigger, execution should stop (except teardown/finally). It should **not** enter the next loop iteration.

## Actual
Execution records next-iteration start (e.g. `round-2-start`) after shutdown trigger.

taras:
Added a reproduction in #1189 (failing tests against `v4`).

**Key finding:** I included a *minimal baseline* test that strips out `scoped()` and `race()` entirely — just `action` + a `for` loop + `sleep` + an external `resolve` via `scope.run(...)`. It still leaks `round-2-start`:

```
round-1-start
round-failed-trigger-shutdown
round-2-start        ← should not happen
```

So this is not a `scoped`/`race` problem. Resolving an `action` from the outside (via `scope.run`) does not unwind the operation currently suspended awaiting that action before the loop takes its next step — the in-flight iteration advances once more.

This confirms #1188 is the broader category: #1185 was the narrower `scoped` + `race` manifestation (fixed in #1186 / `49fe66f0`), but the same "external resolve doesn't halt the in-flight iteration" bug reproduces with a plain `action` loop.

Repro to run:
```
deno test --allow-run --allow-read --allow-env --allow-ffi test/scoped.test.ts
```

taras:
## Investigation: this is not a #1185/#1186-class unwind-propagation bug

I traced the repro through the scheduler and then verified empirically. The leak of `round-2-start` is **structural**: it happens because the observable statement sits *before* the next `yield`, so it runs in the same synchronous step as the shutdown trigger — before any halt can land.

### Trace (minimal baseline)

Let `P` = the `run` task, `OP` = the child task running the loop, `RESOLVE` = the task spawned by `scope.run(() => resolve())`.

1. `OP` resumes after round 1's `sleep`, pushes `round-failed-trigger-shutdown`, then `scope.run(resolveTask)` only **enqueues** `RESOLVE` on the reducer's instruction queue (it does not run inline — `createTask` schedules, and the reducer is already `reducing`).
[...]

PR #1189, opened 2026-06-12
🐛 repro: external resolve advances action-orchestrated loop (#1188)

taras:
Reproduction for #1188. **These tests are expected to fail** — they document the bug and should go green once the underlying fix lands.

## What it shows

When shutdown is triggered externally via `scope.run(() => resolve(...))` from inside an `action`-orchestrated loop, the loop advances into the **next iteration** (`round-2-start`) after the shutdown trigger, instead of stopping once the current iteration finishes.

Two tests are added to `test/scoped.test.ts`:

1. **`does not advance loop after external resolve with nested scoped + race`** — mirrors the original report (`action` wrapping a loop, each round runs `scoped(scoped(race(...)))`). Observed events:
   ```
   round-1-start
   round-1-result:false
   round-failed-trigger-shutdown
   round-2-start        ← leaks
   ```

2. **`does not advance loop after external resolve in minimal action baseline`** — strips out `scoped()` and `race()` entirely; just `action` + a `for` loop + `sleep` + external `resolve`. Still leaks:
   ```
   round-1-start
   round-failed-trigger-shutdown
   round-2-start        ← leaks
   ```

## Why the baseline matters

The minimal baseline has no `scoped` and no `race`, yet it still advances to `round-2-start`. That isolates the root cause to the **`action` + external-resolve path** — i.e. resolving an `action` from outside (via `scope.run`) does not unwind the operation that is currently suspended awaiting that action before the loop takes its next step.

This is the broader category #1188 refers to: #1185 was the narrower `scoped` + `race` manifestation (fixed in #1186 / `49fe66f0`), but the same "external resolve doesn't halt the in-flight iteration" problem reproduces with a plain `action` loop.

## Run

```
deno test --allow-run --allow-read --allow-env --allow-ffi test/scoped.test.ts
```

### DIFF
--- a/test/scoped.test.ts
+++ b/test/scoped.test.ts
@@ -1,12 +1,16 @@
 import { box } from "../lib/box.ts";
 import {
+  action,
   createContext,
+  race,
   resource,
   run,
   scoped,
   sleep,
   spawn,
   suspend,
+  type Operation,
+  useScope,
 } from "../mod.ts";
 import { describe, expect, it } from "./suite.ts";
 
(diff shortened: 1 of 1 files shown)


taras:
this is not a problem. It just shows that we can interrupt sync loop which is not expected to be possible

PR #1195, opened 2026-06-25
document critical()

taras:
# Motivation

We want people to be able to use `critical` for which it needs to be exported and documented.

# Approach

Added JSDoc to explain that `critical()` prevents unwind from interrupting teardown, and included an example showing the intended `finally` usage with multiple teardown steps. Also exported `critical` from `lib/mod.ts` so it is available through the public entry point.

### DIFF
--- a/lib/mod.ts
+++ b/lib/mod.ts
@@ -26,3 +26,4 @@ export * from "./with-resolvers.ts";
 export * from "./async.ts";
 export * from "./scoped.ts";
 export * from "./until.ts";
+export { critical } from "./coroutine.ts";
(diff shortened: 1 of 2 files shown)


lib/mod.ts
```diff
@@ -26,3 +26,4 @@ export * from "./with-resolvers.ts";
 export * from "./async.ts";
 export * from "./scoped.ts";
 export * from "./until.ts";
+export { critical } from "./coroutine.ts";
```
cowboyd, line 29: Do we want to export critical yet? We don't yet have full sticky return, so it won't actually be of benefit yet unless you try to interrupt something that is already in its `finally {}` block.

cowboyd:
We're going to keep it private for now and recommend that you don't yield in finally, you use ensure instead.

Issue #1190, opened 2026-06-12
Halt does not propagate past a `yield*`-delegated sub-operation whose `finally` yields


tarasvalstro:
## Summary

When a task is halted while suspended, and the suspended point is inside a
`yield*`-delegated sub-operation whose `finally` performs **asynchronous work**
(any `yield*`), the coroutine does not stop after that `finally` completes — it
**resumes normal execution past the halt**, running subsequent statements and
even subsequent `yield*` operations. A synchronous `finally` behaves correctly.

## Environment

- effection `4.0.3` (also reproduces on `main` / source tree)
- Reproduced on Node 22 and Deno

## Minimal reproduction

```js
import { run, sleep, suspend } from "effection";

function* halted() {
  try {
    yield* suspend();
  } finally {
    yield* sleep(0); // any async work in the finally
  }
}

let leaked = false;
const task = run(function* () {
  yield* halted(); // delegated sub-operation
  leaked = true;   // BUG: runs after the task was halted
});

await task.halt();
console.log("leaked:", leaked); // prints true; expected false
```

As a failing test in the style of `test/scoped.test.ts`:

```ts
it("does not run code after a halted operation whose finally yields", async () => {
  let leaked = false;

  function* halted() {
    try {
      yield* suspend();
    } finally {
      yield* sleep(0);
    }
  }

  let task = run(function* () {
    yield* halted();
    leaked = true; // must not run: this is after a halted operation
  });

  await task.halt();

  expect(leaked).toBe(false);
});
```

## Observed behavior matrix

| code after the halted delegated op | finally        | result                         |
| ----------------------------------- | -------------- | ------------------------------ |
| bare statement                      | `yield* sleep` | runs (BUG)                     |
| bare statement                      | synchronous    | does not run (OK)              |
| inside `try { } catch { }`          | `yield* sleep` | runs (BUG)                     |
| inside `try { } finally { }`        | `yield* sleep` | runs (BUG)                     |
| statement + another `yield*` after  | `yield* sleep` | both run — full resume (BUG)   |

The last row is the most concerning: the coroutine fully resurrects and keeps
yielding after the halt as if it had never been halted.

The existing test `can suspend in yielded finally block` does not catch this
because it has no normal code after the delegation; it only asserts `finally`
ordering and that the task reports `halted` (which `task.halt()` reports
regardless of whether the body continued).

## Root cause

`lib/coroutine.ts`, `step()`:

```ts
} else if (data.unwinding && !data.critical) {
[...]

PR #1196, opened 2026-06-29
test: failing test suite for #1190 halt-propagation leak

taras:
# Motivation

I wanted to make sure that we covered all of the different permutations of the problem in [#1190](https://github.com/thefrontside/effection/issues/1190).

I don't think we should merge these tests but I'm hoping that they would be useful for figuring out #1190.

# Approach

Many of these cases are duplicated mechanically but they should be unique from syntax use case perspective.


cowboyd:
We're going with `ensure()` not `yield` in `finally{}`

PR #1204, opened 2026-07-05
⬆️ Upgrade @deno/doc to 0.199.0 (native v2 migration)

taras:
# Motivation

The docs site generates its API Reference with `@deno/doc`, pinned at 0.188.0. `@deno/doc` ≥ 0.195.0 replaced the flat v1 `DocNode[]` model with a v2 `Document`/`Symbol`/`Declaration` model, so staying current means migrating the rendering pipeline. This PR does that migration **natively** — no compatibility shim — so the codebase speaks v2 directly.

It also fixes a latent bug the version bump surfaces: v2 changed `Location.line` from 1-indexed to 0-indexed (deno_doc #150/#777), but the "View code" links feed it straight into GitHub's 1-indexed `#L<n>` anchor, so every source link pointed one line above the symbol's declaration.

Base of a 2-PR stack; #1205 (experimental APIs) stacks on this.

# Approach

Two commits.

**⬆️ Upgrade + native v2 migration** — `@deno/doc` 0.188.0 → 0.199.0, `@deno/graph` 0.89.0 → 0.100.1, and the `www` CI job to Deno v2.9.1 (matches #1202). The pipeline consumes v2 directly:

- `doc()` returns `Record<url, Document>`; `Document.symbols` group declarations by name. `DocPageSection` holds a raw `Declaration`; render helpers take the owning symbol via `type SymbolInfo = Omit<Symbol, "declarations">` — `extract(declaration, symbol)`, `Type({ declaration, symbol })`, `exportHash(declaration, symbol, i)`.
- `.def` replaces the per-kind def fields; `ClassMethodDef.def`; every `TsType*Def` payload is `.value`; namespace members recurse over `NamespaceDef.elements` (member `Symbol`s).
- Dropped now-invalid `kind === "import"` checks; `Uint8Array` cast in `blog-image-route.ts`.

**🐛 Fix off-by-one in "View code" source links** — add `+ 1` where the GitHub `#L` anchor is built, in `package.ts` and `package/node.ts`. Verified: `run` is declared on source line 32 and the link points to `#L32` (was `#L31`, landing on the `*/` above it).

> **Why 0.199.0 and not latest (0.202.0)?** `@deno/doc` 0.200.0–0.202.0 ship broken wasm bindings that fail to instantiate under Deno (`WebAssembly.instantiate(): Import #0 "./__wbindgen_placeholder__"`), returning 500 for every docs page. 0.199.0 is the newest usable v2 release.

### Validation

Ran the site locally: API pages (`/api/v4/run`, `/api/v4/call`, `/api/v3/spawn`) render identically to before; `/x/websocket` and `/x/stream-helpers` exercise the `package/node.ts` path; the `run` source link verified at the correct line. `deno fmt --check`, `deno lint`, and `deno check main.tsx` pass.


cowboyd (approved):
I opened it up and browse the API docs. All the links I checked seem to work. It wasn’t exhaustive but hopefully representative.

PR #1205, opened 2026-07-05
✨ Show experimental APIs on the API docs

### LINKED ISSUE #1200: Show Experimental APIs on the apidocs

taras:
# Motivation

Closes #1200.

We're adding experimental APIs in 4.1, but we don't currently have a way to show experimental APIs on the website. 

# Approach

1. `useDenoDoc` includes `experimental` flag which is added for symbols that are exported from `./experimental` package export
2. added a route that shows experimental symbols at `/api/${s.name}/experimental/:symbol``
3. list pages now takes groups of symbols to show which includes experimental and default export
4. showing a banner on experimental routes

image

image

### DIFF
--- a/www/components/api/api-page.tsx
+++ b/www/components/api/api-page.tsx
@@ -1,34 +1,39 @@
-import { all } from "effection";
+import { all, type Operation } from "effection";
 import type { JSXElement } from "revolution";
 import { useConfig } from "../../context/config.ts";
 import { LocalDocPage } from "../../hooks/use-deno-doc.tsx";
 import { ResolveLinkFunction, useMarkdown } from "../../hooks/use-markdown.tsx";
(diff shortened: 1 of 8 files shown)


cowboyd:
One thing I noticed is that it is using the JSR name for the endpoint "@effection/effection/experimental"

But for consumers from NPM, and honestly, consumers from JSR that use an import map to map `@effection/effection` -> `effection` will see this in code as `import { createApi } from "effection/experimental"

taras:
Good catch — fixed in c0766efc. The experimental banner now reads `effection/experimental` (the npm name / import-map form) instead of the JSR-scoped `@effection/effection/experimental`, matching what consumers actually write in an import.

cowboyd (approved):

PR #1207, opened 2026-07-05
✨ Redesign the Effection home page (GC-framed arc + interactive tabs)

taras:
# Motivation

The old home page — logo, tagline, and a feature grid — doesn't make the case for Effection to the people most likely to adopt it: **JS library authors**. This rebuilds it from the mockup around a concrete narrative that reasons from an analogy every JS developer already trusts — garbage collection — and extends it to async.

The page:

1. **The deal** — "JavaScript cleans up your memory. It never cleans up your async." Values belong to a scope and get reclaimed; async work belongs to nothing.
2. **The exception** — the runtime can't reclaim async work because it belongs to no scope, so every `AbortSignal`, unsubscribe, `close()`/`finally`, and `using` is you reconstructing the missing lifetime by hand.
3. **The name** — that missing property is *structured concurrency*.
4. **Effection** — brings it to JS: <6KB, dependency-free, no build step, 5 years in production.
5. **See for yourself** — five real use cases that each look too simple to be correct, in an interactive tab set.
6. **The proof** — an interactive operation tree, three superpowers (**halt**, **context**, **synchronous-by-default**), the GC-analogy capstone, an async/await ↔ Effection rosetta, and a closing CTA.

# Approach

Built entirely on the site's own stack — `revolution` server-rendered JSX + Tailwind v4 — reusing the shared header/footer. No client framework. Dark-mode variants throughout.

Components:

- **`components/usecase-tabs.tsx`** — the "see for yourself" tabs (timeout, parallel, stream, background, async teardown), each a code sample + one-line guarantee. **CSS-only** (radio/checkbox-hack) — pill + panel switching work with **no JavaScript**.
- **`components/code-window.tsx`** — a macOS-window code block, syntax-highlighted at render time with `refractor` (the tokenizer `rehype-prism-plus` already uses), so colours match the site and follow the OS light/dark preference.
- **`components/operation-tree.tsx` + `assets/home-islands.js`** — a static SVG driven by a small vanilla-JS island that animates the halt↓ / teardown↑ cascade on click.
- **`components/install-copy.tsx`** — a copy-to-clipboard install pill (same island).

**Note on colours:** brand colours use arbitrary hex values (`text-[#14315D]` etc.) because the custom Tailwind tokens (`blue-primary` / `blue-secondary` / `pink-secondary`) are **not emitted by the current v4 build** — the compiled `tailwind/main.css` contains none of them (true on `v4` today too). Worth fixing globally in a follow-up.

Rebased on `v4`; `deno fmt --check`, `deno lint`, and `deno check` all pass. Verified end-to-end against a local dev server (light, dark, tab switching, halt animation, copy button).

## Screenshots

**Hero — the memory-vs-async framing + "The exception"**

Hero and the exception

**See for yourself — interactive use-case tabs** (shown with the "Stream" tab selected; switching is pure CSS)

Use-case tabs

**Three superpowers — halt, context, synchronous-by-default**

Three superpowers

**Dark mode**

Dark mode hero

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/www/assets/home-islands.js
+++ b/www/assets/home-islands.js
@@ -9,15 +9,14 @@ const IDLE_EDGE = "#cbd5e1";
 
 function initOperationTree() {
   let root = document.getElementById("operation-tree");
-  let haltBtn = document.getElementById("operation-tree-halt");
-  let resetBtn = document.getElementById("operation-tree-reset");
-  if (!root || !haltBtn || !resetBtn) return;
+  let btn = document.getElementById("operation-tree-halt");
+  if (!root || !btn) return;
 
   let nodes = Array.from(root.querySelectorAll("[data-node]"));
   let edges = Array.from(root.querySelectorAll("[data-edge]"));
   let state = {}; // id -> "halting" | "done"
   let timers = [];
-  let running = false;
+  let phase = "alive"; // "alive" | "running" | "torndown"
 
   function nodeFill(st) {
     if (st === "done") return "#fffbeb";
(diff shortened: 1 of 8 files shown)


joshamaju:
This looks great.

Just a little suggestion, can we replace the gradients with a single colour? They give of "vibe coded".

PR #1208, opened 2026-07-05
♻️ Consume @deno/doc v2 natively + fix source-link off-by-one

taras:
# Motivation

Third in a stack — on top of #1205 (which is on #1204). Review/merge those first.

The `@deno/doc` upgrade (#1204) kept the old flat-node shape alive with a compatibility adapter — `type DocNode = Declaration & { name }` plus a per-declaration name re-attachment — so the renderer never actually spoke v2's `Symbol`/`Declaration` model. This PR removes that shim and consumes v2 natively.

It also fixes a latent bug the upgrade introduced: `@deno/doc` v2 (≥0.195.0) changed `Location.line` from 1-indexed to 0-indexed (deno_doc #150/#777), but the "View code" links feed it straight into GitHub's 1-indexed `#L<n>` anchor — so every source link pointed one line above the symbol's declaration.

# Approach

Two commits.

**🐛 Fix off-by-one in "View code" source links** — add `+ 1` where the GitHub `#L` anchor is built, in `package.ts` and `package/node.ts`. Verified: `run` is declared on source line 32 and the link now points to `#L32` (was `#L31`, landing on the `*/` above it).

**♻️ Drop the `DocNode` compatibility shim** — consume v2's `Symbol`/`Declaration` directly:

- `DocPageSection.node: DocNode` → `declaration: Declaration`.
- Render helpers take the owning symbol's identity via `type SymbolInfo = Omit<Symbol, "declarations">` (a full `Symbol` is structurally assignable, so build-time code passes the real `Symbol`; render-time passes `{ name: page.name }`). Signatures become `extract(declaration, symbol)`, `Type({ declaration, symbol })`, `exportHash(declaration, symbol, i)`.
- Namespace members recurse over `NamespaceDef.elements` (real member `Symbol`s) instead of synthesizing flat nodes.
- `DocPage.name` is intentionally kept (many call sites read it); `SymbolInfo` flows only through the three render helpers.
- Both docs paths updated: `package.ts` (API routes) and `package/node.ts` (the `/x` contrib packages).

No rendered-output change beyond the line-link fix — this is a structural refactor.

### Validation

Ran the site locally: `/api/v4/run`, `/api/v4/call`, `/api/v4-next/experimental/createApi` (banner + 2 badges intact), and `/api/v3/spawn` render with full content; `/x/websocket` (3 source links) and `/x/stream-helpers` (25) exercise the `package/node.ts` path; the `run` source link verified at the correct line. `deno fmt --check`, `deno lint`, and `deno check main.tsx` pass.


taras:
Folded into #1204 — the `@deno/doc` upgrade now migrates straight to the native v2 model (no intermediate `DocNode` shim), and the source-link off-by-one fix rides along. #1205 (experimental APIs) now stacks directly on #1204.

PR #1213, opened 2026-07-10
Deploy To Netlify

jbolda:
# Motivation

Shifting to Netlify as we can directly push our assets. Mirrors https://github.com/thefrontside/frontside.com/pull/488 and also triggers a build there as the content here gets pulled in there.

### DIFF
--- a/.github/workflows/www.yaml
+++ b/.github/workflows/www.yaml
@@ -24,7 +24,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


.github/workflows/www.yaml
```diff
@@ -1,25 +1,30 @@
 name: www
+
 on:
-  workflow_dispatch: # Allows manual trigger
-  schedule:
-    - cron: "0 */8 * * *" # Runs every 8 hours
```
taras, line 5: The nested websites (effection, interactors, graphgen) do not notify frontside.com of changes, so it doesn't get redeployed when they merge.

The cron was supposed to redeploy in case there are updates on the nested sites. 

Are you sure we want to remove the cron?

taras (commented):
One comment about the cron

.github/workflows/www.yaml
jbolda: Aye, I added a workflow dispatch at the end which _should_ trigger it.

taras: Where is this dispatch?

jbolda: Hopefully this diff link works. Otherwise it is the last job in this file.
https://github.com/thefrontside/effection/pull/1213/changes#diff-f1f68329fbd1f97c7d44033c05fe1920f828e8b950febce8be7aca12945ebbd8R158-R173

taras: Oh, i see. Also, I missed that we're in effection repo and not frontside.com :)

Did you try it? does it work?

taras (approved):

.github/workflows/www.yaml
jbolda: Aye, yep it works locally. TBD on if we got all of the other wires in place for it to work here.

PR #1215, opened 2026-07-15
🐛 Run scope destructors in reverse order of registration

taras:
# Motivation

I noticed this problem in TestKit using resources that wrap Playwright and using daemon from @effectionx/process. I didn't have a chance to figure out what it was while I was working on TestKit, but I was able to narrow it down for a problem I had with executable.md.

The problem occurs when async teardown expect another value to be present that was setup by an earlier resource. This dependency chain requires that the dependant resource is torn down before the dependency.  This seems to work correctly when the scope runs to completion, but not when the scope is destroyed explicitly.

# Approach

Invoke teardown operations in reverse (last first) to ensure that dependent teardown are invoked before the dependency. This makes explicit `destroy()` consistent with `TaskGroup.halt()` and normal task unwinding, which already tear down last-in, first-out.

### DIFF
--- a/lib/scope-internal.ts
+++ b/lib/scope-internal.ts
@@ -99,7 +99,8 @@ export function buildScopeInternal(
         while (destructors.size > 0) {
           let current = [...destructors];
           destructors.clear();
-          for (let destructor of current) {
+          for (let i = current.length - 1; i >= 0; i--) {
+            let destructor = current[i];
             try {
               yield* destructor();
             } catch (error) {
--- a/test/scope.test.ts
+++ b/test/scope.test.ts
@@ -1,6 +1,7 @@
 import {
   createContext,
   createScope,
+  ensure,
   resource,
   run,
   sleep,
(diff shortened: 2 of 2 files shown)


taras:
**Scope of the bug — this is a v4 regression from v3, and it affects the current stable release (4.0.3 = npm `latest`), not just the 4.1 alphas.**

Both regression scenarios from this PR, run against the published packages:

| version | teardown order | drain protocol |
|---|---|---|
| `effection@3` | `["third","second","first"]` — reverse, correct | ✅ `destroy()` completes |
| `effection@4.0.3` (`latest`) | `["first","second","third"]` — forward | ❌ `destroy()` deadlocks |
| `effection@4.1.0-alpha.9` (`next`) | forward | ❌ deadlocks |

v3's frame-based teardown ran finalizers in reverse; the forward-order destructor set came in with the v4 scope architecture. Given 4.0.3 is `latest`, this fix is probably worth a backport to the 4.0.x line as well.

lib/scope-internal.ts
```diff
@@ -97,7 +97,7 @@ export function buildScopeInternal(
       let outcome = Ok();
       try {
         while (destructors.size > 0) {
-          let current = [...destructors];
+          let current = [...destructors].reverse();
```
cowboyd: I know it says that it does not impact the benchmarks. However, since this is on the hot path, the call to reverse() means that we walk the array twice and use double the memory instead of walking the array backwards.

```ts
let current = [...destructors];
destructors.clear();
for (let i = current.length - 1 ; i >= 0; i--) {
  let destructor = current[i];
}
```

This is the same pattern used in `TaskGroup`

cowboyd (changes requested):
Hard to believe that this managed to sneak this far in :)

Just one small change

lib/scope-internal.ts
taras: Good catch — switched to walking the array backwards in place, matching the `TaskGroup.halt()` pattern. No more `.reverse()` copy on the teardown hot path.

cowboyd (approved):

PR #1217, opened 2026-07-28
📝 Restore experimental API JSDoc

taras:
# Motivation

The `v4-next` experimental API pages show signatures for `createApi` and `api`, but no descriptions or examples. The website renderer is working correctly: commit `35376609` recreated `lib/api.ts` during the algebraic-effects refactor without carrying forward the public JSDoc previously added in `bee3aaca`, so `@deno/doc` has no JSDoc to render.

# Approach

- Restore current, technically accurate JSDoc for the `createApi` and `api` experimental exports.
- Document the type parameter, parameters, return value, version, and usage examples.

Validated with formatting, linting, type checking, the API test suite, and the website documentation pipeline. Both experimental pages now produce descriptions and examples; `createApi` also produces parameter and return documentation.


jbolda:
I am still seeing a bunch of docs from #1141 that somehow didn't make it into the main branch. I didn't check them all but `Provide` and `withResolver` is stale for example.

cowboyd (approved):

PR #1220, opened 2026-07-28
🐛 Advertise Pagefind bundle in the sitemap so Staticalize captures it

taras:
# Motivation

We need to update pagefinder generated static assets. Those assets execute based on markup in HTML, so the assets needs to be generated after the static content is generated. 

Previously, we were doing this in CI but it got messed up in changing our build setup.

# Approach

Instead of building in CI, we're now building the assets as part of the pagefind route which includes sitemap integration. We run a sub process that staticalizes the running site to a throwaway directory and indexes it with the Pagefind CLI, then `routemap` lists every file in the bundle in `sitemap.xml` so Staticalize captures the whole thing in its normal crawl.

At first glance this looks circular: Staticalize needs to run to build the site, but building the sitemap seems to need the site already built. It isn't, because **the sitemap comes from the live running server, not from a built site.**

### Sequence

```
outer staticalize                server                          generation subprocess
──────────────────               ──────                          ─────────────────────
GET /sitemap.xml ───────────────▶ routemap: bundle missing
                                  spawn generation, await ──────▶ staticalize CLI
                                    GET /sitemap.xml ◀───────────   (crawls live server)
                                    routemap: generation set
                                    → returns []  ───────────────▶ gets pages, no /pagefind/*
                                    (serves pages live) ◀───────── crawls them → pagefind-site/
                                                                   pagefind CLI: index → pagefind/
                                  generation done ◀──────────────  exits
                 ◀─ sitemap WITH  routemap: bundle exists now
                    /pagefind/*    → enumerate files
downloads pages
+ /pagefind/* ──────────────────▶ served from disk → built/
```

The outer sitemap request just blocks once (~2–3 min) while the bundle builds, then returns complete. Both crawls read the same live server; neither reads the other's file output.

# Verification

Verified against a **fresh** server (no pre-build): Staticalize's very first request (`/sitemap.xml`) triggers the build inline, waits for it, and captures all 232 bundle files across ~500 pages — no timeout, no deadlock. `pagefind-ui.js` / `pagefind-entry.json` and the fragment/index/filter/wasm data all land in the output, and `/search` works.


taras:
✅ Verified on the deploy preview (https://pr-1220--effection.netlify.app):

- `/sitemap.xml` lists **131 `/pagefind/*` entries** (runtime, manifest, wasm, and all index/fragment/filter files).
- Every bundle file is captured and served: `pagefind.js`, `pagefind-ui.js`, `pagefind-entry.json`, `wasm.en-us.pagefind`, and the `fragment`/`index`/`filter` data files all return `200`.
- `pagefind-entry.json` is valid — Pagefind 1.5.2, 110 pages indexed.
- `/search/` returns `200` and instantiates PagefindUI with the derived `bundlePath`/`baseUrl`.

So Staticalize captured the whole bundle through its normal crawl, with no separate pagefind step. The `/effection` composition on frontside.com (thefrontside/frontside.com#500) can be verified once this deploys to `effection.netlify.app`, since that PR's preview proxies the live site.

<blockquote><div><strong><a href="https://frontside.com/effection/">Effection</a></strong></div><div>Effection is a structured concurrency and effects framework for JavaScript.</div></blockquote>

taras:
Addressed all three:

1. **`www/pagefind.ts` — kept.** It's the entry point the plugin spawns as the generation subprocess (`deno run pagefind.ts`). Generation staticalizes the running site, and doing that crawl in-process corrupts concurrent page rendering, so it has to be a separate process — `pagefind.ts` is that process's `main()`.

2. **Test removed** — dropped `www/routes/pagefind-route.test.ts` and its fixture.

3. **No more stub — the triggering request now waits.** You're right that waiting is the correct behavior; the stub only existed for the generator's *own* crawl re-entering. The generation subprocess crawls this server, and `/search`'s `<script src=\"/pagefind/pagefind-ui.js\">` makes that crawl re-request `/pagefind/*` while the build is still running. Awaiting generation from inside the crawl that's driving it deadlocks, so only those re-entrant requests skip the wait — they now fall through to a 404 that the non-strict crawl skips, instead of a fake empty 200. The request that starts the build awaits it and serves the real bundle.

Verified locally: generation completes (210 pages, 232 files), `pagefind-ui.js` serves real content, 232 `/pagefind/*` sitemap entries, zero 500s, no deadlock.

cowboyd:
Not doubting this works, but this statement explaining away circularity itself seems circular.

> At first glance this looks circular: Staticalize needs to run to build the site, but building the sitemap seems to need the site already built. It isn't, because the sitemap comes from the live running server, not from a built site.

What happens to the page find route in the sub Staticalize? It must short circuit somehow?

taras:
@cowboyd https://github.com/thefrontside/effection/pull/1220/changes#diff-940f77e2d766b3cce9a60f40c49d368f44914ed93905e3ad3c977fb5a858d5faR40

`if (generation) { return [] }` is the short circuit.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection/pull/1220">🐛 Advertise Pagefind bundle in the sitemap so Staticalize captures it by taras · Pull Request #1220 · thefrontside/effection</a></strong></div><div>Motivation
We need to update pagefinder generated static assets. Those assets execute based on markup in HTML, so the assets needs to be generated after the static content is generated.
Previously,...</div></blockquote>

cowboyd:
Gotcha, so I think the proper framing is that it is self-referential (or recursive) with a base case.

cowboyd (approved):

PR #1221, opened 2026-07-28
🐛 Refresh each package clone once, not on every request

taras:
# Motivation

`useClone` (`www/lib/clones.ts`) runs `git fetch` + `git reset --hard origin/main` against a **single shared working tree on every call**. The `/x` and `/contrib` package pages call it on each render (via `useWorkspaces` and `useTaxonomy`), all pointing at the same `build/clones/thefrontside/effectionx` checkout.

Under a wide concurrent crawl (e.g. building the Pagefind search index), dozens of these git mutations run against the same repository simultaneously — `.git/index.lock` contention and half-applied resets — which surface as intermittent `500`s (`GET /contrib/watch 500 Internal Server Error`).

# Approach

Memoize the checkout per repository on the root scope: the first caller runs the git commands and every concurrent or later caller awaits that same task, so each repository is fetched and reset **exactly once** per server lifetime. This removes the races and the redundant fetches. Uses the same `useScope()` + `scope.run(...)` pattern already used in `lib/octokit.ts`.

# Verification

- 54 `/x` and `/contrib` pages hammered concurrently (40-way parallel) over 3 rounds: **all 200, zero non-200**, and no `index.lock` errors in the server log.
- `deno fmt`, `deno lint`, `deno check` on the changed file.

This is the root cause behind the transient failures the Pagefind indexing work (#1220) had to paper over with crawl retries.

### DIFF
--- a/www/lib/clones.ts
+++ b/www/lib/clones.ts
@@ -1,13 +1,17 @@
-import { createContext, type Operation, type Task, useScope } from "effection";
+import {
+  createContext,
+  Err,
+  Ok,
+  type Operation,
+  type Result,
+  type Task,
+  useScope,
+} from "effection";
 import { $ } from "../context/shell.ts";
 import { existsSync } from "node:fs";
 import { resolve } from "node:path";
 
 type Checkout = (nameWithOwner: string) => Operation<string>;
-
-type Attempt =
-  | { ok: true; dirpath: string }
-  | { ok: false; error: unknown };
 
 const Clones = createContext<Checkout>("clones");
 
@@ -16,7 +20,7 @@ export function* initClones(path: string): Operation<void> {
   yield* $(`mkdir -p ${path}`);
 
   let scope = yield* useScope();
-  let attempts = new Map<string, Task<Attempt>>();
+  let attempts = new Map<string, Task<Result<string>>>();
 
   // `git fetch` / `reset --hard` mutate a single working tree shared by every
   // request, so running them concurrently corrupts it (index-lock contention,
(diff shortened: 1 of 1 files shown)


www/lib/clones.ts
```diff
@@ -1,18 +1,58 @@
...
+}
+
+const ClonesContext = createContext<Clones>("clones");
 
 export function* initClones(path: string): Operation<void> {
   yield* $(`rm -rf ${path}`);
   yield* $(`mkdir -p ${path}`);
-  yield* Clones.set(path);
+  let scope = yield* useScope();
+  yield* ClonesContext.set({ basepath: path, scope, checkouts: new Map() });
```
taras: instead of storing all of these values, can you just store an upsert function that'll take a namWithOwner and return a checkout for that reposittory?

taras: Done — the context now holds a single upsert function (`Checkout`) that memoizes and returns the checkout for a repo; `basepath`/scope/map are all captured in the closure.

Two notes while I was in here:

- You're right that a resource isn't needed just for teardown — the `scope.run` tasks are children of the root scope, so they're already halted on shutdown.
- But holding the tasks did expose a real hazard: a `scope.run` op that **throws** tears down the scope it runs on. I confirmed with a quick test that a failing task exits the shared (app root) scope and raises an uncaught error — so one bad `git clone` would crash the server and poison every other checkout. So `cloneOrRefresh` now returns an outcome instead of throwing; the caller re-throws and the failed entry is evicted so a later request can retry. Verified: a failing repo stays isolated, other repos keep working, and the scope survives.

cowboyd (approved):

PR #1223, opened 2026-07-29
🐛 Fix API-doc pages that crash the MDX renderer on type signatures

### LINKED ISSUE #1222: Some API-doc pages fail to render: `{@link X}` cross-references throw "X is not defined"
## Summary

A handful of API-documentation pages on the effection website fail to render server-side. The markdown→JSX conversion throws `ReferenceError: <Symbol> is not defined`, so those pages return 500 and are silently dropped from the crawl (and therefore from the deployed site and the search index).

This is a **pre-existing docs-rendering bug**, unrelated to search — surfaced while working on the Pagefind indexing in #1220. The Pagefind crawl tolerates these failures (non-strict, skips them), so search still works, but the affected pages themselves are missing.

## Symptoms

During a full crawl of the running www app, 4 pages failed with:

```
Failed to convert markdown to JSXElement for …
  ReferenceError: Api is not defined
  ReferenceError: Chain is not defined
  ReferenceError: Process is not defined
  ReferenceError: Scope is not defined
```

The thrown error originates at `www/hooks/use-markdown.tsx:105` (`Failed to convert markdown to JSXElement for ${markdown}`).

## Likely cause

These symbols (`Api`, `Chain`, `Process`, `Scope`) are JSDoc `{@link Symbol}` cross-references in the API doc comments. When the doc markdown is compiled to MDX/JSX and evaluated, the `{@link X}` references appear to become bare identifiers that aren't in the MDX runtime scope, so evaluation throws `ReferenceError: X is not defined`. Example fragment seen in the logs:

```
* {@link createApi}   →  ReferenceError: Scope is not defined
```

So the fix is likely in how `{@link …}` references are transformed/escaped before/inside the markdown→JSX step (`www/hooks/use-markdown.tsx`, and the API reference rendering in `www/routes/api-reference-route.tsx`).

## Impact

- The affected API pages 500 and are absent from the deployed static site and the search index.
- Because the crawler is non-strict, this fails silently — no deploy error, just missing pages.

## Repro

1. Run the www app: `cd www && deno run -A main.tsx`
2. Crawl it (e.g. `deno task staticalize`, or just request the affected API pages).
3. Observe the `ReferenceError: … is not defined` render failures in the server log for pages whose docs contain `{@link Api|Chain|Process|Scope}` references.

## Suggested follow-up

- Reproduce the specific failing pages (the symbols point at the experimental `Api`/`createApi` docs and the `Chain`/`Process` pages).
- Fix the `{@link}` handling so cross-references render as links (or plain text) instead of being evaluated as identifiers.
- Consider making the docs crawl **strict** in CI once fixed, so a regression here fails the build instead of silently dropping pages.

taras:
Fixes #1222.

# Problem

Several API-documentation pages fail to render server-side. The markdown→JSX conversion (`www/hooks/use-markdown.tsx`) throws `ReferenceError: <Name> is not defined`, the description block renders empty, and — because the docs crawler is non-strict — the pages are silently dropped from the deployed site and the search index.

Two causes, both in JSDoc-derived content:

1. **Type expressions leak into MDX as JSX.** The existing escape only handled flat `<T>` / `<T, U>`. Anything richer left a raw `<Name>` that MDX parsed as a JSX component:
   - nested generics — `Operation<Chain<T>>`
   - unions — `From<A | B>`
   - arrays / nested — `Middleware<[], Promise<void>>`
   - `{@link}` inside a generic — `{@link Operation}<{@link B}>`
   → `ReferenceError: Chain / Api / Process / Scope is not defined`.

2. **Bare `{Name}` braces** (e.g. a malformed `{@link}` that lost its `@link`, leaving `{Scope}` / `{Api}`) are evaluated by MDX as JavaScript expressions → the same `ReferenceError`.

# Fix

Replace the narrow regex with `escapeMdxSyntax`, which walks the text **outside inline code and fenced blocks** and:

- escapes a `<` whenever it opens a type expression — followed by an uppercase identifier, `[`, or `{`, **or** attached to a preceding identifier (`Promise<`, `)<`);
- neutralizes stray `{` / `}` to their HTML entities so they render as literal braces.

Real HTML (`<div>`, ` `, closing `</div>`) sits at a word boundary starting lowercase (or is a closing tag) and is left intact, and nothing inside code is touched. The `.mdx` guides and package READMEs that also go through `useMarkdown` use no `{expr}` / `<Component>` outside code, so they're unaffected.

# Verification

- **Unit tests** (`www/hooks/use-markdown.test.ts`): the type-expression cases, the stray-brace cases, HTML-preserved cases, and code-span/fence preservation.
- **End-to-end**: crawling all **149** `/api/*` pages against a local server now logs **0** render failures (was 4). The pages that previously failed:
  - https://effection.netlify.app/api/v4-next/Api
  - https://effection.netlify.app/api/v4-next/Around
  - (plus the `Chain` / `Process` / `Scope` references on other pages)

## How to verify

```
cd www && deno task test        # unit tests, incl. escapeMdxSyntax
# or end-to-end:
cd www && GITHUB_TOKEN=… deno run -A main.tsx &      # serve
for p in $(curl -s localhost:8000/sitemap.xml | grep -oE '/api/[^< ]+' | sort -u); do
  curl -s -o /dev/null "localhost:8000$p"
done
# server log should contain no "is not defined" / "Failed to convert markdown to JSXElement"
```

Surfaced while working on Pagefind indexing (#1220): the search crawl tolerated these failures, so search still worked, but the affected doc pages were missing.

### DIFF
--- a/www/hooks/use-markdown.test.ts
+++ b/www/hooks/use-markdown.test.ts
@@ -3,8 +3,8 @@ import { expect } from "expect";
 import { createJsDocSanitizer, escapeMdxSyntax } from "./use-markdown.tsx";
 
 describe("createJsDocSanitizer", () => {
-  const sanitizer = createJsDocSanitizer();
-  const cases: [string, string][] = [
+  let sanitizer = createJsDocSanitizer();
+  let cases: [string, string][] = [
     ["{@link Context}", "[Context](Context)"],
     ["@{link Scope}", "[Scope](Scope)"],
     ["{@link spawn()}", "[spawn](spawn)"],
@@ -20,7 +20,7 @@ describe("createJsDocSanitizer", () => {
     ],
   ];
 
-  for (const [input, expected] of cases) {
+  for (let [input, expected] of cases) {
     it(`rewrites ${input}`, function* () {
       expect(yield* sanitizer(input)).toEqual(expected);
     });
--- a/www/hooks/use-markdown.test.ts
+++ b/www/hooks/use-markdown.test.ts
@@ -1,6 +1,6 @@
 import { assertEquals } from "@std/assert";
 import { run } from "effection";
-import { createJsDocSanitizer } from "./use-markdown.tsx";
+import { createJsDocSanitizer, escapeMdxSyntax } from "./use-markdown.tsx";
 
 const sanitizer = createJsDocSanitizer();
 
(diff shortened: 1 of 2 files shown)


cowboyd (dismissed):
We should use the same conventions for testing as elsewhere in the code for `www/`

For example:

https://github.com/thefrontside/effection/blob/v4/www/lib/command-parser.test.ts#L1


Also, looks like this is supported natively with remark so probably worth looking into using that code which will have "all the use cases" https://github.com/orgs/syntax-tree/discussions/123

taras:
Tests moved to the `describe`/`it`/`expect` convention.

Tried the remark route — it escapes `{Api}` but not the `<Type>` generics (CommonMark reads `<Scope>` as inline HTML and re-emits it raw → MDX still crashes), which is most of our cases. Disabling inline HTML fixes that but escapes real HTML too (e.g. `context.mdx`'s `<div>`) and reformats every doc. Kept the char-scanner; it handles all 149 pages and preserves HTML. Can revisit if you'd take that tradeoff.

cowboyd (approved):

cowboyd:
Looks like there are some listing problems.

PR #1228, opened 2026-08-26
📝 Add cleanup guide

joshamaju:
# Motivation

PR #1225 established an important teardown rule: synchronous cleanup belongs in finally, while asynchronous cleanup must use ensure(). The public guides need a focused explanation of that distinction, including why yielding during generator unwinding can lose a halt.

# Approach

Add a Cleanup guide to the Learn Effection section. The guide covers synchronous and asynchronous cleanup, scope boundaries around call() and scoped(), cleanup ordering, and the Effection 4.1 compatibility note for scoped() teardown.

### DIFF
--- a/docs/cleanup.mdx
+++ b/docs/cleanup.mdx
@@ -1,0 +1,126 @@
+Effection shuts down an operation when it returns, throws an error, or is
+halted. Cleanup belongs to that shutdown path, but the way you express it
+depends on whether the cleanup is synchronous or asynchronous.
+
+The rule is simple:
+
+- Put synchronous cleanup in `finally`.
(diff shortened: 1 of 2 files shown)


docs/cleanup.mdx
```diff
@@ -0,0 +1,126 @@
...
+a spawned child, register it after spawning the child. That makes your cleanup
+run first, while the child is still alive.
+
+```ts
+let worker = yield* spawn(runWorker());
+
+yield* ensure(function* () {
+  yield* tellWorkerToStop(worker);
+});
+```
```
cowboyd, line 117: `worker` is going to be a `Task` is the implication that `tellworkerToStop(worker)` will call `.halt()` ?

PR #1229, opened 2026-08-27
🐛 preserve unknown error causes

jrandolf:
# Motivation

Effection currently converts every non-`Error` failure into a `ThrownValueError` so `Result.error` remains an `Error`. [#1149](https://github.com/thefrontside/effection/pull/1149) introduced that contract with the public `Result` API. Changing it is a breaking change.

The original value remains available through `cause`, but Effection also uses `Result` inside the runtime and rethrows or rejects with `result.error`. Once a non-`Error` value crosses an Effection boundary, callers catch the wrapper instead of the value that was thrown. Identity checks and type guards therefore receive a different object.

SvelteKit is one concrete case. Its [`HttpError` and `Redirect` values are plain classes](https://github.com/sveltejs/kit/blob/95ee35083f663622a3f370af0a41c530c7d1902c/packages/kit/src/exports/internal/shared.js#L3-L44), and [`isHttpError()` and `isRedirect()` identify them with `instanceof`](https://github.com/sveltejs/kit/blob/95ee35083f663622a3f370af0a41c530c7d1902c/packages/kit/src/exports/index.js#L96-L150). Passing either value through an Effection operation makes those guards fail.

Keeping `Result.error: Error` would require every automatic propagation path to unwrap `ThrownValueError.cause`, while explicit `Err()` and `allSettled()` results would continue to contain the wrapper. This PR uses JavaScript semantics in both cases: `Result.error` is `unknown`, and `Err(cause)` stores the exact cause. Code that needs an `Error` for logging or presentation can normalize it at that boundary.

# Approach

- change `Result.error` and rejection parameters to `unknown`
- store the exact cause in `Err()` and remove `ThrownValueError`
- verify identity preservation through `Err()` and `until()`

### DIFF
--- a/lib/action.ts
+++ b/lib/action.ts
@@ -13,7 +13,7 @@ import type { Operation } from "./types.ts";
  *           action is resolved, rejected, or discarded.
  */
 interface Executor<T> {
-  (resolve: (value: T) => void, reject: (error: Error) => void): () => void;
+  (resolve: (value: T) => void, reject: (error: unknown) => void): () => void;
 }
 
 /**
@@ -53,7 +53,7 @@ export function action<T>(executor: Executor<T>, desc?: string): Operation<T> {
       let resolve = (value: T) => {
         settle(Ok(value));
       };
-      let reject = (error: Error) => {
+      let reject = (error: unknown) => {
         settle(Err(error));
       };
       let discard = executor(resolve, reject);
--- a/lib/future.ts
+++ b/lib/future.ts
@@ -5,7 +5,7 @@ import { withResolvers } from "./with-resolvers.ts";
 export interface FutureWithResolvers<T> {
   future: Future<T>;
   resolve(value: T): void;
-  reject(error: Error): void;
+  reject(error: unknown): void;
 }
 export function createFuture<T>(): FutureWithResolvers<T> {
   let promise = lazyPromiseWithResolvers<T>();
@@ -16,7 +16,7 @@ export function createFuture<T>(): FutureWithResolvers<T> {
     operation.resolve(value);
   };
 
-  let reject = (error: Error) => {
+  let reject = (error: unknown) => {
     promise.reject(error);
     operation.reject(error);
   };
--- a/lib/lazy-promise.ts
+++ b/lib/lazy-promise.ts
@@ -14,7 +14,7 @@ export function lazyPromiseWithResolvers<T>(): PromiseWithResolvers<T> {
   let resolve = ((value: T) => settle(Ok(value))) as PromiseWithResolvers<
     T
   >["resolve"];
-  let reject = (error: Error) => settle(Err(error));
+  let reject = (error: unknown) => settle(Err(error));
 
   let promise = lazyPromise<T>((resolve, reject) => {
     let record = (result: Result<T>) => {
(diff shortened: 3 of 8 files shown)


cowboyd:
Hey @jrandolf, Thanks for the PR!

This is definitely a breaking change. I'm not opposed to it, but I'm curious why you think it is lossy? All information is retained, and If you are capturing a result object in the first place then by definition you are already preventing code upwards on the stack from seeing the throw value. In order to re-expose the stack to the error, you are would already need to destructure via the `.ok` guard. It seems to me at the point of de-structuring, you could re-raise any thrown value at the same point.

```
export function unbox<T>(result: Result<T>): T {
  if (result.ok) {
    return result.value;
  } else {
    throw result.error instanceof ThrowValueError ? result.error.cause : result.error;
  }
}
```

Or is the problem that Effection _itself_ does not do this when it re-throws an error? That would definitely be a problem

jrandolf:
Yes — your last sentence is exactly the problem. And you are right that “lossy” was imprecise: the original value is retained in `.cause`; what is not preserved is the identity and semantics of the value being thrown or rejected.

Effection wraps the value in `Err()`, then re-exposes `result.error` unchanged. The existing `unbox()` throws `result.error`; `Do`, `Task` (both the yield and Promise paths), and `race()` do the same. So this rejects with `ThrownValueError`, not with `reason`:

```ts
const reason = {};
await run(function* () {
  yield* until(Promise.reject(reason));
});
```

That means `caught === reason` and guards against the caught value fail. Your proposed change to `unbox()` would fix the paths that call it. The remaining issue is that several runtime paths bypass `unbox()` and rethrow or reject `result.error` directly.

We could update every one of those paths to unwrap, or preserve the cause in `Err()` at the common boundary. This PR takes the latter route because it makes `Result` match JavaScript semantics—a thrown or rejected value is `unknown`, not necessarily an `Error`—and prevents future propagation paths from forgetting to unwrap. I will use “does not preserve thrown-value identity/semantics” instead of “lossy” to make that distinction clear.

jrandolf:
Yes. Effection itself rethrows the wrapper.

[#1149](https://github.com/thefrontside/effection/pull/1149) established `Result.error: Error`, added `ThrownValueError`, and added coverage requiring `until(Promise.reject("error"))` to expose that wrapper. The behavior is deliberate.

Changing `unbox()` would fix the paths that call it, including `callcc`, scope destruction, and task-group shutdown. Other runtime paths bypass it: `Do`, `race`, coroutine resumption, the task iterator, the task Promise adapter, and background error propagation all use the normalized `result.error` directly. Preserving `Result.error: Error` while restoring the original thrown value would require the same unwrapping rule at each of those boundaries.

This PR changes the common representation instead. `Result.error` becomes `unknown`, and `Err()` stores the exact cause, so explicit results and automatic propagation follow the same rule. The wrapper retains the original information in `cause`; the behavioral difference is that Effection changes the identity and observable type of the caught value.

SvelteKit shows why that matters. Its `HttpError` and `Redirect` are plain classes, and `isHttpError()` and `isRedirect()` recognize them with `instanceof`. Wrapping either value breaks those guards. Next.js does not have this problem because its redirect sentinel is an `Error` and already passes through unchanged.

cowboyd:
Right. This could be a change that could be accepted as part of a 5.0 release, but since it breaks a public API, it cannot go into the v4 release series as-is. However, failure to unbox _is_ a bug in v4 and should be fixed.

cowboyd:
I think the path forward here is to start a v5 branch with this as the first change on it, and then add the unwrapping behavior to v4. Is that latter item something you'd be willing to take on?

jrandolf:
Let's do it!

jrandolf:
@cowboyd I opened #1230 with the v4-compatible fix: explicit `Result.error` values remain `Error`, while automatic operation, coroutine, task, race, halt, and Promise boundaries restore the original thrown value. The full 235-step suite, lint, formatting, and type checks pass. I left this PR unchanged for the future v5 branch.

PR #1230, opened 2026-08-31
🐛 unwrap non-Error runtime failures

jrandolf:
# Motivation

Effection v4 deliberately keeps `Result.error` typed as `Error`, wrapping non-`Error` causes in `ThrownValueError`. Automatic propagation currently rethrows or rejects that wrapper, so caught values lose their original identity and runtime guards such as `instanceof` fail. This is the v4-compatible follow-up discussed in #1229.

# Approach

- keep explicit failed `Result` values wrapped as `Error`
- restore the original cause when failures cross operation, coroutine, task, race, halt, and Promise boundaries
- cover promise rejections, races, background child failures, and direct `unbox()` behavior

Validation: `deno fmt`, `deno lint`, `deno check mod.ts experimental.ts`, and the full 235-step test suite.

### DIFF
--- a/lib/coroutine.ts
+++ b/lib/coroutine.ts
@@ -1,5 +1,5 @@
 import { ReducerContext } from "./reducer.ts";
-import { Err, Ok } from "./result.ts";
+import { Err, Ok, unbox } from "./result.ts";
 import type { Coroutine, Effect, Operation, Scope } from "./types.ts";
 import type { Maybe } from "./maybe.ts";
 import type { Result } from "./result.ts";
@@ -72,10 +72,16 @@ export function createCoroutine<T>(
       let { resumeWith } = routine.data;
       if (!resumeWith.ok) {
         data.unwinding = false;
+        let error: unknown = resumeWith.error;
+        try {
+          unbox(resumeWith);
+        } catch (cause) {
+          error = cause;
+        }
         if (iterator.throw) {
-          return iterator.throw(resumeWith.error);
+          return iterator.throw(error);
         } else {
-          throw resumeWith.error;
+          throw error;
         }
       } else if (data.unwinding && !data.critical) {
         data.unwinding = false;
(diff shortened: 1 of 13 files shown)


cowboyd:
@jrandolf Looks good. 

There is one potential regression: I don't think that the type of `Exit` in `main()` is safe any more https://github.com/thefrontside/effection/blob/v4/lib/main.ts#L167

Whereas before, we knew all throws would be boxed in an error, we don't anymore because we could (and should) see the original cause. I think now `Exit.error` should be more properly typed as `unknown`, not `Error | undefined`

As it stands, if someone throws "null" or "0" or "false" it won't print anything. A minor thing, but it should actually simplify the type and remove the need for the type assertion on https://github.com/thefrontside/effection/blob/v4/lib/main.ts#L139

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection/blob/v4/lib/main.ts">effection/lib/main.ts at v4 · thefrontside/effection</a></strong></div><div>Structured concurrency and effects for JavaScript. Contribute to thefrontside/effection development by creating an account on GitHub.</div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/effection/blob/v4/lib/main.ts">effection/lib/main.ts at v4 · thefrontside/effection</a></strong></div><div>Structured concurrency and effects for JavaScript. Contribute to thefrontside/effection development by creating an account on GitHub.</div></blockquote>

jrandolf:
@cowboyd Fixed in 26587dc5: `Exit.error` is now `unknown`, the assertion is gone, and `main()` checks for the property so falsy causes such as `false`, `0`, and `null` are still printed. I added a `throw false` regression; formatting, lint, type checks, and all 208 core test steps pass.

cowboyd (approved):
The codspeed benchmark regression is a false negative because it does not actually test error cases. Most likely due to differing node envs.

PR #1231, opened 2026-09-01
Accept PromiseLike values in call() and until()

jrandolf:
# Motivation

`call()` and `until()` only consume thenables, but their public types require full `Promise` implementations. This rejects valid `PromiseLike` values, and `until()` also calls `.catch()`, which is not part of the `PromiseLike` contract.

# Approach

- Accept `PromiseLike` values in `Callable`, `call()`, and `until()`.
- Pass the rejection handler directly to `.then()` so `until()` uses only the `PromiseLike` interface.
- Cover both APIs with promise-like regression tests.

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -27,7 +27,7 @@ import type { Operation } from "./types.ts";
  * @since 3.0
  */
 export interface Callable<
-  T extends Operation<unknown> | Promise<unknown> | unknown,
+  T extends Operation<unknown> | PromiseLike<unknown> | unknown,
   TArgs extends unknown[] = [],
 > {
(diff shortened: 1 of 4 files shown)


cowboyd (approved):
Looks like there is a merge conflict, but otherwise great! Thanks 🎉

jrandolf:
The Windows failure looks transient and unrelated to this PR: it failed in `race() ... rejects when one of the given operations rejects asynchronously first`, while both new `call()`/`until()` tests passed. The same `90926724` base commit passed Windows on `v4` and in #1230. I don't have permission to rerun the failed Actions job; could a maintainer rerun it?

PR #1238, opened 2026-09-18
📝 Show the full `until()` + `useAbortSignal()` idiom

taras:
# Motivation

Two places taught a `fetch()` example that cannot run. `docs/scope.mdx`
§ "Abort Signal" and the `useAbortSignal()` JSDoc (which renders as
`/api/v4/useAbortSignal`) both `yield*`'d a promise directly, omitting
`until()`:

```javascript
let response = yield* fetch('/some/url', { signal });
```

As `docs/faq.mdx` already states, that is a TypeScript error. The scope example
compounded it with a bare `yield response.text()`.

Separately, `docs/async-rosetta-stone.mdx` § `Promise` <=> `Operation` — the
section a reader or an agent searching for "Promise", "Operation" or `until()`
actually lands on — showed `until(promise)` by itself. That covers waiting on a
promise but not cancelling one, so the complete interop pattern had to be
inferred by stitching two pages together.

# Approach

**`docs/async-rosetta-stone.mdx` becomes the canonical location** for Promise
interoperability. Its `Promise` <=> `Operation` section gains one sentence and
the full cancellable example, placed right after the existing `until(promise)`
example:

```javascript
import { until, useAbortSignal } from 'effection';

function* fetchUser(id) {
  let signal = yield* useAbortSignal();
  let response = yield* until(fetch(`/users/${id}`, { signal }));

  return yield* until(response.json());
}
```

`until()` waits for the promise; `useAbortSignal()` binds the work it
represents to the lifetime of the current operation. Both halves are needed,
and both are now visible in one place.

**The other two changes are kept to the minimum that makes the broken examples
correct**, rather than restating the explanation:

- `docs/scope.mdx` keeps its original `request(url)` example, shape and error
  handling intact. It imports `until`, wraps the `fetch()` promise, and
  replaces `yield response.text()` with `yield* until(response.text())`. The
  surrounding Abort Signal prose is unchanged from the base branch.
- `lib/abort-signal.ts` gets the corrected example in its `@example` block.

## Deliberately out of scope

`docs/tutorial.mdx` is already correct. `lib/until.ts`, `docs/upgrade.mdx` and
`docs/scope.mdx` § "Creating a brand new `Scope`" show `until(fetch(url))` with
no signal; they teach `until()` on its own and do not contradict this pattern.
Broader stale-example cleanup is a separate task. No new sections anywhere —
the rosetta stone addition sits inside the existing `Promise` <=> `Operation`
section — and `docs/structure.json` / site navigation are untouched.


cowboyd (approved):
👍🏻

PR #1240, opened 2026-09-19
📝 Correct async cleanup guidance and document orderly Promise teardown

taras:
# Motivation

The documentation incorrectly said that Effection safely handles `yield*` inside `finally`. During halt, this can break generator unwinding and allow execution to continue past the operation being halted.

Related teardown behavior was also underspecified: aborting a Promise API does not mean its Promise has settled, `all()` did not document how it handles sibling tasks after an error, and `llms.txt` did not clearly distinguish EffectionX from Effection core.

# Approach

- Use `finally` for synchronous cleanup and `ensure()` for asynchronous cleanup.
- Show Promise teardown in the correct order: request cancellation, then wait for settlement.
- Document that `all()` halts started sibling tasks and waits for their teardown before propagating an error.
- Identify `@effectionx/*` packages as extensions maintained in the separate EffectionX repository.

These are documentation and JSDoc changes only; runtime behavior and public APIs are unchanged.


docs/async-rosetta-stone.mdx
```diff
@@ -141,6 +141,38 @@ function* fetchUser(id) {
...
+  });
+
+  let signal = yield* useAbortSignal();
+
+  request = fetch('/uploads', { method: 'POST', body, signal });
+
+  return yield* until(request);
+}
+```
+
```
cowboyd, line 198: Can't this be simplified? e.g.:

```javascript
import { ensure, until } from 'effection';

function* upload(body) {
  let controller = new AbortController();
  let { signal } = controller;

  let request = fetch('/uploads', { method: 'POST', body, signal });

  yield* ensure(function*() {
    controller.abort();
    yield* until(request.catch(() => {})
  });

  return yield* until(request);
}
```

taras: Is this simpler?

cowboyd: I think so, because first all variables are bound at all times after they're declared and second it does not rely on implicit timing of code blocks.

taras: I agree that `all variable are bound at all times after they're declared` and not relying on `implicit timing of code blocks` is an improvement, but I don't necessarily think it's simpler because this example weaves together 5 different concepts

1. manually creating abort controller which most people have never done (our focus is agents, but still)
2. calling fetch without handling the process
3. aborting the controller explicitly
4. using promise chaining api mixed with effection operations
5. using .catch to swallow a promise

This is a lot. It feels like there is an opportunity to actually make it simpler.

taras: We might just have to covert all of this in documentation.

